PR #5, opened 2025-12-02
refactor: Make build more сonvenient for outside contributors

MBkkt:
Now they don't need manually call `git submodule update --init --jobs 32`
Also removed `SDB_CLUSTER ON` from cmake preset.
Make cmake presets more user friendly


mkornaukhov (approved):
OK

PR #7, opened 2025-12-03
fix: Use new velox coercions

MBkkt:


server/pg/sql_analyzer_velox.cpp
```diff
@@ -165,26 +166,17 @@ const containers::FlatHashMap<std::string_view, lp::SpecialForm> kSpecialForms{
...
+  {"in", lp::SpecialForm::kIn},
+};
 
 velox::TypePtr ResolveFunction(const std::string& function_name,
                                const std::vector<velox::TypePtr>& arg_types,
                                std::vector<velox::TypePtr>* arg_coercions) {
-  if (IsSpecialFormFunction(function_name)) {
-    if (function_name == "in") {
-      return velox::BOOLEAN();
-    }
```
mkornaukhov, line 180: Why is this hack no longer needed?
I don't know the  reason for its presence, but didn't see anything in new Velox commits that related with it.

MBkkt: There's no test for this, if it's needed let's revert it/implement in velox and add test

pashandor789: There is test i know.

mkornaukhov: Tests with IN are passing, so looks like OK

MBkkt: @pashandor789 write it?

PR #15, opened 2025-12-05
[WIP] Add pg::unknown type

mkornaukhov:


mkornaukhov:
For now there are 2 problems.

- [ ] Does not work without `varchar -> unknown` cast support
- [ ] `UnknownCastOperator::castTo` sometimes called with unknown type. Find out is it OK or not.

PR #32, opened 2025-12-08
feat: Make CTE appear in subqueries

pashandor789:


server/pg/sql_analyzer_velox.cpp
```diff
@@ -460,11 +460,17 @@ struct AliasResolver : ColumnResolver {
...
-struct RootState;
+struct State;
 
 using ColumnRefHook =
   std::function<lp::ExprPtr(std::string_view, const ColumnRef&)>;
 
+struct CTE {
+  const CommonTableExpr* node = nullptr;
+  // ^ we could use LogicalPlanNode, but axiom renaming logic
+  // doesn't consider DAG; so we reprocess CTE each time to make it tree
```
MBkkt, line 471: also unused fields detection doesn't work with DAG

PR #35, opened 2025-12-08
support Wordnet for synonym tokenizer

k-morozov:


libs/iresearch/include/iresearch/analysis/wordnet_synonyms_tokenizer.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+  SynonymsGroups synset;
+  SynonymsMap mapping;
+
+  for (const auto& line : lines) {
+    line_number++;
+    if (line.empty())
+      continue;
+
+    const std::string_view syn_set_id = line.substr(2, 9);
```
gnusi: I think we need to harden checks a bit:
* let's ensure that line conforms WordNet format, i.e. `s(<comma-separated-list>)`, otherwise let's return an error
* It's unclear if all possible `syn_set_id`s are of length 7, let's search for the first comma in order to find the end of the syn_set_id

libs/iresearch/include/iresearch/analysis/wordnet_synonyms_tokenizer.cpp
```diff
@@ -0,0 +1,93 @@
...
+      synset.push_back(syn_set_id);
+    }
+
+    const size_t start = line.find('\'') + 1;
+    const size_t end = line.rfind('\'');
+
+    if (start == std::string::npos || end == std::string::npos ||
+        start >= end) {
+      return std::unexpected<sdb::Result>{std::in_place,
+                                          sdb::ERROR_VALIDATION_BAD_PARAMETER,
```
gnusi: Let's change to sdb::ERROR_BAD_PARAMETER please (in other places too)

libs/iresearch/include/iresearch/analysis/wordnet_synonyms_tokenizer.cpp
```diff
@@ -0,0 +1,93 @@
...
+    const std::string_view syn_set_id = line.substr(2, 9);
+
+    // I couldn't find explicit guarantees that SynSet IDs must appear
+    // sequentially, but popular implementations rely on this assumption.
+    if (last_syn_set_id != syn_set_id) {
+      synset.push_back(syn_set_id);
+    }
+
+    const size_t start = line.find('\'') + 1;
+    const size_t end = line.rfind('\'');
```
gnusi: Given a WordNet line is a CSV, using on `,` for extracting values seems to be more reliable, e.g. what if we have a string with `'` inside it

libs/iresearch/include/iresearch/analysis/wordnet_synonyms_tokenizer.cpp
```diff
@@ -0,0 +1,93 @@
+#include "wordnet_synonyms_tokenizer.hpp"
```
gnusi, line 21: Would you mind to add SereneDB disclaimer to the every new file please, you can copy it from  https://github.com/serenedb/serenedb/blob/main/server/pg/executor.cpp#L1

Thanks!

libs/iresearch/include/iresearch/analysis/wordnet_synonyms_tokenizer.cpp
k-morozov: @gnusi Hi.
I had the same thought - I had considered a more universal approach. But then I found the implementation in [lucene](https://github.com/apache/lucene/blob/main/lucene/analysis/common/src/java/org/apache/lucene/analysis/synonym/WordnetSynonymParser.java#L54), and it seemed simpler. Of course, we could use a more reliable approach and it makes sense.

libs/iresearch/include/iresearch/analysis/wordnet_synonyms_tokenizer.cpp
k-morozov: Added

libs/iresearch/include/iresearch/analysis/wordnet_synonyms_tokenizer.cpp
k-morozov: Updated. Solr tokenizer as well.

libs/iresearch/include/iresearch/analysis/wordnet_synonyms_tokenizer.cpp
k-morozov: Updated logic

libs/iresearch/include/iresearch/analysis/wordnet_synonyms_tokenizer.cpp
```diff
@@ -0,0 +1,155 @@
...
+#include <string_view>
+#include <utility>
+
+namespace irs::analysis {
+
+namespace {
+
+constexpr static size_t kWordnetCountParams = 6;
+
+const std::regex kWordnetPattern(R"(s\(([^)]*)\)\.)");
```
MBkkt: Can you use re2 instead, please?

libs/iresearch/include/iresearch/analysis/wordnet_synonyms_tokenizer.cpp
```diff
@@ -0,0 +1,155 @@
...
+  WordnetSynonymsTokenizer::SynonymsGroups&& groups,
+  WordnetSynonymsTokenizer::SynonymsMap&& mapping)
+  : _groups(std::move(groups)), _mapping(std::move(mapping)) {}
+
+bool WordnetSynonymsTokenizer::next() {
+  if (!_term_exists) {
+    return false;
+  }
+  _term_exists = false;
+  return true;
```
MBkkt, line 136: NIT
```suggestion
  return std::exchange(_term_exists, false);
```

libs/iresearch/include/iresearch/analysis/wordnet_synonyms_tokenizer.hpp
```diff
@@ -0,0 +1,73 @@
...
+  explicit WordnetSynonymsTokenizer(SynonymsGroups&& groups,
+                                    SynonymsMap&& mapping);
+  Attribute* GetMutable(TypeInfo::type_id type) final {
+    return irs::GetMutable(_attrs, type);
+  }
+  bool next() final;
+  bool reset(std::string_view data) final;
+
+ private:
+  const SynonymsGroups _groups;
```
MBkkt: This is completely unused, right? Lets remove it

libs/iresearch/include/iresearch/analysis/wordnet_synonyms_tokenizer.cpp
```diff
@@ -0,0 +1,155 @@
...
+}
+
+bool WordnetSynonymsTokenizer::reset(std::string_view data) {
+  auto& offset = std::get<irs::OffsAttr>(_attrs);
+  offset.start = 0;
+  offset.end = data.size();
+
+  if (const auto it = _mapping.find(data); it == _mapping.end()) {
+    _term_exists = false;
```
MBkkt, line 145: I think we want to assign input data in such case.
Because we want to index unknown words as is (this can be controlled by some bool settings, but in general both behavior valid and potentially useful in both cases (wordnet and solr)

libs/iresearch/include/iresearch/analysis/wordnet_synonyms_tokenizer.cpp
```diff
@@ -0,0 +1,155 @@
...
+  } else {
+    auto& term = std::get<TermAttr>(_attrs);
+    term.value = ViewCast<byte_type>(it->second);
+    _term_exists = true;
+  }
+
+  return true;
```
MBkkt: if term_exists_ can be false, here should be false

k-morozov: Seems we could return `_term_exists`.

libs/iresearch/include/iresearch/analysis/wordnet_synonyms_tokenizer.hpp
k-morozov: Updated

libs/iresearch/include/iresearch/analysis/wordnet_synonyms_tokenizer.cpp
k-morozov: Added re2

libs/iresearch/include/iresearch/analysis/wordnet_synonyms_tokenizer.cpp
k-morozov: Without assign input data.

libs/iresearch/include/iresearch/analysis/wordnet_synonyms_tokenizer.cpp
```diff
@@ -0,0 +1,172 @@
...
+constexpr static size_t kWordnetCountParams = 6;
+
+const RE2 kWordnetPattern(R"(s\(([^)]*)\)\.)");
+
+bool RegexWordnet(const std::string_view s_list, std::string_view* result) {
+  if (s_list.length() <= 2) {
+    return false;
+  }
+
+  re2::StringPiece input(s_list.data(), s_list.size());
```
MBkkt: NIT: it's unnecessary re2::StringView is using for std::string_view, so this variable can be removed

libs/iresearch/include/iresearch/analysis/wordnet_synonyms_tokenizer.cpp
```diff
@@ -0,0 +1,172 @@
...
+sdb::ResultOr<std::vector<std::string_view>> ParseParams(
+  const std::string_view line) {
+  std::string_view params;
+
+  if (!RegexWordnet(line, &params)) {
+    return std::unexpected<sdb::Result>{std::in_place,
+                                        sdb::ERROR_BAD_PARAMETER};
+  }
+
+  std::vector<std::string_view> outputs(absl::StrSplit(params, ','));
```
MBkkt: NIT codestyle
```suggestion
  std::vector<std::string_view> outputs = absl::StrSplit(params, ',');
```

libs/iresearch/include/iresearch/analysis/wordnet_synonyms_tokenizer.cpp
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    mapping[synonym].push_back(synset.back());
+
+    last_syn_set_id = syn_set_id;
+  }
+
+  for (auto& [_, synset] : mapping) {
+    // expect that we don't have duplicate.
+    std::sort(synset.begin(), synset.end());
```
MBkkt:
```suggestion
    // expect that we don't have duplicate.
    absl::c_sort(synset);
    if (absl::c_adjacent_find(synset) != synset.end()) {
       return error;
    }
```

libs/iresearch/include/iresearch/analysis/wordnet_synonyms_tokenizer.cpp
```diff
@@ -0,0 +1,172 @@
...
+                                          "Failed parse line ", line_number};
+    }
+
+    const std::string_view syn_set_id = (*params)[0];
+
+    // I couldn't find explicit guarantees that SynSet IDs must appear
+    // sequentially, but popular implementations rely on this assumption.
+    if (last_syn_set_id != syn_set_id) {
+      synset.push_back(syn_set_id);
+    }
```
MBkkt: I think this and synset vector can be removed

libs/iresearch/include/iresearch/analysis/wordnet_synonyms_tokenizer.cpp
```diff
@@ -0,0 +1,172 @@
...
+#include <iresearch/analysis/pipeline_tokenizer.hpp>
+#include <iresearch/analysis/token_attributes.hpp>
+#include <string_view>
+#include <utility>
+
+namespace irs::analysis {
+
+namespace {
+
+constexpr static size_t kWordnetCountParams = 6;
```
MBkkt: NIT: static isn't needed in anon namespace
Also codestyle: no spaces between namespaces and "static constexpr" instead of "constexpr static"
```suggestion
namespace irs::analysis {
namespace {

constexpr size_t kWordnetCountParams = 6;
```

libs/iresearch/include/iresearch/analysis/wordnet_synonyms_tokenizer.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+const RE2 kWordnetPattern(R"(s\(([^)]*)\)\.)");
+
+bool RegexWordnet(const std::string_view s_list, std::string_view* result) {
+  if (s_list.length() <= 2) {
+    return false;
+  }
+
+  re2::StringPiece input(s_list.data(), s_list.size());
+  re2::StringPiece params;
```
MBkkt: NIT: it's unnecessary, because re2::StringView is using for std::string_view, so these variables can be removed
```suggestion
```

libs/iresearch/include/iresearch/analysis/wordnet_synonyms_tokenizer.cpp
```diff
@@ -0,0 +1,172 @@
...
+  std::string_view last_syn_set_id = "";
+  size_t line_number{};
+
+  SynonymsGroups synset;
+  SynonymsMap mapping;
+
+  for (const auto& line : lines) {
+    line_number++;
+    if (line.empty())
+      continue;
```
MBkkt: codestyle
```suggestion
    if (line.empty()) {
      continue;
    }
```

libs/iresearch/include/iresearch/analysis/solr_synonyms_tokenizer.hpp
MBkkt, line 1: Can you remove `_holder` variable and use term.value instead?

MBkkt (commented):
Image
Also there's no test for this error case

libs/iresearch/include/iresearch/analysis/wordnet_synonyms_tokenizer.cpp
k-morozov: Updated

libs/iresearch/include/iresearch/analysis/wordnet_synonyms_tokenizer.cpp
k-morozov: Removed

libs/iresearch/include/iresearch/analysis/wordnet_synonyms_tokenizer.cpp
k-morozov: Updated

libs/iresearch/include/iresearch/analysis/wordnet_synonyms_tokenizer.cpp
k-morozov: Updated.

libs/iresearch/include/iresearch/analysis/solr_synonyms_tokenizer.hpp
k-morozov: I suggest keeping `_holder`.  

If the synonym exists in our table, we use  `_synonyms` for loading `term`. However, if there is no synonym, we will use `term.value`. 

Given that `term` is part of the public interface, this could be ambiguous: in some cases it will be filled, and in others it will be empty after a reset.

libs/iresearch/include/iresearch/analysis/wordnet_synonyms_tokenizer.cpp
```diff
@@ -0,0 +1,159 @@
...
+  }
+  return outputs;
+}
+}  // namespace
+
+sdb::ResultOr<WordnetSynonymsTokenizer::SynonymsMap>
+WordnetSynonymsTokenizer::Parse(const std::string_view input) {
+  std::vector<std::string_view> lines = absl::StrSplit(input, '\n');
+
+  std::string_view last_syn_set_id = "";
```
MBkkt: Unused variable (only assign)

PR #43, opened 2025-12-10
Improve DataSink locking and write performance

Dronplane:
Improves DataSink locking (closes https://github.com/serenedb/cluster/issues/5 ) by locking only keys and bypass regular rocksdb concurrency controls.

Improves general performance by reducing calls to memory reclaim (as it is quite slow) - reclaim starts only when at least 10Mb is allocated.

Input Kind checks are moved to DEBUG check  - as we have implemented proper type casts outside TableWriter so  DataSink should never encounter invalid Kinds.

Benchmarking results:
Inserted numeric column with server side data generation. Rows count 125k
Timings before:
1  column  260ms
2 columns 510ms
3 columns 770ms

Timings after:
1 column 84ms
2 columns 140ms
3 columns 205ms

### DIFF
--- a/server/storage_engine/engine_selector_feature.cpp
+++ b/server/storage_engine/engine_selector_feature.cpp
@@ -28,6 +28,7 @@
 #ifdef SDB_CLUSTER
 #include "replication/replication_feature.h"
 #endif
+
 namespace sdb {
 
 EngineSelectorFeature::EngineSelectorFeature(Server& server)


server/connector/data_sink.cpp
```diff
@@ -93,18 +93,29 @@ void RocksDBDataSink::appendData(velox::RowVectorPtr input) {
   std::string name = _object_key;
   const auto parent_size = name.size();
   velox::IndexRange all_rows(0, input->size());
-  const auto& input_type = input->type()->asRow();
+  [[maybe_unused]] const auto& input_type = input->type()->asRow();
```
mkornaukhov, line 96: Looks like there are no reasons for `[[maybe_unused]]`?

server/storage_engine/engine_selector_feature.cpp
```diff
@@ -28,7 +28,6 @@
 #ifdef SDB_CLUSTER
 #include "replication/replication_feature.h"
 #endif
-
```
mkornaukhov, line 31: Extra diff

server/connector/data_sink.cpp
```diff
@@ -1993,10 +2010,13 @@ void RocksDBDeleteDataSink::appendData(velox::RowVectorPtr input) {
...
   absl::StrAppend(&key, _object_key, ".");
 
-  const size_t key_old_size = _object_key.size() + 1;  // erase to _object_key.
+  const size_t key_old_size = key.size();
   for (velox::vector_size_t row_idx = 0; row_idx < num_rows; ++row_idx) {
     row_key.clear();
     primary_key::Create(*input, row_idx, row_key);
+    absl::StrAppend(&key, row_key);
+    VELOX_CHECK(_transaction.GetKeyLock(&_cf, key, false, true).ok(),
+                "Failed to acquire row lock for table {}", _object_key);
```
mkornaukhov, line 2019: When lock will be released?

server/connector/serenedb_connector.hpp
```diff
@@ -412,9 +412,12 @@ class SereneDBConnector final : public velox::connector::Connector {
 
     auto* transaction = serene_insert_handle.GetTransaction();
     if (!transaction) {
+      rocksdb::TransactionOptions txn_options;
+      // we will lock rows explicitly
+      txn_options.skip_concurrency_control = true;
       serene_insert_handle.SetLocalTransaction(
         std::unique_ptr<rocksdb::Transaction>(
-          _db.BeginTransaction(rocksdb::WriteOptions{})));
+          _db.BeginTransaction(rocksdb::WriteOptions{}, txn_options)));
```
mkornaukhov, line 420: Previously on `_transaction.Put()` in writing a cell we acquired lock explicitly, for now rocksdb won't try to acquire  a lock inside `_transaction.Put()`, right? And exactly these lines are the reason?

server/connector/data_sink.cpp
Dronplane: Reason is that type check was set to run only in debug builds. Without maybe_unused release build will fail.

server/connector/data_sink.cpp
Dronplane: When transaction is committed/rollbacked. this is done by rocksdb internally.

server/connector/serenedb_connector.hpp
Dronplane: Yes

Dronplane:
> Why do we need such a granular lock? Why can't we store 1 lock for all columns (instead of per column) of the same table at all?

We store exactly one lock per row (e.g. for all columns)

PR #45, opened 2025-12-10
Message buffer small fixes

codeworse:


libs/basics/message_sequence_view.cpp
```diff
@@ -43,6 +43,7 @@ SequenceView::ConstIterator& SequenceView::ConstIterator::operator++() {
 boost::asio::const_buffer SequenceView::ConstIterator::operator*() const {
   SDB_ASSERT(_it != _end);
   SDB_ASSERT(_it.in_chunk == _it.chunk->GetBegin());
+  SDB_ASSERT(_it.in_chunk <= _it.chunk->GetEnd());
```
MBkkt: it's data race for last chunk

libs/basics/message_buffer.cpp
```diff
@@ -122,6 +116,8 @@ void Buffer::FlushDone() {
 
 void Buffer::SendData() const {
   SDB_ASSERT(_socket_end);
+  SDB_ASSERT(_head != _socket_end.chunk ||
+             _head->GetBegin() < _socket_end.in_chunk);
   SequenceView data{BufferOffset{_head, _head->GetBegin()}, _socket_end};
   if (!_send_callback || data.Empty()) {
```
MBkkt, line 122: assert above doesn't corresponds to data.Empty

codeworse: yes, it's more like sanity check for _head and _socket_end(_head should be always <= _socket). Although, current assert will raise only in case, when _head and _socket_end will be in the same block, but I think it would nice to have at least some check

PR #49, opened 2025-12-11
perf: Less condtions in Message Buffer

MBkkt:


libs/basics/message_buffer.cpp
```diff
@@ -29,7 +29,9 @@ Buffer::Buffer(size_t min_growth, size_t max_growth, size_t flush_size,
   : _send_callback{std::move(send_callback)},
     _flush_size{flush_size},
     _max_growth{max_growth},
-    _growth{min_growth} {}
+    _growth{min_growth} {
+  _head = _tail = CreateChunk(0);
```
codeworse, line 33: is there assert in chunk for 0 capacity?

MBkkt: there's no such assert

PR #50, opened 2025-12-12
Add Postgres unknown type

mkornaukhov:


server/pg/sql_analyzer_velox.cpp
```diff
@@ -182,8 +182,15 @@ velox::TypePtr ResolveFunction(const std::string& function_name,
 
 std::vector<velox::TypePtr> GetExprsTypes(
   std::span<const lp::ExprPtr> arg_exprs) {
-  return arg_exprs |
-         std::views::transform([](const auto& arg) { return arg->type(); }) |
+  return arg_exprs | std::views::transform([](const auto& arg) {
+           const auto& arg_type = arg->type();
+           // Literal
+           if (arg_type == velox::VARCHAR() &&
+               arg->kind() == lp::ExprKind::kConstant) {
```
MBkkt: It's incorrect
Because in such case `select 1 || '1'` it will produce `'11'`

```suggestion
           // Literal
           if (arg->kind() == lp::ExprKind::kConstant && arg->type() one of list) {
```

list
```
 {velox::VARCHAR(), velox::TINYINT(), velox::SMALLINT(), velox::INTEGER(),
       velox::BIGINT(), velox::HUGEINT(), velox::REAL(), velox::DOUBLE(),
       velox::BOOLEAN()}
```

mkornaukhov: It isn't true, it's concat operator specific behavior

server/pg/sql_analyzer_velox.cpp
```diff
@@ -3857,6 +3911,9 @@ lp::ExprPtr SqlAnalyzer::ResolveVeloxFunctionAndInferArgsCommonType(
       ERR_MSG("function ", ToPgFunctionString(name, args), " does not exist"));
   }
 
+  // TODO rewrite it recursive and make separate function for this
+  // type = FixupReturnType(type);
+
```
MBkkt: remove?

server/pg/sql_analyzer_velox.cpp
```diff
@@ -4077,6 +4134,10 @@ lp::ExprPtr SqlAnalyzer::ProcessTypeCast(State& state, const TypeCast& expr) {
 
   auto arg = ProcessExprNodeImpl(state, expr.arg);
 
+  if (arg->type() == PG_UNKNOWN()) {
+    arg = MakeCast(velox::VARCHAR(), arg);
+  }
```
MBkkt: why?

server/pg/sql_analyzer_velox.cpp
```diff
@@ -829,7 +875,14 @@ class SqlAnalyzer {
...
     for (size_t i = 0; i < coercions.size(); ++i) {
-      if (coercions[i]) {
+      if (args[i]->type() == PG_UNKNOWN()) {
+        if (!coercions[i] || coercions[i] == velox::VARCHAR()) {
+          // do nothing
+        } else {
+          args[i] = MakeCast(velox::VARCHAR(), args[i]);
+        }
+        // prevent varchar -> varchar coercion
+      } else if (coercions[i] && coercions[i] != args[i]->type()) {
```
MBkkt: I think this should be like this
```suggestion
      } else if (coercions[i]) {
        SDB_ASSERT(!coercions[i].equals(args[i]->type())
```

server/pg/sql_analyzer_velox.cpp
```diff
@@ -181,17 +181,63 @@ velox::TypePtr ResolveFunction(const std::string& function_name,
 }
 
 std::vector<velox::TypePtr> GetExprsTypes(
-  std::span<const lp::ExprPtr> arg_exprs) {
-  return arg_exprs |
-         std::views::transform([](const auto& arg) { return arg->type(); }) |
+  std::span<const lp::ExprPtr> arg_exprs, bool use_pg_unknown = true) {
```
MBkkt, line 185:
```suggestion
  std::span<const lp::ExprPtr> arg_exprs) {
```

server/pg/sql_analyzer_velox.cpp
```diff
@@ -181,17 +181,63 @@ velox::TypePtr ResolveFunction(const std::string& function_name,
...
+  const auto& params = ret_type->parameters();
+  const auto params_cnt = params.size();
+
+  std::vector<velox::TypeParameter> new_params;
+  bool changed = false;
+
+  new_params.reserve(params_cnt);
+  for (size_t i = 0; i < params_cnt; ++i) {
+    if (params[i].kind != velox::TypeParameterKind::kType) {
+      continue;
```
MBkkt: params should be filled in such case

server/pg/pg_feature.cpp
```diff
@@ -125,6 +136,15 @@ velox::AllowedCoercions AllowedCoercions() {
   add(velox::REAL(), {velox::DOUBLE()});
   add(velox::DATE(), {velox::TIMESTAMP()});
 
+  // add_same_cost(PG_UNKNOWN(),
+  //               {velox::VARCHAR()},
+  //               kNullCoercionCost + 1);
```
MBkkt:
```suggestion
```

server/pg/sql_analyzer_velox.cpp
mkornaukhov: No, coercion[i] may be equal to args[i]->type(), `varchar -> varchar`.
In case of `GetExprsTypes()` returns `PG_UNKNOWN()` for argument, binding may return coercion from `PG_UNKNOWN()` into `velox::VARCHAR()`, but real type (`args[i]->type()`) is still `varchar`.

server/pg/sql_analyzer_velox.cpp
mkornaukhov: It's better with default argument

server/pg/sql_analyzer_velox.cpp
mkornaukhov: Fixed

server/pg/pg_feature.cpp
```diff
@@ -125,6 +136,12 @@ velox::AllowedCoercions AllowedCoercions() {
   add(velox::REAL(), {velox::DOUBLE()});
   add(velox::DATE(), {velox::TIMESTAMP()});
 
+  add_same_cost(PG_UNKNOWN(),
```
Dronplane: Should we also have cast cost from UNKNOWN to DATE/DATETIME? Maybe more expensive.

server/pg/sql_analyzer_velox.cpp
```diff
@@ -181,17 +181,64 @@ velox::TypePtr ResolveFunction(const std::string& function_name,
...
+    return velox::VARCHAR();
+  }
+  const auto& params = ret_type->parameters();
+  const auto params_cnt = params.size();
+
+  std::vector<velox::TypeParameter> new_params;
+  bool changed = false;
+
+  new_params.reserve(params_cnt);
+  for (size_t i = 0; i < params_cnt; ++i) {
```
Dronplane: NIT
Could we use range loop here instead of index variable?

server/pg/pg_feature.cpp
mkornaukhov: Which UNKNOWN do you mean?
Let's use velox::UNKNOWN or PG_UNKNOWN to avoid confusion.
For velox::UNKNOWN (which is inhabited with NULL only) it's done by default.
For PG_UNKNOWN it make sense, I'll try to add.

Dronplane: I see. I meant PG_UNKNOWN then

server/pg/sql_analyzer_velox.cpp
```diff
@@ -181,17 +181,64 @@ velox::TypePtr ResolveFunction(const std::string& function_name,
...
+    return ret_type;
+  }
+  return velox::getType(ret_type->name(), std::move(new_params));
+}
 velox::TypePtr ResolveFunction(const std::string& function_name,
                                std::span<const lp::ExprPtr> arg_exprs,
                                std::vector<velox::TypePtr>* arg_coercions) {
-  return ResolveFunction(function_name, GetExprsTypes(arg_exprs),
-                         arg_coercions);
+  if (const auto ret_type = ResolveFunction(
```
Dronplane, line 234: Looks like calling `ResolveFunction` might end up calling one of implementations of 
FunctionCallToSpecialForm::resolvewType. Which is annotated with `Throws if the input Types do not match what's expected for the SpecialForm` In that case we will have no chance to execute second call to ResolveFunction below. Is this expected and exception indeed mean there is no point of running second check?

server/pg/pg_feature.cpp
mkornaukhov: I've added, but it's impossible to test it for now. Supporting coercions for user sql function would allow us to do that.

server/pg/sql_analyzer_velox.cpp
mkornaukhov: It's required for something like `least<T>(Variadic<T>) -> T`. A lot of overloadings, so try with explicit varchar next time.

Dronplane: I see. But my point here is that ResolveFunction can throw on first run

mkornaukhov: This case is for functions overloaded with multiple types, so first call return nullptr only in case of multiple appropriate signatures selected.
Special forms are not such cases.

So both calls will throw or none, so it's OK.

server/pg/pg_feature.cpp
mkornaukhov: #64

PR #51, opened 2025-12-12
Use binary numerics as key parts in connector

Dronplane:
Change RocksDB key format to use Table ObjectId + Column id as BE encoded numerics.
Add utility functions for key building

### DIFF
--- a/server/connector/data_sink.cpp
+++ b/server/connector/data_sink.cpp
@@ -102,7 +102,7 @@ void RocksDBDataSink::appendData(velox::RowVectorPtr input) {
   velox::IndexRange all_rows(0, input->size());
   [[maybe_unused]] const auto& input_type = input->type()->asRow();
   for (const auto& key : _row_keys) {
-    table_key.resize(parent_size);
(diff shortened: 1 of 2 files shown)


server/connector/data_sink.cpp
```diff
@@ -85,26 +86,30 @@ RocksDBDataSink::RocksDBDataSink(
...
-  std::string name = _object_key;
-  const auto parent_size = name.size();
+  std::string table_key = key_utils::PrepareTableKey(_object_key);
+  const auto parent_size = table_key.size();
   velox::IndexRange all_rows(0, input->size());
   [[maybe_unused]] const auto& input_type = input->type()->asRow();
-  std::string lock_key = absl::StrCat(_object_key, ".");
-  const auto lock_key_base_size = lock_key.size();
   for (const auto& key : _row_keys) {
-    lock_key.resize(lock_key_base_size);
```
gnusi, line 100: NIT: `basics::StrResize` will make it slightly faster

PR #59, opened 2025-12-13
feat: Report count of affected rows correctly

MBkkt:


server/pg/pg_comm_task.cpp
```diff
@@ -1006,6 +1006,16 @@ void PgSQLCommTaskBase::SendBatch(const velox::RowVectorPtr& batch) {
     return;
   }
 
+  // We need to check output type name, because explain command
+  if (portal.stmt->query->IsDML() && output_type->nameOf(0) == "rows") {
```
codeworse, line 1013: I don't quite get the comment here

MBkkt: Fixed

PR #67, opened 2025-12-17
fix: Use ForeignId for distributeShardsLike serialization

MBkkt:
Proper fix should include rollback recent changes made by @pashandor789 

We don't want to use stateful struct for serialization


MBkkt:
This doesn't fix schema, so I think we will close this in favor of proper fix https://github.com/serenedb/serenedb/pull/68

PR #71, opened 2025-12-17
misc: Update deps

MBkkt:
mainly because of new serenedb:axiom version

VeloxHistory => NoopHistory, because we don't use it, and VeloxHistory is quite heavy

### DIFF
--- a/server/pg/sql_collector.h
+++ b/server/pg/sql_collector.h
@@ -73,7 +73,7 @@ class Objects : public irs::memory::Managed {
     // TODO(mbkkt) Maybe remove this and instead make catalog::Table be able
     // to implement connector::Table without allocation.
     // This probably requires changing axiom::connector::Table.
-    void Ensure() const;
+    void EnsureTable() const;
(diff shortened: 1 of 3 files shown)


server/pg/sql_collector.h
```diff
@@ -67,6 +69,12 @@ class Objects : public irs::memory::Managed {
   struct ObjectData {
     AccessType type = AccessType::None;
     std::shared_ptr<catalog::SchemaObject> object;
+
+    // TODO(mbkkt) Maybe remove this and instead make catalog::Table be able
+    // to implement connector::Table without allocation.
+    // This probably requires changing axiom::connector::Table.
+    void Ensure() const;
```
Dronplane: NIT: Maybe `EnsureTable` to be more specific?

MBkkt: ok, https://github.com/serenedb/serenedb/pull/71/changes/9130b384dc12cc112ab8c5c6745532a87ef55c55

PR #74, opened 2025-12-17
Support tables without primary keys

mkornaukhov:
This PR add support of tables without primary keys.

We just add fake column with uniquely generated values (they are of type `uint64_t`, `BIGINT` in terms of velox). For update and delete, we provide this column from data source, but for select queries we do not do that.  This column is not physically stored as rocksdb items, we generate it from the rocksdb's keys, see `ReadFakeColumn()`.

### DIFF
--- a/server/catalog/table.cpp
+++ b/server/catalog/table.cpp
@@ -23,7 +23,6 @@
 
 #include <absl/algorithm/container.h>
 #include <absl/strings/str_cat.h>
-#include <velox/type/Type.h>
 #include <vpack/builder.h>
 #include <vpack/iterator.h>
 #include <vpack/slice.h>
--- a/server/connector/key_utils.hpp
+++ b/server/connector/key_utils.hpp
@@ -24,7 +24,6 @@
 #include <velox/vector/BaseVector.h>
 #include <velox/vector/ComplexVector.h>
 
-#include "basics/logger/logger.h"
 #include "catalog/identifiers/object_id.h"
 #include "catalog/table_options.h"
 #include "connector/primary_key.hpp"
--- a/server/connector/serenedb_connector.hpp
+++ b/server/connector/serenedb_connector.hpp
@@ -174,8 +174,8 @@ class RocksDBTable final : public axiom::connector::Table {
       columns.push_back(serenedb_column.get());
 
       _column_map.emplace(generated_pk_name, serenedb_column.get());
+      order_columns.push_back(serenedb_column.get());
       _column_handles.push_back(std::move(serenedb_column));
-      order_columns.push_back(findColumn(generated_pk_name));
       _pk_type = velox::ROW({std::move(generated_pk_name)}, {velox::BIGINT()});
       sort_order.resize(1, axiom::connector::SortOrder{.isAscending = true,
                                                        .isNullsFirst = false});
--- a/server/pg/sql_analyzer_velox.cpp
+++ b/server/pg/sql_analyzer_velox.cpp
@@ -1316,16 +1316,10 @@ void SqlAnalyzer::ProcessInsertStmt(State& state, const InsertStmt& stmt) {
 
   const auto& table = basics::downCast<catalog::Table>(logical_object);
   const auto& table_type = *table.RowType();
-  size_t table_columns = table_type.size();
-  if (!table.PKColumns().empty() &&
-      table.PKColumns()[0] == catalog::Column::kGeneratedPKId) {
-    table_columns--;
-  }
-
   std::vector<std::string> column_names;
   std::vector<lp::ExprPtr> column_exprs;
-  column_names.reserve(table_columns);
-  column_exprs.reserve(table_columns);
+  column_names.reserve(table_type.size());
+  column_exprs.reserve(table_type.size());
 
   const uint32_t explicit_columns = list_length(stmt.cols);
   const auto& input_type = *state.root->outputType();
@@ -1334,7 +1328,7 @@ void SqlAnalyzer::ProcessInsertStmt(State& state, const InsertStmt& stmt) {
                     ERR_MSG("INSERT has more target columns than expressions"),
                     CURSOR_POS(ErrorPosition(ExprLocation(&stmt))));
   }
-  if ((stmt.cols ? explicit_columns : table_columns) < input_type.size()) {
+  if ((stmt.cols ? explicit_columns : table_type.size()) < input_type.size()) {
     THROW_SQL_ERROR(ERR_CODE(ERRCODE_SYNTAX_ERROR),
                     ERR_MSG("INSERT has more expressions than target columns"),
                     CURSOR_POS(ErrorPosition(ExprLocation(&stmt))));
@@ -1454,6 +1448,7 @@ void SqlAnalyzer::ProcessUpdateStmt(State& state, const UpdateStmt& stmt) {
   column_exprs.reserve(pk_type.size() + list_length(stmt.targetList));
 
   const auto& output_type = state.root->outputType();
+
   for (const auto& [name, type] :
        std::ranges::views::zip(pk_type.names(), pk_type.children())) {
     auto column = state.resolver.Resolve(output_type, name);
--- a/server/catalog/table.h
+++ b/server/catalog/table.h
@@ -23,7 +23,6 @@
 
 #include <velox/type/Type.h>
 
-#include "basics/assert.h"
 #include "basics/fwd.h"
 #include "catalog/identifiers/identifier.h"
 #include "catalog/identifiers/object_id.h"


server/connector/serenedb_connector.hpp
```diff
@@ -153,7 +153,13 @@ class RocksDBTable final : public axiom::connector::Table {
     catalog::Column::Id col_id = 0;
     for (const auto& [name, type] : std::views::zip(
            collection.RowType()->names(), collection.RowType()->children())) {
-      auto column = std::make_unique<SereneDBColumn>(name, type, col_id++);
+      catalog::Column::Id id{};
+      if (name == catalog::Column::kFakeName) {
```
Dronplane: What if someone will create a table with column named "fake"?
Like `CREATE TABLE foo (fake INTEGER PRIMARY KEY)` ?

server/connector/key_utils.hpp
```diff
@@ -50,6 +51,15 @@ void MakeColumnKey(const velox::RowVectorPtr& input,
   std::memcpy(key_buffer.data() + sizeof(catalog::Column::Id), object_id.data(),
               sizeof(ObjectId));
   primary_key::Create(*input, pk_columns, row_idx, key_buffer);
+
+  if (pk_columns.empty()) {
```
Dronplane: primary_key::Create call looks redundant in case we have empty pk_columns

server/connector/serenedb_connector.hpp
```diff
@@ -186,7 +192,6 @@ class RocksDBTable final : public axiom::connector::Table {
   std::vector<velox::connector::ColumnHandlePtr> rowIdHandles(
     axiom::connector::WriteKind kind) const final {
     SDB_ASSERT(_pk_type);
-    SDB_ASSERT(!_pk_type->children().empty());
```
Dronplane, line 189: just `return {}` for empty children?

server/pg/sql_analyzer_velox.cpp
```diff
@@ -740,7 +743,8 @@ class SqlAnalyzer {
                     const SqlQueryView& view, const RangeVar* node);
   State ProcessTable(State* parent, std::string_view schema_name,
                      std::string_view table_name,
-                     const Objects::ObjectData& object, const RangeVar* node);
+                     const Objects::ObjectData& object, const RangeVar* node,
+                     bool show_fake_column = false);
```
Dronplane: NIT - maybe name it more explicit like "implicit_pk_column" or smth. So it would be clear for the reader what is this "fake" column for.

tests/sqllogic/any/pg/simple/basic_dml.test
```diff
@@ -120,3 +120,76 @@ INSERT INTO cast_table VALUES ('qqq', '2025-02-01 12:59:59',  '9');
...
+42	300	aboba
+
+
+statement count 1
+INSERT INTO no_pk VALUES (123, 888, 'new str');
+
+statement count 3
+INSERT INTO no_pk VALUES (111, 222, 'same'), (111, 222, 'same'), (111, 222, 'same');
+
+query
```
Dronplane: order might be unstable on postgresql runs - maybe add ORDER BY clause?

server/connector/data_sink.cpp
```diff
@@ -103,6 +104,11 @@ void RocksDBDataSink::appendData(velox::RowVectorPtr input) {
   _keys_buffers.clear();
   _keys_buffers.reserve(num_rows);
 
+  if (!_skip_primary_key_columns &&
+      _column_ids.back() == catalog::Column::kFakeId) {
+    _key_childs = {};
```
Dronplane: Shouldn't it be just assert that key_childs is empty? Connector uses `rowIdHandles` for this  and it should be empty anyway

server/connector/key_utils.hpp
mkornaukhov: Fixed

server/connector/serenedb_connector.hpp
mkornaukhov: Use `"@fake"` now. It's invalid column name in SQL and PostgreSQL.

server/pg/sql_analyzer_velox.cpp
mkornaukhov: Fixed

tests/sqllogic/any/pg/simple/basic_dml.test
mkornaukhov: Fixed

server/connector/serenedb_connector.hpp
mkornaukhov: Children are always not empty. I return `{}` for case of insert and fake primary column

server/connector/serenedb_connector.hpp
Dronplane: That is still possible 
```
postgres=> CREATE TABLE Foo2 ("@fake" INTEGER Primary KEY);
CREATE TABLE
postgres=> select * FROM Foo2;
 @fake 
-------
(0 rows)
```
I guess we should rely on column id somehow

server/connector/key_utils.hpp
```diff
@@ -49,7 +50,17 @@ void MakeColumnKey(const velox::RowVectorPtr& input,
...
-  primary_key::Create(*input, pk_columns, row_idx, key_buffer);
+
+  if (!pk_columns.empty()) {
+    primary_key::Create(*input, pk_columns, row_idx, key_buffer);
+  } else {
+    auto generated_pk = RevisionId::create().id();
+    const auto old_size = key_buffer.size();
+    basics::StrResize(key_buffer, old_size + sizeof(generated_pk));
+    absl::big_endian::Store64(key_buffer.data() + old_size, generated_pk);
+    key_buffer[old_size] = static_cast<uint8_t>(key_buffer[old_size]) ^ 0x80;
```
Dronplane: I guess we do not need this trick here (and in read inside data source) as generated keys are all unsigned anyway.

server/connector/data_sink.cpp
mkornaukhov: Just removed this code

server/connector/key_utils.hpp
mkornaukhov: We read primary keys from column of `bigint` in case of update. It does flipping, see `AppendKeyValue()`. I've added this to be consistent with this behaviour.

server/connector/serenedb_connector.hpp
mkornaukhov: But we may use empty literal that is not valid column name.
What do you think about it?

Dronplane: Hm yes. Empty literal should work 👍

server/connector/key_utils.hpp
Dronplane: I see. So it is to avoid additional check for fake column id during update?

mkornaukhov: Yes, I think this check would be awful

server/connector/serenedb_connector.hpp
mkornaukhov: There are some asserts on it, so just fixed it to deal with id

server/connector/key_utils.hpp
Dronplane: Ok. It would be good if we move this reading/writing in some helper inside primary_key.
Like extract writing logic of primitive type from AppendKeyValue. 
Like have 
```
template<typename T>
AppendKeyValue (std::string& key, const T& value)
```
And call it from current `AppendKeyValue` after extracting value from velox vector and here for generated value.
Or at least put a TODO for that.

server/connector/data_source.cpp
```diff
@@ -101,11 +104,20 @@ std::optional<velox::RowVectorPtr> RocksDBDataSource::next(
...
     for (velox::column_index_t col_idx = 0; col_idx < num_columns; ++col_idx) {
-      basics::StrResize(key, key_old_size);
-      key_utils::AppendColumnKey(key, _column_ids[col_idx]);
+      basics::StrResize(key, table_prefix_size);
+      const auto column_id = _column_ids[col_idx];
+
+      size_t read_col_idx = col_idx;
+      if (column_id == catalog::Column::kFakeId) {
+        // At least 1 column non-fake required
+        SDB_ASSERT(col_idx >= 1 || col_idx + 1 < num_columns);
```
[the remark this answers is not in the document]
mkornaukhov: It's impossible in our code for now

server/connector/data_source.cpp
```diff
@@ -101,11 +104,20 @@ std::optional<velox::RowVectorPtr> RocksDBDataSource::next(
...
+  const auto table_prefix_size = key.size();
   if (num_columns) {
     for (velox::column_index_t col_idx = 0; col_idx < num_columns; ++col_idx) {
-      basics::StrResize(key, key_old_size);
-      key_utils::AppendColumnKey(key, _column_ids[col_idx]);
+      basics::StrResize(key, table_prefix_size);
+      const auto column_id = _column_ids[col_idx];
+
+      size_t read_col_idx = col_idx;
+      if (column_id == catalog::Column::kFakeId) {
```
Dronplane: BTW is it possible now to have implicit pk column not last? Analyzer always adds it to the end. It could not be requested by user - so maybe it is just an assert and we can safely assume that FakeId column is the last one and always use "col_idx - 1" ?

mkornaukhov: This is not correct. During updates it reads only primary key and, for example, column that are in filter. It's possible that the first column is the fake one and filtering columns are after it.

server/connector/key_utils.hpp
mkornaukhov: Fixed

server/connector/serenedb_connector.hpp
```diff
@@ -150,13 +151,12 @@ class RocksDBTable final : public axiom::connector::Table {
...
       axiom::connector::SortOrder{.isAscending = true, .isNullsFirst = false});
     // TODO(Dronplane) take it from the catalog when available
-    catalog::Column::Id col_id = 0;
-    for (const auto& [name, type] : std::views::zip(
-           collection.RowType()->names(), collection.RowType()->children())) {
-      auto column = std::make_unique<SereneDBColumn>(name, type, col_id++);
-      columns.push_back(column.get());
-      _column_map.emplace(name, column.get());
-      _column_handles.push_back(std::move(column));
+    for (const auto& catalog_column : collection.Columns()) {
```
Dronplane, line 155: I guess TODO on line above could be removed now?

server/connector/serenedb_connector.hpp
```diff
@@ -186,7 +186,12 @@ class RocksDBTable final : public axiom::connector::Table {
   std::vector<velox::connector::ColumnHandlePtr> rowIdHandles(
     axiom::connector::WriteKind kind) const final {
     SDB_ASSERT(_pk_type);
-    SDB_ASSERT(!_pk_type->children().empty());
+
+    if (kind == axiom::connector::WriteKind::kInsert && _pk_type->size() == 1 &&
```
Dronplane: "_pk_type->size() == 1" should be an assert. 
Or we somehow expect generated Id as part of more complex PK?

server/connector/serenedb_connector.hpp
mkornaukhov: Removed

server/connector/serenedb_connector.hpp
mkornaukhov: No, it may be only single.
Fixed

server/catalog/table.cpp
MBkkt: absl::c_find_if

server/catalog/table.cpp
```diff
@@ -127,7 +133,7 @@ Table::Table(TableOptions&& options, ObjectId database_id)
     _columns{std::move(options.columns)},
     _pk_columns{std::move(options.pkColumns)},
     _pk_type{BuildPkType(_columns, _pk_columns)},
-    _row_type{BuildRowType(_columns)},
+    _row_type{BuildRowType(_columns, _pk_columns)},
```
MBkkt:
```suggestion
    _row_type{BuildRowType(_columns)},
```

server/catalog/table_options.cpp
```diff
@@ -364,6 +364,12 @@ Result MakeTableOptions(CreateTableRequest&& request, ObjectId database_id,
   options.shardKeys = std::move(request.shardKeys);
   options.columns = std::move(request.columns);
   options.pkColumns = std::move(request.pkColumns);
+  if (options.pkColumns.empty()) {
+    options.pkColumns = {Column::kFakeId};
+    options.columns.push_back(Column{.id = Column::kFakeId,
+                                     .type = Column::kFakeType,
+                                     .name = std::string{Column::kFakeName}});
+  }
```
MBkkt: move this to commands/create_table

server/catalog/table_options.h
```diff
@@ -144,6 +145,10 @@ struct Column {
 
   using Id = uint64_t;
 
+  static constexpr Id kFakeId = std::numeric_limits<Id>::max();
+  static constexpr std::string_view kFakeName = "@fake";
```
MBkkt: `sdb_generated_pk`

Add test for table with same column name

server/catalog/table_options.h
```diff
@@ -144,6 +145,10 @@ struct Column {
...
+  static inline velox::TypePtr kFakeType = velox::BIGINT();
```
MBkkt: 1. fake? maybe generated primary key?
2. reserve some column ids (for an example 1e6) for special columns

MBkkt:
```suggestion
```
maybe we can remove this?

server/connector/data_sink.cpp
```diff
@@ -125,14 +126,18 @@ void RocksDBDataSink::appendData(velox::RowVectorPtr input) {
     // child lookup by name as it is also a necessary check.
     // Even keys during UPDATE should be validated as it is important for
     // RocksDB key generation
+    _column_id = _column_ids[i];
+    if (_column_id == catalog::Column::kFakeId) {
+      continue;
+    }
     VELOX_DCHECK(_row_type->findChild(input_type.nameOf(i))
```
MBkkt: move this below debug assert

server/connector/key_utils.hpp
```diff
@@ -49,7 +50,15 @@ void MakeColumnKey(const velox::RowVectorPtr& input,
...
   std::memcpy(key_buffer.data() + sizeof(catalog::Column::Id), object_id.data(),
               sizeof(ObjectId));
-  primary_key::Create(*input, pk_columns, row_idx, key_buffer);
+
+  if (!pk_columns.empty()) {
+    primary_key::Create(*input, pk_columns, row_idx, key_buffer);
+  } else {
+    // Signed type is required as it will be stored in velox vector
+    auto generated_pk = std::bit_cast<int64_t>(RevisionId::create().id());
+    primary_key::AppendIntegral(key_buffer, generated_pk);
```
MBkkt: todo + issue about we want u64

server/connector/primary_key.hpp
```diff
@@ -45,4 +47,27 @@ void Create(const velox::RowVector& data,
...
+  absl::big_endian::Store(key.data() + base_size, value);
+  key[base_size] = static_cast<uint8_t>(key[base_size]) ^ 0x80;
+}
+
+template<typename T>
+T ReadIntegral(const char* buf) {
+  SDB_ASSERT(std::is_signed_v<T>,
+             "Cannot correctly read unsigned value due to sign bit flipping");
+  using UnsignedT = typename std::make_unsigned<T>::type;
+  constexpr int kBits = sizeof(T) * CHAR_BIT;
```
MBkkt:
```suggestion
  static constexpr int kBits = sizeof(T) * CHAR_BIT;
```

server/connector/primary_key.hpp
```diff
@@ -45,4 +47,27 @@ void Create(const velox::RowVector& data,
...
+  basics::StrAppend(key, sizeof(T));
+  absl::big_endian::Store(key.data() + base_size, value);
+  key[base_size] = static_cast<uint8_t>(key[base_size]) ^ 0x80;
+}
+
+template<typename T>
+T ReadIntegral(const char* buf) {
+  SDB_ASSERT(std::is_signed_v<T>,
+             "Cannot correctly read unsigned value due to sign bit flipping");
+  using UnsignedT = typename std::make_unsigned<T>::type;
```
MBkkt:
```suggestion
  using Unsigned = std::make_unsigned_t<T>;
```

server/connector/primary_key.hpp
```diff
@@ -45,4 +47,27 @@ void Create(const velox::RowVector& data,
...
+  SDB_ASSERT(std::is_signed_v<T>,
+             "Cannot correctly store unsigned value due to sign bit flipping");
+  const auto base_size = key.size();
+  basics::StrAppend(key, sizeof(T));
+  absl::big_endian::Store(key.data() + base_size, value);
+  key[base_size] = static_cast<uint8_t>(key[base_size]) ^ 0x80;
+}
+
+template<typename T>
+T ReadIntegral(const char* buf) {
```
MBkkt: ReadSigned

server/connector/primary_key.hpp
```diff
@@ -45,4 +47,27 @@ void Create(const velox::RowVector& data,
 void Create(const velox::RowVector& data, velox::vector_size_t idx,
             std::string& key);
 
+template<typename T>
+void AppendIntegral(std::string& key, T value) {
```
MBkkt: AppendSigned

server/pg/sql_analyzer_velox.cpp
```diff
@@ -1312,10 +1316,16 @@ void SqlAnalyzer::ProcessInsertStmt(State& state, const InsertStmt& stmt) {
 
   const auto& table = basics::downCast<catalog::Table>(logical_object);
   const auto& table_type = *table.RowType();
+  size_t table_columns = table_type.size();
+  if (!table.PKColumns().empty() &&
+      table.PKColumns()[0] == catalog::Column::kFakeId) {
+    table_columns--;
+  }
```
MBkkt: Lets do this differently:
Instead of modifying table definition, lets add additional column in update/delete for scan (analyzer) and in insert data_sink.

server/catalog/table_options.h
Dronplane: BTW good point - it might trigger name uniqueness assertion in analyzer

server/catalog/table_options.cpp
mkornaukhov: is not needed as table description will not have it

server/catalog/table_options.h
mkornaukhov: It's become useful now

server/catalog/table.cpp
mkornaukhov: Fixed

server/catalog/table.cpp
mkornaukhov: Fixed

server/catalog/table_options.h
mkornaukhov: Fixed and added test

server/catalog/table_options.h
mkornaukhov: 1. Yep
2. Done

server/connector/data_sink.cpp

server/connector/key_utils.hpp
mkornaukhov: Done

server/connector/primary_key.hpp
mkornaukhov: Fixed

server/connector/primary_key.hpp
mkornaukhov: Fixed

server/connector/primary_key.hpp
mkornaukhov: Fixed

server/connector/primary_key.hpp
mkornaukhov: Fixed

server/connector/key_utils.hpp
mkornaukhov: https://github.com/serenedb/serenedb/issues/94

server/pg/sql_analyzer_velox.cpp
mkornaukhov: Done

server/catalog/table.cpp
```diff
@@ -23,6 +23,7 @@
 
 #include <absl/algorithm/container.h>
 #include <absl/strings/str_cat.h>
+#include <velox/type/Type.h>
```
MBkkt: why do you add random includes? it's not first time when I noticed this.
You probably need to specify this argument for clangd:
```
--header-insertion=never
```

server/catalog/table_options.h
```diff
@@ -144,6 +145,34 @@ struct Column {
 
   using Id = uint64_t;
 
+  static constexpr Id kMaxRealId = 1'000'000;
```
MBkkt: It's not correct, should be something like

```suggestion
  static constexpr Id kMaxRealId = std::numeric_limits<Id>::max() - 1'000'000;
```

server/catalog/table_options.h
```diff
@@ -144,6 +145,34 @@ struct Column {
...
+
+  static constexpr Id kGeneratedPKId = kMaxRealId + 1;
+  static inline velox::TypePtr kGeneratedPKType = velox::BIGINT();
```
MBkkt:
```suggestion
```
remove this plz

server/catalog/table_options.h
```diff
@@ -144,6 +145,34 @@ struct Column {
...
+
+  static std::string GeneratePKName(std::span<const std::string> column_names) {
```
MBkkt: move this to cpp file

server/connector/data_sink.cpp
```diff
@@ -128,11 +129,15 @@ void RocksDBDataSink::appendData(velox::RowVectorPtr input) {
...
+    if (_column_id == catalog::Column::kGeneratedPKId) {
+      continue;
+    }
 
     if (_skip_primary_key_columns && i < _key_childs.size()) {
       continue;
     }
-    _column_id = _column_ids[i];
+
     WriteColumn(input->childAt(i), folly::Range{&all_rows, 1}, {});
```
MBkkt: NIT
```cpp
    if (_skip_primary_key_columns && i < _key_childs.size()) {
      continue;
    }
    _column_id = _column_ids[i];
    if (_column_id != catalog::Column::kGeneratedPKId) {
      WriteColumn(input->childAt(i), folly::Range{&all_rows, 1}, {});
```

server/connector/data_source.cpp
```diff
@@ -101,11 +106,20 @@ std::optional<velox::RowVectorPtr> RocksDBDataSource::next(
...
-      basics::StrResize(key, key_old_size);
-      key_utils::AppendColumnKey(key, _column_ids[col_idx]);
+      basics::StrResize(key, table_prefix_size);
+      const auto column_id = _column_ids[col_idx];
+
+      size_t read_col_idx = col_idx;
+      if (column_id == catalog::Column::kGeneratedPKId) {
+        SDB_ASSERT(col_idx > 0 || col_idx + 1 < num_columns,
+                   "Table without columns are not supported");
+        read_col_idx = (col_idx == 0) ? col_idx + 1 : col_idx - 1;
```
MBkkt: Why we can't make it same?

server/connector/key_utils.hpp
```diff
@@ -49,7 +51,15 @@ void MakeColumnKey(const velox::RowVectorPtr& input,
   basics::StrResize(key_buffer, sizeof(catalog::Column::Id) + sizeof(ObjectId));
   std::memcpy(key_buffer.data() + sizeof(catalog::Column::Id), object_id.data(),
               sizeof(ObjectId));
-  primary_key::Create(*input, pk_columns, row_idx, key_buffer);
+
+  if (!pk_columns.empty()) {
+    primary_key::Create(*input, pk_columns, row_idx, key_buffer);
+  } else {
+    // TODO: make unsigned when such types will be supported in Velox
+    auto generated_pk = std::bit_cast<int64_t>(RevisionId::create().id());
```
MBkkt:
```suggestion
    const auto generated_pk = std::bit_cast<int64_t>(RevisionId::create().id());
```

server/connector/primary_key.hpp
```diff
@@ -45,4 +47,27 @@ void Create(const velox::RowVector& data,
...
+  SDB_ASSERT(std::is_signed_v<T>,
+             "Cannot correctly store unsigned value due to sign bit flipping");
+  const auto base_size = key.size();
+  basics::StrAppend(key, sizeof(T));
+  absl::big_endian::Store(key.data() + base_size, value);
+  key[base_size] = static_cast<uint8_t>(key[base_size]) ^ 0x80;
+}
+
+template<typename T>
+T ReadSigned(const char* buf) {
```
MBkkt: string_view + assert size is enough?

server/connector/serenedb_connector.hpp
```diff
@@ -186,7 +207,8 @@ class RocksDBTable final : public axiom::connector::Table {
   std::vector<velox::connector::ColumnHandlePtr> rowIdHandles(
     axiom::connector::WriteKind kind) const final {
     SDB_ASSERT(_pk_type);
-    SDB_ASSERT(!_pk_type->children().empty());
+    // SDB_ASSERT(!_pk_type->children().empty());
```
MBkkt: remove?

server/connector/serenedb_connector.hpp
```diff
@@ -460,6 +482,7 @@ class SereneDBConnector final : public velox::connector::Connector {
         [&]<bool IsUpdate>() {
           std::vector<velox::column_index_t> pk_indices;
           if constexpr (IsUpdate) {
+            // here?
```
MBkkt: remove?

server/pg/sql_collector.h
```diff
@@ -73,7 +73,7 @@ class Objects : public irs::memory::Managed {
     // TODO(mbkkt) Maybe remove this and instead make catalog::Table be able
     // to implement connector::Table without allocation.
     // This probably requires changing axiom::connector::Table.
-    void EnsureTable() const;
+    void EnsureTable(bool is_insert = false) const;
```
MBkkt: bool named differently

server/connector/data_source.cpp
mkornaukhov: ?

server/pg/sql_analyzer_velox.cpp
```diff
@@ -2658,19 +2688,37 @@ SqlAnalyzer::TableAliasAndColumnNames SqlAnalyzer::ProcessTableColumns(
...
-    ProcessTableColumns(parent, node, table.RowType());
+  auto type = table.RowType();
+
+  auto [table_alias, column_names] = ProcessTableColumns(parent, node, type);
 
   object.EnsureTable();
 
+  if (load_implicit_pk) {
+    SDB_ASSERT(object.object);
+    if (table.PKColumns().empty()) {
```
MBkkt, line 2696:
```suggestion
  if (load_implicit_pk && table.PKColumns().empty()) {
```

server/connector/serenedb_connector.hpp
```diff
@@ -149,20 +152,35 @@ class RocksDBTable final : public axiom::connector::Table {
...
+      const auto generated_pk_name =
+        catalog::Column::GeneratePKName(collection.RowType()->names());
+
+      auto serenedb_column = std::make_unique<SereneDBColumn>(
+        generated_pk_name, velox::BIGINT(), catalog::Column::kGeneratedPKId);
+      columns.push_back(serenedb_column.get());
+
+      _column_map.emplace(generated_pk_name, serenedb_column.get());
+      _column_handles.push_back(std::move(serenedb_column));
+      order_columns.push_back(findColumn(generated_pk_name));
```
Dronplane: we could avoid calling `findColumn`.
We already have desired pointer in `serenedb_column`. If we move this push_back one line up (before we move from `serenedb_column`) we can just take a pointer.

server/pg/sql_analyzer_velox.cpp
```diff
@@ -1312,10 +1316,16 @@ void SqlAnalyzer::ProcessInsertStmt(State& state, const InsertStmt& stmt) {
 
   const auto& table = basics::downCast<catalog::Table>(logical_object);
   const auto& table_type = *table.RowType();
+  size_t table_columns = table_type.size();
+  if (!table.PKColumns().empty() &&
+      table.PKColumns()[0] == catalog::Column::kGeneratedPKId) {
+    table_columns--;
```
Dronplane: Is it still possible? I guess only on connector level generated pk would be added in this case.

server/connector/serenedb_connector.hpp
mkornaukhov: fixed

server/pg/sql_analyzer_velox.cpp
mkornaukhov: fixed

server/catalog/table.cpp
```diff
@@ -90,9 +90,8 @@ velox::RowTypePtr BuildPkType(const std::vector<Column>& columns,
   types.reserve(pk_columns.size());
 
   for (auto pk_col_id : pk_columns) {
-    auto it = std::find_if(
-      columns.begin(), columns.end(),
-      [pk_col_id](const Column& col) { return col.id == pk_col_id; });
+    auto it = absl::c_find_if(
+      columns, [pk_col_id](const Column& col) { return col.id == pk_col_id; });
```
MBkkt, line 94: NIT:
```suggestion
      columns, [&](const Column& col) { return col.id == pk_col_id; });
```
it's more logical, because you want always inline

server/catalog/table_options.cpp
```diff
@@ -92,6 +92,28 @@ Result ResolveId(ObjectId database, const auto& name, ObjectId& id) {
...
+  std::string candidate{kGeneratedPKPrefix};
+  size_t suffix = 0;
+  bool found_unique = false;
+
+  while (!found_unique) {
+    found_unique = true;
+    for (const auto& str : column_names) {
+      if (str == candidate) [[unlikely]] {
+        found_unique = false;
+        candidate.resize(kGeneratedPKPrefix.size());
```
MBkkt, line 107:
```suggestion
        candidate.erase(kGeneratedPKPrefix.size());
```

server/connector/data_source.cpp
```diff
@@ -101,11 +105,20 @@ std::optional<velox::RowVectorPtr> RocksDBDataSource::next(
...
     for (velox::column_index_t col_idx = 0; col_idx < num_columns; ++col_idx) {
-      basics::StrResize(key, key_old_size);
-      key_utils::AppendColumnKey(key, _column_ids[col_idx]);
+      basics::StrResize(key, table_prefix_size);
+      const auto column_id = _column_ids[col_idx];
+
+      size_t read_col_idx = col_idx;
+      if (column_id == catalog::Column::kGeneratedPKId) {
+        SDB_ASSERT(col_idx + 1 < num_columns,
+                   "Tables without columns are not supported");
```
MBkkt, line 117: do we have test for this? can you please add it if not

server/connector/data_source.cpp
```diff
@@ -231,11 +250,41 @@ velox::VectorPtr RocksDBDataSource::ReadUnknownColumn(
...
+    it, max_size, column_key,
+    [](uint64_t, std::string_view, std::string_view) {}, last_key);
   return velox::BaseVector::createNullConstant(velox::UNKNOWN(), vector_size,
                                                &_memory_pool);
 }
 
+velox::VectorPtr RocksDBDataSource::ReadColumnFromKey(
+  rocksdb::Iterator& it, uint64_t max_size, std::string_view column_key,
+  size_t table_prefix_size, std::string* last_key) {
+  static constexpr uint64_t kInitialVectorSize = 1;  // arbitrary value
```
MBkkt, line 262: share this constant?

server/connector/key_utils.hpp
```diff
@@ -20,6 +20,7 @@
 
 #pragma once
 
+#include <absl/base/internal/endian.h>
```
Dronplane, line 23: is this used? I guess we moved endian related logic into primary_key::AppendSigned

server/connector/data_source.cpp
```diff
@@ -249,7 +298,7 @@ uint64_t RocksDBDataSource::IterateColumn(rocksdb::Iterator& it,
     if (last_key) {
       *last_key = it.key().ToStringView().substr(column_key.size());
     }
-    func(vector_size, it.value().ToStringView());
+    func(vector_size, it.key().ToStringView(), it.value().ToStringView());
```
MBkkt, line 301: is it  expensive or not call `.key()` multiple times?

server/connector/key_utils.hpp
MBkkt, line 23:
```suggestion
```

Dronplane (approved):
LGTM

server/pg/sql_analyzer_velox.cpp
```diff
@@ -2658,19 +2683,37 @@ SqlAnalyzer::TableAliasAndColumnNames SqlAnalyzer::ProcessTableColumns(
...
-    ProcessTableColumns(parent, node, table.RowType());
+  auto type = table.RowType();
+
+  auto [table_alias, column_names] = ProcessTableColumns(parent, node, type);
 
   object.EnsureTable();
 
+  if (load_implicit_pk) {
+    SDB_ASSERT(object.object);
+    if (table.PKColumns().empty()) {
```
MBkkt, line 2696:
```suggestion
  if (load_implicit_pk && table.PKColumns().empty()) {
```
this assert is always true because code above, you can move it above if you want

server/connector/key_utils.hpp
mkornaukhov: IDE have added it. It's actually used, see `SetupColumnForKey()`

server/pg/sql_analyzer_velox.cpp
```diff
@@ -1455,6 +1459,16 @@ void SqlAnalyzer::ProcessUpdateStmt(State& state, const UpdateStmt& stmt) {
...
+  if (pk_type.size() == 0) {
+    auto generated_pk_name =
+      catalog::Column::GeneratePKName(table.RowType()->names());
+    auto column = state.resolver.Resolve(output_type, generated_pk_name);
+    std::string resolved{column.GetColumnName()};
+    auto expr = std::make_shared<lp::InputReferenceExpr>(velox::BIGINT(),
+                                                         std::move(resolved));
+    column_exprs.emplace_back(std::move(expr));
+    column_names.emplace_back(std::move(generated_pk_name));
+  }
```
MBkkt, line 1471: helper function/method for this code? it completely duplicates

maybe code above or below too.

deduplicate this plz

server/connector/serenedb_connector.hpp
```diff
@@ -186,7 +204,11 @@ class RocksDBTable final : public axiom::connector::Table {
   std::vector<velox::connector::ColumnHandlePtr> rowIdHandles(
     axiom::connector::WriteKind kind) const final {
     SDB_ASSERT(_pk_type);
```
MBkkt, line 206:
```suggestion
    SDB_ASSERT(_pk_type);
    SDB_ASSERT(!_column_handles.empty());
```

server/connector/serenedb_connector.hpp
```diff
@@ -149,20 +152,35 @@ class RocksDBTable final : public axiom::connector::Table {
...
+      auto serenedb_column = std::make_unique<SereneDBColumn>(
+        generated_pk_name, velox::BIGINT(), catalog::Column::kGeneratedPKId);
+      columns.push_back(serenedb_column.get());
+
+      _column_map.emplace(generated_pk_name, serenedb_column.get());
+      order_columns.push_back(serenedb_column.get());
+      _column_handles.push_back(std::move(serenedb_column));
+      _pk_type = velox::ROW({std::move(generated_pk_name)}, {velox::BIGINT()});
+      sort_order.resize(1, axiom::connector::SortOrder{.isAscending = true,
+                                                       .isNullsFirst = false});
```
MBkkt, line 181: lets remove this and do resize(max(1, size) above

server/connector/primary_key.hpp
```diff
@@ -45,4 +47,28 @@ void Create(const velox::RowVector& data,
...
+template<typename T>
+T ReadSigned(std::string_view buf) {
+  SDB_ASSERT(std::is_signed_v<T>,
+             "Cannot correctly read unsigned value due to sign bit flipping");
+  SDB_ASSERT(buf.size() >= sizeof(T));
+  using Unsigned = std::make_unsigned_t<T>;
+  static constexpr int kBits = sizeof(T) * CHAR_BIT;
+
+  auto val = absl::big_endian::Load<Unsigned>(buf.data());
+  val ^= static_cast<Unsigned>(1) << (kBits - 1);
```
MBkkt, line 69: https://godbolt.org/z/Wr4314Ye9

I made small check, and there's no diff how to read this

PR #75, opened 2025-12-18
Optimize data sink

mkornaukhov:
Get rid of extra primary key in-memory copies when constructing rocksdb key for cells

### DIFF
--- a/server/connector/key_utils.cpp
+++ b/server/connector/key_utils.cpp
@@ -46,11 +46,6 @@ void AppendColumnKey(std::string& key, catalog::Column::Id column_oid) {
   rocksutils::Append(key, column_oid);
 }
 
-void SetupColumnForKey(std::string& buf, catalog::Column::Id column_id) {
-  SDB_ASSERT(buf.size() >= sizeof(ObjectId) + sizeof(catalog::Column::Id));
-  absl::big_endian::Store(buf.data() + sizeof(ObjectId), column_id);
-}
-
 std::pair<std::string, std::string> CreateTableRange(ObjectId id) {
   SDB_ASSERT(id.isSet());
   if (id.id() != std::numeric_limits<decltype(id.id())>::max()) {
--- a/server/connector/key_utils.hpp
+++ b/server/connector/key_utils.hpp
@@ -58,7 +58,10 @@ void MakeColumnKey(const velox::RowVectorPtr& input,
 // Takes buffer in format
 // 'object_id | reserved for column_id | pk'
 // and fills column_id
-void SetupColumnForKey(std::string& buf, catalog::Column::Id column_id);
+inline void SetupColumnForKey(std::string& buf, catalog::Column::Id column_id) {
+  SDB_ASSERT(buf.size() >= sizeof(ObjectId) + sizeof(catalog::Column::Id));
+  absl::big_endian::Store(buf.data() + sizeof(ObjectId), column_id);
+}
 
 // creates range covering all rows of all columns of the table
 std::pair<std::string, std::string> CreateTableRange(ObjectId id);


server/connector/data_sink.cpp
```diff
@@ -92,25 +96,43 @@ void RocksDBDataSink::appendData(velox::RowVectorPtr input) {
...
                     false, true)
         .ok(),
       "Failed to acquire row lock for table {}", _object_key.id());
+
+    auto& cell_key_buffer = _cell_keys_buffers.emplace_back();
+    // [object_id] [column_id] [pk]
+    basics::StrResize(cell_key_buffer,
+                      combined_key.size() + sizeof(catalog::Column::Id));
+    std::memcpy(cell_key_buffer.data(), combined_key.data(), kObjectKeySize);
+    std::memcpy(
```
Dronplane: This logic is duplicated here and in the DeleteDataSink. Could it be moved to helper function in key_utils?

server/connector/data_sink.cpp
```diff
@@ -92,25 +96,43 @@ void RocksDBDataSink::appendData(velox::RowVectorPtr input) {
...
-  velox::IndexRange all_rows(0, input->size());
-  [[maybe_unused]] const auto& input_type = input->type()->asRow();
-  for (const auto& key : _row_keys) {
-    basics::StrResize(table_key, parent_size);
-    key_utils::AppendPrimaryKey(table_key, key);
+
+  // TODO(Dronplane) implement updating PK fields
+  std::string combined_key = key_utils::PrepareTableKey(_object_key);
+  SDB_ASSERT(sizeof(ObjectId) == combined_key.size());
+  static constexpr auto kObjectKeySize = sizeof(ObjectId);
```
Dronplane: Could it be just
```suggestion
  const auto kObjectKeySize = combined_key.size();
```
So we would remove here assumption about building Object Key.
constexpr would possibly have some benefit for memcpy below, but anyway I suggest to move that logic into key_utils where it could be constexpr.

server/connector/data_sink.cpp
```diff
@@ -1857,14 +1865,18 @@ void RocksDBDataSink::WritePrimitive(const T& value) {
...
+  std::span<const velox::vector_size_t> original_idx) {
   SDB_ASSERT(original_idx.empty() ||
              original_idx.size() > static_cast<size_t>(idx));
   const auto row_id = original_idx.empty() ? idx : original_idx[idx];
-  SDB_ASSERT(static_cast<size_t>(row_id) < _row_keys.size());
-  return _row_keys[row_id];
+  SDB_ASSERT(static_cast<size_t>(row_id) < _cell_keys_buffers.size());
+
+  auto& row_key = _cell_keys_buffers[row_id];
+  absl::big_endian::Store32(row_key.data() + sizeof(ObjectId), _column_id);
```
Dronplane: "Inserting" column_id logic is duplicated here and in the DeleteDataSink. Could we move this logic as helper method to the key_utils?
Also that would be good for separating the responsibility - DataSink should not know that much about how key is built.

MBkkt (changes requested):
Should be done differently.

Now it's still contains extra copies

server/connector/data_sink.cpp

server/connector/data_sink.cpp
mkornaukhov: Did it in key_utils

server/connector/data_sink.cpp
mkornaukhov: Duplicated code is in key_utils now

server/connector/data_sink.cpp
```diff
@@ -92,25 +95,34 @@ void RocksDBDataSink::appendData(velox::RowVectorPtr input) {
...
+
+  // TODO(Dronplane) implement updating PK fields
+  const std::string table_key = key_utils::PrepareTableKey(_object_key);
+  const auto num_rows = input->size();
+  _cell_keys_buffers.clear();
+  _cell_keys_buffers.reserve(num_rows);
+
+  std::string key_buffer;
+  for (size_t row_idx = 0; row_idx < num_rows; ++row_idx) {
+    key_utils::ReserveBuffer(key_buffer);
```
Dronplane: ReserveBuffer is always followed by PrepareBufferForLockKey.
Do we really need it as a separate function?

mkornaukhov: I think there might be use cases where where we don't need take a lock for row (for some reason, for example it's already taken), but need to write in rocksdb. So it's better to have separate functions

server/connector/key_utils.cpp
```diff
@@ -41,17 +45,27 @@ void AppendColumnKey(std::string& key, catalog::Column::Id column_oid) {
...
+  return {buf.begin() + sizeof(catalog::Column::Id), buf.end()};
+}
+
+void PrepareBufferForCellKey(std::string& buf, std::string_view table_key) {
+  SDB_ASSERT(table_key.size() == sizeof(ObjectId));
+  std::memcpy(buf.data(), table_key.data(), sizeof(ObjectId));
+}
+
+void SetupColumnForCellKey(std::string& buf, catalog::Column::Id column_id) {
+  absl::big_endian::Store32(buf.data() + sizeof(ObjectId), column_id);
```
MBkkt:
```suggestion
  absl::big_endian::Store(buf.data() + sizeof(ObjectId), column_id);
```
without this it's impossible to change column id type

server/connector/data_sink.hpp
```diff
@@ -188,8 +188,9 @@ class RocksDBDataSink : public velox::connector::DataSink {
   std::vector<catalog::Column::Id> _column_ids;
   velox::memory::MemoryPool& _memory_pool;
   SliceVector _row_slices;
-  primary_key::Keys _row_keys;
+  primary_key::Keys _cell_keys_buffers;  // buffer per row
```
MBkkt: NIT: I don't like name. it's meaningless, maybe just `_keys`? It's lock (row) keys and rocksdb (column) keys

server/connector/key_utils.cpp
```diff
@@ -41,17 +45,27 @@ void AppendColumnKey(std::string& key, catalog::Column::Id column_oid) {
...
-  rocksutils::Append(key, primary_key);
+void PrepareBufferForLockKey(std::string& buf, std::string_view table_key) {
+  SDB_ASSERT(table_key.size() == sizeof(ObjectId));
+  std::memcpy(buf.data() + sizeof(catalog::Column::Id), table_key.data(),
+              sizeof(ObjectId));
+}
+
+std::string_view GetLockKey(std::string& buf) {
+  return {buf.begin() + sizeof(catalog::Column::Id), buf.end()};
+}
```
MBkkt:
```suggestion
std::string_view GetLockKey(std::string_view buf) {
  return buf.substr(sizeof(catalog::Column::Id));
}
```

server/connector/key_utils.cpp
Dronplane: Assert buf size?

server/connector/key_utils.hpp
```diff
@@ -36,14 +39,26 @@ std::string PrepareColumnKey(ObjectId id, catalog::Column::Id column_oid);
...
-// Gets full row key by appending column OID and primary key to the Table key
-// created with PrepareTableKey.
-void AppendCellKey(std::string& key, catalog::Column::Id column_oid,
-                   std::string_view primary_key);
+// Prepare buffer for column key and call 'row_key_handle' on row_key
+template<typename Func>
+void MakeColumnKey(const velox::RowVectorPtr& input,
+                   const std::vector<velox::column_index_t>& pk_columns,
+                   velox::vector_size_t row_idx, std::string_view object_id,
+                   Func row_key_handle, std::string& key_buffer) {
```
Dronplane:
```suggestion
                   Func&& row_key_handle, std::string& key_buffer) {
```

Dronplane (approved):
LGTM

PR #88, opened 2025-12-22
perf: Fix serenedb tables column family options

MBkkt:
table scan perf was improved in few times in some cases

### DIFF
--- a/server/rocksdb_engine_catalog/rocksdb_option_feature.cpp
+++ b/server/rocksdb_engine_catalog/rocksdb_option_feature.cpp
@@ -48,6 +48,7 @@
 #include "basics/physical_memory.h"
 #include "basics/process-utils.h"
 #include "basics/system-functions.h"
+#include "catalog/table_options.h"
 #include "rocksdb_engine_catalog/rocksdb_column_family_manager.h"
 
 #ifdef SDB_CLUSTER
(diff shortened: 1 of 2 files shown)


server/rocksdb_engine_catalog/rocksdb_option_feature.cpp
```diff
@@ -1745,8 +1747,14 @@ rocksdb::ColumnFamilyOptions RocksDBOptionFeature::getColumnFamilyOptions(
     }
     if (_partition_files_for_documents_cf) {
       // partition .sst files by object id prefix
-      result.sst_partitioner_factory =
-        rocksdb::NewSstPartitionerFixedPrefixFactory(sizeof(uint64_t));
+      if (family == RocksDBColumnFamilyManager::Family::Data) {
+        result.sst_partitioner_factory =
+          rocksdb::NewSstPartitionerFixedPrefixFactory(
+            RocksDBKey::objectIdSize() + sizeof(catalog::Column::Id));
```
Dronplane, line 1753: NIT: Maybe better to use ObjectId size from catalog?

server/rocksdb_engine_catalog/rocksdb_options_provider.cpp
```diff
@@ -76,7 +77,12 @@ rocksdb::ColumnFamilyOptions RocksDBOptionsProvider::getColumnFamilyOptions(
   // EdgeIndex                   => Lookup
   switch (family) {
     case sdb::RocksDBColumnFamilyManager::Family::Data: {
-      // TODO(Dronplane): prefix extractor for better reads etc.
+      result.prefix_extractor.reset(rocksdb::NewFixedPrefixTransform(
+        RocksDBKey::objectIdSize() + sizeof(catalog::Column::Id)));
```
Dronplane, line 81: same here - ObjectId size from catalog

MBkkt: I will fix it later

PR #89, opened 2025-12-22
Add postgres transaction

codeworse:
- Decouple config and transaction state(Config, TxnState)
- Store transaction rocksdb pointer into TxnState, in order to allow multiple statements inside single transaction(BEGIN, COMMIT, ABORT etc.)

### DIFF
--- a/server/query/transaction.cpp
+++ b/server/query/transaction.cpp
@@ -51,10 +51,9 @@ TxnState::LazyTransaction::GetTransaction() const {
 }
 
 yaclib::Future<Result> TxnState::Begin() {
-  if (InsideTransaction()) {
-    return {};
+  if (!InsideTransaction()) {
+    _txn.SetTransaction();
   }
-  _txn.SetTransaction();
   return {};
 }
 
--- a/server/query/transaction.h
+++ b/server/query/transaction.h
@@ -39,6 +39,7 @@ class TxnState : public Config {
     void SetTransaction() { _initialized = true; }
 
     const std::shared_ptr<rocksdb::Transaction>& GetTransaction() const;
+
     bool IsInitialized() const noexcept { return _initialized; }
 
     void Reset() {
@@ -64,4 +65,5 @@ class TxnState : public Config {
  private:
   LazyTransaction _txn;
 };
+
 }  // namespace sdb


server/query/transaction.h
```diff
@@ -0,0 +1,90 @@
...
+
+  TxnState(Config& config) : _config(config) {}
+
+  yaclib::Future<Result> Begin();
+
+  yaclib::Future<Result> Commit();
+
+  yaclib::Future<Result> Abort();
+
+  bool InsideTransaction() const { return _txn.get() != nullptr; }
```
Dronplane: NIT it could be noexcept.

server/connector/serenedb_connector.hpp
```diff
@@ -64,11 +65,15 @@ class SereneDBConnectorTableHandle final
...
+  const std::shared_ptr<rocksdb::Transaction>& GetTransaction() const noexcept {
+    return _txn;
+  }
+
  private:
   std::string _name;
   ObjectId _table_id;
   catalog::Column::Id _table_count_field;
-  // TODO(Dronplane) transaction/snapshot management here
+  std::shared_ptr<rocksdb::Transaction> _txn = nullptr;
```
Dronplane: Can our handle outlive session? If not - we could store just raw pointer here

server/connector/serenedb_connector.hpp
```diff
@@ -251,7 +256,16 @@ class SereneDBConnectorInsertTableHandle final
   explicit SereneDBConnectorInsertTableHandle(
     const axiom::connector::ConnectorSessionPtr& session,
     const axiom::connector::TablePtr& table, axiom::connector::WriteKind kind)
-    : _session{session}, _table{table}, _kind{kind} {}
+    : _session{session}, _table{table}, _kind{kind} {
+    if (!session->config()) {
```
Dronplane: should this be an assert?

server/connector/serenedb_connector.hpp
```diff
@@ -373,8 +395,13 @@ class SereneDBConnectorMetadata final
...
                          "Wrong type of insert table handle");
-    if (serene_insert_handle->GetTransaction()) {
-      serene_insert_handle->GetTransaction()->Rollback();
+    SDB_ASSERT(session->config());
+    auto& config = basics::downCast<Config>(*session->config());
+    if (serene_insert_handle->GetTransaction() &&
+        !config.GetTxnState().InsideTransaction()) {
+      auto status = serene_insert_handle->GetTransaction()->Rollback();
+      VELOX_CHECK(status.ok(),
+                  "Failed to rollback transaction: ", status.ToString());
```
MBkkt: velox check has fmt::format message, so such thing doesn't return valid message.

please add test/replace with `SDB_THROW`?

server/query/transaction.cpp
```diff
@@ -0,0 +1,91 @@
...
+  Result r;
+  if (!InsideTransaction()) {
+    return yaclib::MakeFuture(std::move(r));
+  }
+  _variables.clear();
+  auto status = _txn->Rollback();
+  if (!status.ok()) {
+    r = {ERROR_INTERNAL,
+         absl::StrCat("Failed to rollback RocksDB transaction: ",
+                      status.ToString())};
```
MBkkt:
```suggestion
    r = {ERROR_INTERNAL, "Failed to rollback RocksDB transaction: ",
                      status.ToString()};
```

server/connector/serenedb_connector.hpp
```diff
@@ -67,11 +68,15 @@ class SereneDBConnectorTableHandle final
...
+  const std::shared_ptr<rocksdb::Transaction>& GetTransaction() const noexcept {
+    return _txn;
+  }
+
  private:
   std::string _name;
   ObjectId _table_id;
   catalog::Column::Id _table_count_field;
-  // TODO(Dronplane) transaction/snapshot management here
+  std::shared_ptr<rocksdb::Transaction> _txn = nullptr;
```
```suggestion
  std::shared_ptr<rocksdb::Transaction> _txn;
```

server/connector/serenedb_connector.hpp
```diff
@@ -67,11 +68,15 @@ class SereneDBConnectorTableHandle final
     return _table_count_field;
   }
 
+  const std::shared_ptr<rocksdb::Transaction>& GetTransaction() const noexcept {
+    return _txn;
+  }
```
```suggestion
  const auto& GetTransaction() const noexcept {
    return _txn;
  }
```

server/connector/serenedb_connector.hpp
```diff
@@ -273,7 +278,16 @@ class SereneDBConnectorInsertTableHandle final
...
-    : _session{session}, _table{table}, _kind{kind} {}
+    : _session{session}, _table{table}, _kind{kind} {
+    if (!session->config()) {
+      return;
+    }
+    const auto& config = basics::downCast<Config>(*_session->config());
+    if (config.GetTxnState().InsideTransaction()) {
+      _local_transaction = config.GetTxnState().GetTransaction();
+      SDB_ASSERT(_local_transaction);
+    }
```
MBkkt: It looks like same code is written for `SereneDBConnectorTableHandle`.
It's strange. Are you sure it should be like this?
To me it looks like we need to do this only here

server/connector/serenedb_connector.hpp
```diff
@@ -283,14 +297,14 @@ class SereneDBConnectorInsertTableHandle final
 
   const axiom::connector::TablePtr& Table() const noexcept { return _table; }
 
-  rocksdb::Transaction* GetTransaction() const noexcept {
+  const std::shared_ptr<rocksdb::Transaction>& GetTransaction() const noexcept {
     // TODO(Dronplane) we will have here non-owned transaction from upper layer
     // so we will check it first when we have it.
-    return _local_transaction.get();
+    return _local_transaction;
   }
```
MBkkt: const auto&

server/connector/serenedb_connector.hpp
```diff
@@ -300,7 +314,7 @@ class SereneDBConnectorInsertTableHandle final
   axiom::connector::ConnectorSessionPtr _session;
   axiom::connector::TablePtr _table;
   axiom::connector::WriteKind _kind;
-  mutable std::unique_ptr<rocksdb::Transaction> _local_transaction;
+  mutable std::shared_ptr<rocksdb::Transaction> _local_transaction;
```
MBkkt: it's just transaction, not local anymore

server/connector/serenedb_connector.hpp
```diff
@@ -283,14 +297,14 @@ class SereneDBConnectorInsertTableHandle final
...
+  const std::shared_ptr<rocksdb::Transaction>& GetTransaction() const noexcept {
     // TODO(Dronplane) we will have here non-owned transaction from upper layer
     // so we will check it first when we have it.
-    return _local_transaction.get();
+    return _local_transaction;
   }
 
   void SetLocalTransaction(
-    std::unique_ptr<rocksdb::Transaction>&& transaction) const noexcept {
+    std::shared_ptr<rocksdb::Transaction>&& transaction) const noexcept {
```
MBkkt: `std::shared_ptr<rocksdb::Transaction>` smart pointers have trivial abi, don't pass them as reference/pointer to smart pointer if it's not necessary

server/connector/serenedb_connector.hpp
```diff
@@ -283,14 +297,14 @@ class SereneDBConnectorInsertTableHandle final
 
   const axiom::connector::TablePtr& Table() const noexcept { return _table; }
 
-  rocksdb::Transaction* GetTransaction() const noexcept {
+  const std::shared_ptr<rocksdb::Transaction>& GetTransaction() const noexcept {
     // TODO(Dronplane) we will have here non-owned transaction from upper layer
     // so we will check it first when we have it.
-    return _local_transaction.get();
+    return _local_transaction;
```
MBkkt: Do we except it's not null? In such case maybe we can return `rocksdb::Transaction&` or `const rocksdb::Transaction&`

server/connector/serenedb_connector.hpp
MBkkt: why do we need this?

server/connector/serenedb_connector.hpp
```diff
@@ -67,11 +68,15 @@ class SereneDBConnectorTableHandle final
     return _table_count_field;
   }
 
+  const std::shared_ptr<rocksdb::Transaction>& GetTransaction() const noexcept {
```
MBkkt: why do we need this?

server/query/config.cpp
```diff
@@ -91,9 +91,9 @@ std::optional<std::string> Config::Get(std::string_view key) const {
 
 std::string_view Config::GetNonDefault(std::string_view key) const {
   {  // get from txn variables
-    auto it = _transaction.find(key);
-    if (it != _transaction.end()) {
-      return it->second.value;
+    auto str = _txn.Get(key);
+    if (str.data()) {
+      return str;
```
MBkkt: looks strange

server/query/transaction.h
```diff
@@ -0,0 +1,90 @@
...
+      dest.emplace(key, var.value);
+    }
+  }
+
+  const std::shared_ptr<rocksdb::Transaction>& GetTransaction() const noexcept {
+    return _txn;
+  }
+
+ private:
+  containers::FlatHashMap<std::string_view, Variable> _variables;
```
MBkkt: I don't think transaction should contains something about this

server/query/transaction.h
```diff
@@ -0,0 +1,90 @@
...
+      std::move(value),
+    };
+  }
+
+  void CopyVariablesTo(
+    std::unordered_map<std::string, std::string>& dest) const {
+    for (const auto& [key, var] : _variables) {
+      dest.emplace(key, var.value);
+    }
+  }
```
MBkkt: We don't want such methods in transaction

server/connector/serenedb_connector.hpp
```diff
@@ -32,9 +32,23 @@
...
+  SDB_ASSERT(session->config());
+  auto& txn = basics::downCast<TxnState>(*session->config());
+
+  if (txn.InsideTransaction()) {
+    return txn.GetTransaction();
+  }
+  return nullptr;
+}
+
+}  // namespace
```
MBkkt: remove anon namespace from header + make function inline?

server/connector/serenedb_connector.hpp
codeworse: txn could be fetched from session or be created(single query without postgres transaction)

server/connector/serenedb_connector.hpp
```diff
@@ -32,10 +32,22 @@
 #include "catalog/table_options.h"
 #include "data_sink.hpp"
 #include "data_source.hpp"
+#include "query/transaction.h"
 #include "rocksdb/utilities/transaction_db.h"
 
 namespace sdb::connector {
 
+static inline std::shared_ptr<rocksdb::Transaction> ExtractTransaction(
```
```suggestion
inline std::shared_ptr<rocksdb::Transaction> ExtractTransaction(
```

PR #100, opened 2025-12-24
misc: Make some things better

MBkkt:
- Remove unused ifdef
- Put fuzzer utils under test ifdefs instead of remove them (we can use this now!)
- Fix presto_min/max was duplicate sensitive for single argument


Dronplane (approved):
LGTM

PR #107, opened 2025-12-26
fix: Fix Axiom downstream columns bug

MBkkt:
Actually bug quite simple, when we enumerate joins we forgot about reset state, so order by from different state affect our own downstream columns.

We strongly believe that downstream columns is a not suitable abstraction, because it's very expensive to compute during join enumeration and it's actually incorrect for cases when different join key columns possible (consider clique join).
Maybe other bugs also exists with these abstraction.
But for now we consider such simple fix, instead of full downstream column rewrite.

 
Also this PR do following things:
1. Add convenient flag for explain all plans (just `"all"` doesn't work because of Postgres grammar for some reason)
2. Update docs about current compiler version


mkornaukhov (approved):
LGTM

PR #108, opened 2025-12-26
Fix aggregate functions on tables without pk + tables without columns

mkornaukhov:
- Support tables without columns at all
- Fix queries like `select 1 from t` and `select count(*) from t` where `t` is without explicit primary key
- Fix some review notes from previous PRs

### DIFF
--- a/server/pg/sql_analyzer_velox.cpp
+++ b/server/pg/sql_analyzer_velox.cpp
@@ -262,9 +262,7 @@ std::shared_ptr<const T> MakePtrView(const T& val) {
 
 template<typename T>
 std::shared_ptr<const T> MakePtrView(const std::shared_ptr<const T>& ptr) {
-  if (!ptr) {
-    return {};
-  }
+  SDB_ASSERT(ptr);
   return MakePtrView(*ptr);
 }
 


server/connector/primary_key.hpp
```diff
@@ -62,13 +62,12 @@ T ReadSigned(std::string_view buf) {
   SDB_ASSERT(std::is_signed_v<T>,
              "Cannot correctly read unsigned value due to sign bit flipping");
   SDB_ASSERT(buf.size() >= sizeof(T));
-  using Unsigned = std::make_unsigned_t<T>;
   static constexpr int kBits = sizeof(T) * CHAR_BIT;
 
-  auto val = absl::big_endian::Load<Unsigned>(buf.data());
-  val ^= static_cast<Unsigned>(1) << (kBits - 1);
+  auto val = absl::big_endian::Load<T>(buf.data());
+  val ^= static_cast<T>(1) << (kBits - 1);
```
MBkkt: it doesn't look ok? bit operation with signed integer

server/connector/serenedb_connector.cpp
```diff
@@ -27,11 +27,23 @@ SereneDBConnectorTableHandle::SereneDBConnectorTableHandle(
...
+  _table_count_field =
+    basics::downCast<const SereneDBColumn>(column_map.begin()->second)->Id();
+  if (_table_count_field == catalog::Column::kGeneratedPKId) {
+    // Iterating over generated primary key gives 0 rows,
+    // use another one
+    SDB_ASSERT(column_map.size() >= 2);
+    _table_count_field = basics::downCast<const SereneDBColumn>(
+                           std::next(column_map.begin())->second)
+                           ->Id();
+  }
```
MBkkt, line 46: Why can't we use pk column instead?

server/pg/pg_comm_task.cpp
```diff
@@ -365,13 +365,19 @@ void PgSQLCommTaskBase::HandleClientPacket(std::string_view packet) {
...
   const uint16_t num_fields = output_type ? output_type->size() : 0;
+  const auto* pg_node = castNode(RawStmt, statement.tree.GetRoot());
+  SDB_ASSERT(pg_node);
+
   // If it's DML but name of column isn't "rows" it means it's explain for DML.
   // In such case we should send rows as usual.
-  if (num_fields == 0 || (query.IsDML() && output_type->nameOf(0) == "rows")) {
+  if ((pg_node->stmt->type == T_CallStmt) ||
+      (!query.IsDataQuery() && num_fields == 0) ||
+      (query.IsDML() && output_type->nameOf(0) == "rows")) {
```
MBkkt, line 385:
```
  if (!query.IsDataQuery() ||
      (pg_node->stmt->type == T_CallStmt) ||
      (query.IsDML() && output_type->nameOf(0) == "rows")) {
```

server/pg/sql_analyzer_velox.cpp
```diff
@@ -2881,21 +2865,28 @@ State SqlAnalyzer::ProcessTable(State* parent, std::string_view schema_name,
...
-    if (table.PKColumns().empty()) {
-      auto generated_pk_name = catalog::Column::GeneratePKName(type->names());
+  if (table.Columns().empty()) {
+    auto state = parent->MakeChild();
+    auto dummy_row = std::make_shared<velox::RowVector>(
+      &_memory_pool, velox::ROW({}), velox::BufferPtr{}, 0,
+      std::vector<velox::VectorPtr>{});
+    state.root = std::make_shared<lp::ValuesNode>(
+      _id_generator.NextPlanId(), std::vector{std::move(dummy_row)});
+    state.resolver.CreateTable(table_alias, velox::ROW({}));
```
MBkkt: `velox::ROW({})` should be reused here

server/pg/pg_comm_task.cpp
```diff
@@ -365,13 +365,19 @@ void PgSQLCommTaskBase::HandleClientPacket(std::string_view packet) {
...
   const auto& output_type = query.GetOutputType();
   const uint16_t num_fields = output_type ? output_type->size() : 0;
+  const auto* pg_node = castNode(RawStmt, statement.tree.GetRoot());
+  SDB_ASSERT(pg_node);
+
   // If it's DML but name of column isn't "rows" it means it's explain for DML.
   // In such case we should send rows as usual.
-  if (num_fields == 0 || (query.IsDML() && output_type->nameOf(0) == "rows")) {
+  if ((pg_node->stmt->type == T_CallStmt) ||
+      (!query.IsDataQuery() && num_fields == 0) ||
```
Dronplane, line 384: Could you please add a clarification comment about this case.
I guess this for case when we query table without columns?

server/query/query.h
```diff
@@ -71,6 +71,8 @@ class Query {
            _logical_plan->is(axiom::logical_plan::NodeKind::kTableWrite);
   }
 
+  bool IsDataQuery() const { return static_cast<bool>(_logical_plan); }
```
MBkkt: NIT: I know codestyle isn't clear about this, but it's ok to write `!=/== nullptr` or `has_value()` if alternative is `static_cast<bool>` or `!!expr`

I think main point of current codestyle that we want to use explicit operator bool when it's possible to do this implicit (inside if condition for an example).
`shared_ptr` or `optional`, both has `explicit operator bool`, the only exception is raw pointer, and we want to interpret this raw pointer as smart pointer, we don't want difference in how code written for raw pointers compared to smart pointers (except smart pointers have default ctor)

```suggestion
  bool IsDataQuery() const { return _logical_plan != nullptr; }
```

tests/sqllogic/any/pg/simple/table_without_pk.test
```diff
@@ -114,3 +133,27 @@ query
...
+
+# Table without columns
+statement ok
+CREATE TABLE no_columns ();
+
+
+# TODO(mkornaukhov) fix sqllogictest-rs to process this case properly
+#query
+#SELECT * FROM no_columns;
+#----
```
MBkkt, line 146: What is the issue?

server/connector/serenedb_connector.cpp
mkornaukhov: We have to iterate existing in rocksdb column to generate from it's keys the primary key.
There no keys in rocksdb that starts with `object id | kGeneratedPKId `

tests/sqllogic/any/pg/simple/table_without_pk.test
mkornaukhov: Some oddities with empty line, I gonna fix it soon, not in this PR

server/connector/primary_key.hpp
mkornaukhov: It's ok since C++20: https://en.cppreference.com/w/cpp/language/operator_arithmetic.html
https://godbolt.org/z/73xoTfao5

server/pg/pg_comm_task.cpp
mkornaukhov: We want to **not** describe columns in the cases, where query:
1) is not data query (does not have logical plan)
2) is `CALL some_procedure()` (previously it was checked with with `num_fiels == 0`)
3) is `INSERT` or `DELETE`

server/pg/pg_comm_task.cpp
mkornaukhov: that is not true, we have to check `num_fields == 0` in case of non data queries.
For example, `SHOW`

server/pg/pg_comm_task.cpp
MBkkt: I think main Andrei point, is that we want to write this in this file, not in the PR

MBkkt: Please also mention about num_fields == 0 needed for show

Dronplane: Yes, condition is not very straightforward so better to leave some remarks for future readers

server/connector/primary_key.hpp
mkornaukhov: 6.8.1, C++20 standard
> For each value x of a signed integer type, the value of the
corresponding unsigned integer type congruent to x modulo 2
N has the same value of corresponding bits in
its value representation

MBkkt (approved):

PR #113, opened 2025-12-27
fix: Fix different bugs

MBkkt:
- Enable assert and velox/etc dcheck
- Fix rocksdb prefix extractor
- Fix races in axiom runner
- Fix pg_comm_task deadlock possible
- Fix general server deadlock possible
- Fix crash in velox abort
- Fix and speedup axiom common conjunction extraction
- Fix data race in libpg_query
- Enable select5.test, but there's two tricks: replace varchar(40) with varchar and set join_order_algorithm to greedy
- Add join_order_algorithm variable and implement greedy algorithm in Axiom
- Add execution_threads parameter also allow to have serial execution (a lot of more convenient for debug and also faster for simple queries)

### DIFF
--- a/tests/sqllogic/any/pg/sqlite/select5.test
+++ b/tests/sqllogic/any/pg/sqlite/select5.test
@@ -3,6 +3,7 @@ flat-values true
 hash-threshold 8
 
 # TODO(mbkkt) Improve join order algorithm
+onlyif serenedb
 statement ok
 SET join_order_algorithm TO greedy;
 


server/connector/data_sink.cpp
```diff
@@ -120,16 +120,6 @@ void RocksDBDataSink::appendData(velox::RowVectorPtr input) {
...
   for (velox::column_index_t i = 0; i < num_columns; ++i) {
-    // Exclude only UNKNOWN from equivalency check as UNKNOWN is just a NULL and
-    // compatible with everything.
-    // Intentionally check UNKNOWN after equivalency to make compiler execute
-    // child lookup by name as it is also a necessary check.
-    // Even keys during UPDATE should be validated as it is important for
-    // RocksDB key generation
-    VELOX_DCHECK(_row_type->findChild(input_type.nameOf(i))
-                   ->equivalent(*input_type.childAt(i)) ||
-                 input_type.childAt(i)->kind() == velox::TypeKind::UNKNOWN);
```
MBkkt, line 131: this doesn't work, and quite fragile, so I removed it. should be checked in planning, not on execution

PR #116, opened 2025-12-29
Add numRows into rocksdb table

codeworse:
- Enable `numRows` to plan optimizer
- Add `pg_class.reltuples` for tables
axiom: https://github.com/serenedb/axiom/pull/21


server/connector/serenedb_connector.cpp
```diff
@@ -47,4 +49,19 @@ SereneDBConnectorTableHandle::SereneDBConnectorTableHandle(
   _txn = ExtractTransaction(session);
 }
 
+uint64_t RocksDBTable::numRows() const {
+  uint64_t count = 0;
```
gnusi: Let's keep rocksdb related stuff in a single place

codeworse: Then there would be multiple changes in the header file, since RocksDBTable uses SereneDBConnector and is defined before connector

server/connector/serenedb_connector.cpp
```diff
@@ -47,4 +49,25 @@ SereneDBConnectorTableHandle::SereneDBConnectorTableHandle(
...
+    column_handle = _column_handles[1].get();
+  }
+  auto [start, end] =
+    key_utils::CreateTableColumnRange(TableId(), column_handle->Id());
+  auto* db = GetServerEngine().db();
+  auto connector = std::dynamic_pointer_cast<SereneDBConnector>(
+    velox::connector::getConnector(StaticStrings::kSereneDBConnector));
+  SDB_ASSERT(connector);
+  auto* cf = &connector->GetColumnFamily();
+  db->GetApproximateMemTableStats(cf, rocksdb::Range{start, end}, &count,
```
Dronplane: Would that include  rows that are in sst files?

server/connector/serenedb_connector.cpp
codeworse: done

server/connector/serenedb_connector.cpp
codeworse: I guess no: https://github.com/facebook/rocksdb/wiki/Approximate-Size/64238d689ac8f95d6c7ec6f5c192fd55e8c6ada8
> The `GetApproximateMemTableStats` is only supported for memtables created by SkipListFactory.

Dronplane: If I got it correctly we should use  GetApproximateSizes() instead and gather sst and memtables  statistics

codeworse: But `GetApproximateSizes()` returns the information about memory usage, not number of rows(entities)

server/rocksdb_engine_catalog/rocksdb_sync_thread.cpp
```diff
@@ -49,6 +51,8 @@ Result RocksDBSyncThread::syncWal() {
   // > Currently only works if allow_mmap_writes = false in Options.
   SDB_ASSERT(!_engine.rocksDBOptions().allow_mmap_writes);
 
+  SyncStats();
```
gnusi: Move to background thread

server/connector/data_sink.cpp
```diff
@@ -128,6 +132,9 @@ void RocksDBDataSink::appendData(velox::RowVectorPtr input) {
       WriteColumn(input->childAt(i), folly::Range{&all_rows, 1}, {});
     }
   }
+  auto table_shard = catalog::GetTableShard(_object_key);
```
Dronplane: maybe better to do this once in finish? not for every batch.

server/connector/data_sink.cpp
```diff
@@ -128,6 +132,9 @@ void RocksDBDataSink::appendData(velox::RowVectorPtr input) {
...
+  SDB_ASSERT(table_shard);
+  table_shard->IncreaseRows(num_rows);
```
Dronplane: How this changes would be rollbacked in case of transaction abort?

server/connector/data_sink.cpp
```diff
@@ -2039,6 +2046,9 @@ void RocksDBDeleteDataSink::appendData(velox::RowVectorPtr input) {
       }
     }
   }
+  auto table_shard = catalog::GetTableShard(_object_key);
```
Dronplane: Move to finish for single call not for every batch?

server/connector/data_sink.cpp
codeworse: moved update table shard to `finishWrite`, that way after `abortWrite` the stats won't change

Dronplane: But what if we have a global transaction? That migh be rollbacked several queries after.
Maybe we can do "stats update" in the same transaction as insert? RocksDB supports multi columnf family transactions. So that should be ok.

codeworse: The table stats don't update in the same time as data, since it is updated by background thread. So, in case of global transaction after the commit stmt the stats will be firstly updated in `TableShard` and only after some time will be dropped to the disk(or in case of shutdown)

server/connector/serenedb_connector.hpp
```diff
@@ -398,6 +403,17 @@ class SereneDBConnectorMetadata final
         SDB_THROW(ERROR_INTERNAL,
                   "Failed to commit transaction: ", r.errorMessage());
       }
+      auto shard = catalog::GetTableShard(
```
Dronplane: And if we don't get there (because of `HasTransactionBegin` returned true) - looks like stats won't be updated. At least I can't find where it would be done then.

Dronplane: Ah I guess I've got it - then we just wait until background update happens?
But then what if our global transaction is rollbacked?

codeworse: The stats are updated only after `COMMIT` call. So, in case of in-progress transaction table stats won't change, since we have single `TableShard` instance.
```sql
CREATE TABLE t(...)
BEGIN;
INSERT INTO t VALUES (1);
SELECT * FROM t; -- num_rows = 0;
COMMIT;
SELECT * FROM t; -- num_rows = 1;
```

server/pg/serialize.cpp
```diff
@@ -411,6 +411,8 @@ void SerializePrimitiveType(SerializationContext context,
       const auto time_us = absl::ToInt64Microseconds(time - kGapUs);
       absl::big_endian::Store64(context.buffer->GetContiguousData(8), time_us);
     }
+  } else if constexpr (Kind == velox::TypeKind::REAL) {
+    SerializeFloat<float, Format>(context, decoded_vector, row);
```
MBkkt:
```suggestion
```

server/storage_engine/table_shard.cpp
```diff
@@ -68,8 +68,9 @@ catalog::TableMeta MakeTableMeta(const catalog::Table& c) {
   };
 }
 
-TableShard::TableShard(catalog::TableMeta collection_meta)
-  : _collection_meta{std::move(collection_meta)} {}
+TableShard::TableShard(catalog::TableMeta collection_meta,
+                       catalog::TableStats stats)
```
Dronplane:
```suggestion
                       const catalog::TableStats& stats)
```

server/connector/serenedb_connector.cpp
```diff
@@ -20,15 +20,19 @@
 
 #include "serenedb_connector.hpp"
 
+#include "basics/assert.h"
```
Dronplane: looks redundant - this include exists in serenedb_connector.hpp:33

server/connector/serenedb_connector.hpp
```diff
@@ -392,6 +397,19 @@ class SereneDBConnectorMetadata final
     }
 
     int64_t number_of_locked_primary_keys = rocksdb_transaction->GetNumKeys();
+    if (serene_insert_handle->Kind() == axiom::connector::WriteKind::kInsert ||
+        serene_insert_handle->Kind() == axiom::connector::WriteKind::kDelete) {
+      auto& rocksdb_table =
+        basics::downCast<const RocksDBTable>(*serene_insert_handle->Table());
+      // TODO(codeworse) : Not quite correct, should be updated
```
Dronplane: please add some more context - what is not correct?

codeworse: I was wrong, since `number_of_locked_primary_keys = 2 * affected rows` in case of `UPDATE` stmt. However, in `INSERT/DELETE` stmts should be okay. So, the comment is incorrect, I removed it

tests/sqllogic/sdb/pg/explain/row_stat.test
```diff
@@ -0,0 +1,31 @@
...
+
+statement ok
+CREATE TABLE f(id text primary key, value int);
+
+
+statement ok
+INSERT INTO f VALUES ('3', 1), ('7', 2), ('13', 3);
+
+
+query ?
```
MBkkt:
```suggestion
query
```

tests/sqllogic/sdb/pg/explain/row_stat.test
```diff
@@ -0,0 +1,31 @@
+statement ok
+CREATE TABLE t(id text);
+
+
+statement ok
```
MBkkt: statement count

tests/sqllogic/sdb/pg/system/pg_class.test
```diff
@@ -87,3 +87,40 @@ query
...
+
+
+query
+SELECT relname, reltuples FROM pg_class WHERE relname = 't';
+----
+relname	reltuples
+t	6
+
+
+statement ok
```
MBkkt: statement count

tests/sqllogic/sdb/pg/system/pg_class.test
```diff
@@ -87,3 +87,40 @@ query
...
```
MBkkt: statement count

tests/sqllogic/sdb/pg/system/pg_class.test
```diff
@@ -87,3 +87,40 @@ query
 SELECT relname, reloptions FROM pg_class WHERE reloptions <> ARRAY[]::varchar[];
 ----
 relname	reloptions
+
+
+statement ok
+INSERT INTO t VALUES ('a'), ('b'), ('c');
+
+
+statement ok
```
MBkkt: statement count

tests/sqllogic/sdb/pg/system/pg_class.test
```diff
@@ -87,3 +87,40 @@ query
 SELECT relname, reloptions FROM pg_class WHERE reloptions <> ARRAY[]::varchar[];
 ----
 relname	reloptions
+
+
+statement ok
```
MBkkt: statement count

tests/sqllogic/sdb/pg/explain/row_stat.test
```diff
@@ -0,0 +1,31 @@
...
+INSERT INTO t
+SELECT ((x * x + x + 1) % 1000)::text FROM generate_series(1, 1000) AS t(x);
+
+
+statement ok
+CREATE TABLE f(id text primary key, value int);
+
+
+statement ok
```
MBkkt: statement count

server/rocksdb_engine_catalog/rocksdb_engine_catalog.cpp
```diff
@@ -1582,7 +1582,27 @@ Result RocksDBEngineCatalog::MarkDeleted(const catalog::Table& c,
...
+  if (!is_new) {
+    // Load Table stats
+    RocksDBKeyWithBuffer key;
+    key.constructSchemaObject(RocksDBEntryType::CounterValue,
+                              collection.GetDatabaseId(),
+                              collection.GetSchemaId(), collection.GetId());
+    std::string value;
+    _db->Get(
+      rocksdb::ReadOptions{},
+      RocksDBColumnFamilyManager::get(RocksDBColumnFamilyManager::Family::Data),
```
MBkkt: we don't want to store stats in data column family

server/rocksdb_engine_catalog/rocksdb_engine_catalog.cpp
```diff
@@ -1582,7 +1582,27 @@ Result RocksDBEngineCatalog::MarkDeleted(const catalog::Table& c,
...
+      key.string(), &value);
+    if (auto r = vpack::ReadObjectNothrow<catalog::TableStats>(
```
MBkkt: you write this object differently

server/rocksdb_engine_catalog/rocksdb_value.cpp
```diff
@@ -162,6 +162,7 @@ RocksDBValue::RocksDBValue(RocksDBEntryType type, vpack::Slice data)
     case RocksDBEntryType::Index:
     case RocksDBEntryType::IndexTombstone:
     case RocksDBEntryType::Role:
+    case RocksDBEntryType::CounterValue:
```
MBkkt: we don't want to use existing EntryType for this

server/catalog/table_options.h
```diff
@@ -196,6 +196,18 @@ struct CreateTableRequest {
...
 };
 
+struct TableStats {
+  struct Diff {
+    int64_t delta_num_rows = 0;
+    Diff& Add(const Diff& other) {
+      delta_num_rows += other.delta_num_rows;
+      return *this;
+    }
+  };
```
MBkkt: can we remove this struct?

server/rocksdb_engine_catalog/rocksdb_value.cpp
codeworse: Create `RocksDBEntryType::Stats`

server/catalog/table_options.h
codeworse: removed

server/rocksdb_engine_catalog/rocksdb_engine_catalog.cpp
codeworse: change with `WriteTuple`/`ReadTupleNothrow`

server/rocksdb_engine_catalog/rocksdb_engine_catalog.cpp
codeworse: change to `Family::Definitions`

MBkkt (approved):

Issue #109, opened 2025-12-26
Single snapshot in simple query


codeworse:
For now in simple query(outside of transaction) snapshot is set to nullptr:
```cpp
  const rocksdb::Snapshot* snapshot = nullptr;
  if (auto txn = serene_table_handle.GetTransaction()) {
    // Inside transaction
    snapshot = txn->GetSnapshot();
  }
  return std::make_unique<RocksDBDataSource>(
    *connector_query_ctx->memoryPool(), snapshot, _db, _cf, output_type,
    column_oids, object_key);
```
That way, different table in the same query can fetch different snapshots(for instance, if there is concurrent transaction).

PR #125, opened 2025-12-30
Single snapshot in simple query

codeworse:
closing: https://github.com/serenedb/serenedb/issues/109
A little bug in `EXPLAIN (EXECUTION)`: Previously the finish_write was assigned even in explain query with execution option, however this option do not require execution itself(actually, only stats option requires). Therefore, in dtor of query the finish_write was called and run `abortWrite()`, which was incorrect.


server/connector/serenedb_connector.cpp
```diff
@@ -44,7 +47,16 @@ SereneDBConnectorTableHandle::SereneDBConnectorTableHandle(
...
-  _txn = ExtractTransaction(session);
+  const auto& txn_state = ExtractTransactionState(session);
+  const auto& txn = txn_state.GetTransaction();
+  if (txn_state.InsideTransaction()) {
+    SDB_ASSERT(txn);
+    const auto* snapshot = txn->GetSnapshot();
+    SDB_ASSERT(snapshot);
+    _snapshot = std::make_shared<RocksDBSnapshotView>(snapshot);
+  } else {
+    _snapshot = std::make_shared<RocksDBSnapshot>(*GetServerEngine().db());
```
Dronplane: At first glance this part of issue "different table in the same query can fetch different snapshots" is still there.
If we are not "InsideTransaction" each table handle will get here and acquire own snapshot. So several different snapshots are possible in one session.
I'd expect this snapshot is set somewhere inside the session and used in all consequent table handles. Like it is now in case of Transaction where snapshot from transaction will be used for every table handle in the session.

codeworse: I've changed the implementation. Shortly, the snapshot is stored inside `TxnState`, it could be acquired via `GetSnapshot` and the method would return:
- snapshot from transaction(if we inside transaction)
- snapshot from db(outside of transaction)

Snapshot from db is created _on demand_ and released with cursor dtor(when the simple query is finished).

Dronplane: The only case that bothers me now it this:
We have query like `INSERT INTO foo (blah, bar) SELECT a.foo, b.bar FROM a JOIN b ON a.id = b.id`
So there is two TableHandles (for a and for b) and one InsertTableHandle (for foo).

Could it be that the order of initialization would be:
1. TableHandle for a. Here GetSnapshot would be called and snapshot would be acquired from db by lazy snapshot.
2. Some parallel transaction is committed to the database.
3. InsertTableHandle for foo. Implicit transaction would be created in the transaction state.
4. TableHandle for b. Here as txn exists in transaction state - snapshot from transaction would be used.

So we end up reading from different snapshots while iterating a and b.
Is this a valid concern?

codeworse: Yes, it is a valid concern. I guess, there is an approach to handle it:
1) Add `SetExternalTransaction` method, so connectors will be able to set custom transaction in case of simple query
2) Store lazy snapshot inside TableHandler. That way it won't initialize snapshot into ctor

Therefore, in case we create TableHandler before InsertTableHandler we still will acquire the snapshot after InsertTableHandler is created(and external transaction is put inside TxnState), that way they both will have the same snapshot from external transaction(transaction only for simple query)

What do you think?

Dronplane: I guess it would be enough to store LazySnapshot inside TableHandle and provide it to the DataSource.  And DataSource only use that lazy snapshot  to get actual snapshot when first read is executed. Key point here would be that no implicit snapshots are acquired before execution phase starts. In that case we have a guarantee that if there is a write operation in the query - we have transaction set in the transaction state. If there is no writes - first DataSource to run will set a snapshot and others will use it. So code for storing implicit snapshot might need some thread safety.

codeworse: I agree, but in case of single query: `INSERT INTO foo (blah, bar) SELECT a.foo, b.bar FROM a JOIN b ON a.id = b.id` TxnState won't store any transaction. The transaction is stored only inside `InsertTableHandler`, so we need to manually set the transaction inside TxnState

Dronplane: Ah, I see. Missed that. Then indeed we will need something to put transaction into state.

server/connector/serenedb_connector.hpp
```diff
@@ -357,25 +359,21 @@ class SereneDBConnectorMetadata final
...
     const axiom::connector::ConnectorWriteHandlePtr& handle,
     const std::vector<velox::RowVectorPtr>& write_results) final {
-    // TODO(Dronplane) if we have here transaction from upstream, we should not
-    // commit here. But currently we don't have such case so transaction must be
-    // local and stored in the insert handle.
-    auto serene_insert_handle =
+    const auto serene_insert_handle =
       std::dynamic_pointer_cast<const SereneDBConnectorInsertTableHandle>(
         handle->veloxHandle());
+    auto& txn = serene_insert_handle->GetTxnState();
```
Dronplane: I guess this line should be moved after SDB_ENSURE.

Dronplane (approved):
LGTM

server/connector/serenedb_connector.hpp
codeworse: done

PR #130, opened 2026-01-01
perf: Speedup iresearch search and count

MBkkt:
- perf: Implement count for DocIterator
- perf: Implement and use advance instead of next for DocIterator
- perf: Make scorer specialization only for 1 scorer, 2+ scorers can be slow case => less template instantiations

### DIFF
--- a/libs/iresearch/include/iresearch/formats/columnstore2.cpp
+++ b/libs/iresearch/include/iresearch/formats/columnstore2.cpp
@@ -228,6 +228,15 @@ class RangeColumnIterator : public ResettableDocIterator,
     return std::get<DocAttr>(_attrs).value;
   }
 
+  doc_id_t advance() final {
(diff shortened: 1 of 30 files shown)


libs/iresearch/include/iresearch/formats/columnstore2.cpp
```diff
@@ -313,26 +311,20 @@ class BitmapColumnIterator : public ResettableDocIterator,
...
-  }
-
-  bool next() final {
-    if (_bitmap.next()) {
-      std::get<PayAttr>(_attrs).value = this->payload(_bitmap.index());
-      return true;
-    }
-    std::get<PayAttr>(_attrs).value = {};
-    return false;
+    std::get<PayAttr>(_attrs).value =
```
gnusi, line 325: eof unlike?

PR #134, opened 2026-01-02
feat: Make function and table resolved separately

pashandor789:


server/pg/sql_analyzer_velox.cpp
```diff
@@ -2778,6 +2778,7 @@ void SqlAnalyzer::ProcessDistinctAll(State& state, const List* sort_clause,
   }
 
   state.has_groupby = true;
+  ValidateAggrInputRefs(state, std::span<const lp::ExprPtr>{grouping_keys});
```
pashandor789, line 2781: that's about pretty bug select distinct coalesce blah-blah test

PR #144, opened 2026-01-09
misc: Some improvments for search

MBkkt:
note for @gnusi find_if + accumulate works slower

### DIFF
--- a/libs/iresearch/include/iresearch/formats/formats_impl.cpp
+++ b/libs/iresearch/include/iresearch/formats/formats_impl.cpp
@@ -2668,8 +2668,9 @@ doc_id_t Wanderator<IteratorTraits, FieldTraits, WandExtent, Root>::seek(
     [[maybe_unused]] uint32_t notify = 0;
     [[maybe_unused]] const auto threshold = _skip.Reader().threshold;
 
+    uint32_t doc{};
     while (this->_begin != std::end(this->_buf.docs)) {
-      const auto doc = *this->_begin++;
+      doc = *this->_begin++;
 
       if constexpr (IteratorTraits::Position()) {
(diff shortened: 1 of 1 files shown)


libs/iresearch/include/iresearch/formats/formats_impl.cpp
```diff
@@ -2663,60 +2660,54 @@ doc_id_t Wanderator<IteratorTraits, FieldTraits, WandExtent, Root>::seek(
...
+    if constexpr (Root) {
+      if constexpr (IteratorTraits::Position()) {
+        std::get<Position>(_attrs).Notify(notify);
+      }
+      target = doc_value + 1;
     }
-
-    target = doc_value + 1;
-  }
+  } while (Root);
```
MBkkt, line 2718: this (remove cycle for non-root) gives speedup for wand disj/conj btw (around 10%)

libs/iresearch/include/iresearch/formats/formats_impl.cpp
```diff
@@ -2204,78 +2206,70 @@ doc_id_t DocIteratorImpl<IteratorTraits, FieldTraits, WandExtent>::seek(
...
-  if (!_docs_count) [[likely]] {
+  if (_docs_count == 0) [[likely]] {
   seek_after_initialization:
     SDB_ASSERT(_skip.NumLevels());
 
     this->_left = _skip.Seek(target);
-    this->_doc_in->Seek(last.doc_ptr);
     std::get<DocAttr>(_attrs).value = last.doc;
+
+    this->_begin = std::end(this->_buf.docs);
```
MBkkt, line 2271: because of this line I'm able to remove advance from seek

PR #147, opened 2026-01-12
DataSink for search Part 1

Dronplane:
Implements Insert and Delete operation for IResearch index as a "writer" that could be integrated into connector DataSink.
This PR adds support for booleans, nulls, strings and numerics.
Complex types support would be added after we have proper metadata for index.
Same way for now all columns are indexed but later list of column ids would be set in the index meta together with the analyzers.

Next step would be implementing Update as Delete + Re-Insert operation in terms of index.

Related to https://github.com/serenedb/cluster/issues/6

### DIFF
--- a/server/connector/serenedb_connector.hpp
+++ b/server/connector/serenedb_connector.hpp
@@ -522,7 +522,8 @@ class SereneDBConnector final : public velox::connector::Connector {
           }
           return std::make_unique<RocksDBDataSink>(
             *(txn.GetTransaction()), _cf, *connector_query_ctx->memoryPool(),
-            object_key, pk_indices, column_oids, std::vector<std::unique_ptr<SinkInsertWriter>>{}, IsUpdate);
+            object_key, pk_indices, column_oids,
+            std::vector<std::unique_ptr<SinkInsertWriter>>{}, IsUpdate);
         });
     }
     if (serene_insert_handle.Kind() == axiom::connector::WriteKind::kDelete) {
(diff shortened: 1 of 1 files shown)


server/connector/search_sink_writer.cpp
```diff
@@ -0,0 +1,335 @@
...
+
+template<typename T>
+SearchSinkInsertWriter::Field& SearchSinkInsertWriter::WriteNumericValue(
+  std::string_view, std::span<const rocksdb::Slice> cell_slices,
+  SearchSinkInsertWriter::Field& field) {
+  SDB_ASSERT(cell_slices.size() == 1);
+  SDB_ASSERT(sizeof(T) == cell_slices[0].size());
+  // this is true as long as we match machine ending with storage ending
+  static_assert(basics::IsLittleEndian());
+  field.SetNumericValue(*reinterpret_cast<const T*>(cell_slices[0].data()));
```
gnusi: maybe?
`absl::little_endian::Load<T>(cell_slices[0].data()):`

server/connector/search_sink_writer.hpp
```diff
@@ -0,0 +1,154 @@
...
+  template<velox::TypeKind Kind>
+  void SetupColumnWriter(sdb::catalog::Column::Id column_id, bool have_nulls);
+
+  Field _field;
+  Field _pk_field;
+  Field _null_field;
+  std::string _name_buffer;
+  std::string _null_name_buffer;
+  irs::IndexWriter::Transaction& _trx;
+  std::unique_ptr<irs::IndexWriter::Document> _document;
```
gnusi: `std::optional` would be better or
Making `IndexWriter::Document` is the best option

server/connector/data_sink.cpp
```diff
@@ -1764,24 +1768,19 @@ void RocksDBDataSink::WriteArrayValue(const velox::BaseVector& input,
...
+    writer->Write(_row_slices, key);
   }
-  if (!status.ok()) {
-    SDB_THROW(rocksutils::ConvertStatus(status));
+}
+
+void RocksDBDataSink::WriteNull(std::string_view key) {
+  // empty slice denotes NULL value
+  static rocksdb::Slice gNullSlice;
+  static constexpr std::span<const rocksdb::Slice> kSlices{&gNullSlice, 1};
```
gnusi: do we really need static here, maybe just: 
```
rocksdb::Slice gNullSlice;
std::span kSlices{&NullSlice, 1};
```

Dronplane: No we don't. Removed.

PR #150, opened 2026-01-12
Fix `AsyncReadSome`

codeworse:


server/general_server/generic_comm_task.cpp
```diff
@@ -79,6 +79,10 @@ void GenericCommTask<SocketType, Base>::Close(asio_ns::error_code ec) {
 
 template<rest::SocketType SocketType, typename Base>
 void GenericCommTask<SocketType, Base>::AsyncReadSome() try {
+  if (!_protocol || Stopped()) {
+    return;
+  }
```
MBkkt, line 84: it doesn't look correct

PR #151, opened 2026-01-12
fix: Transaction management

MBkkt:
Some follow-ups

### DIFF
--- a/server/query/transaction.cpp
+++ b/server/query/transaction.cpp
@@ -115,9 +115,6 @@ void Transaction::CreateRocksDBTransaction() {
 }
 
 void Transaction::Destroy() noexcept {
-  if (_state == State::None) {
-    return;
-  }
   _state = State::None;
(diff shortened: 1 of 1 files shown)


server/connector/serenedb_connector.hpp
```diff
@@ -389,15 +382,16 @@ class SereneDBConnectorMetadata final
...
-        SDB_THROW(ERROR_INTERNAL,
-                  "Failed to rollback transaction: ", status.errorMessage());
-      }
+    auto& transaction = serene_insert_handle->GetTransaction();
+    auto* rocksdb_transaction = transaction.GetRocksDBTransaction();
+    if (!rocksdb_transaction) [[unlikely]] {
+      return velox::ContinueFuture::make();
+    }
+    // TODO: should be rollback to last save point
+    auto r = transaction.Rollback();
```
MBkkt, line 391: we should rollback even in case of "begin"

PR #152, opened 2026-01-13
Fix bug with deleting all rows from table without PK

### LINKED ISSUE #149: Delete from table without primary keys does not work
Looks like `delete from table_without_pk` does not work, find out details and make fix if needed.

mkornaukhov:
In some scans in `RocksDBDatasource` we need only columns that are not stored as effective columns (using RocksDB values), or even we need not any specific column at all as for count query. But we need some column id to construct prefix of RocksDB key to iterate over all existing keys and column specific values. I've named such a column as **effective** one, it means this column **effectively** stored as a RocksDB values. Connector always passes such a column id to `RocksDBDatasource` creation.

### DIFF
--- a/server/connector/data_source.cpp
+++ b/server/connector/data_source.cpp
@@ -41,7 +41,8 @@ constexpr uint64_t kInitialVectorSize = 1;  // arbitrary value
 RocksDBDataSource::RocksDBDataSource(
   velox::memory::MemoryPool& memory_pool, const rocksdb::Snapshot* snapshot,
   rocksdb::DB& db, rocksdb::ColumnFamilyHandle& cf, velox::RowTypePtr row_type,
-  std::vector<catalog::Column::Id> column_oids, ObjectId object_key)
+  std::vector<catalog::Column::Id> column_oids,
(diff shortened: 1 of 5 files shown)


server/connector/data_source.cpp
```diff
@@ -41,14 +41,16 @@ constexpr uint64_t kInitialVectorSize = 1;  // arbitrary value
...
+  std::vector<catalog::Column::Id> column_oids,
+  catalog::Column::Id effective_column_id, ObjectId object_key)
   : velox::connector::DataSource{},
     _memory_pool{memory_pool},
     _snapshot{snapshot},
     _db{db},
     _cf{cf},
     _row_type{std::move(row_type)},
     _column_ids(std::move(column_oids)),
+    _effective_column_id(std::move(effective_column_id)),
```
Dronplane, line 53: NIT: Do we really need move here?

MBkkt (approved):

server/connector/data_source.cpp
mkornaukhov: No, you are right, it has no semantic sense. I just did it by analogy

PR #155, opened 2026-01-14
SearchSink 

Dronplane:
Implements Insert, Update and Delete operations for IResearch index as a "writer" that could be integrated into connector DataSink.
This PR adds support for booleans, nulls, strings and numerics.
Complex types support would be added after we have proper metadata for index.
Same way for now all columns are indexed but later list of column ids would be set in the index meta together with the analyzers.
Update is implemented as Delete + Re-Insert operation in terms of index.

Closes https://github.com/serenedb/cluster/issues/6


libs/iresearch/include/iresearch/formats/empty_term_reader.hpp
```diff
@@ -46,7 +46,9 @@ class EmptyTermReader final : public irs::TermReader {
     return 0;
   }
 
-  size_t read_documents(bytes_view, std::span<doc_id_t>) const noexcept final {
+  size_t read_documents(bytes_view, std::span<doc_id_t>,
```
MBkkt: FunctionRef instead of ptr + ptr?

Dronplane: if you mean absl::FunctionRef - it can not be empty. And providing pointer to FunctionRef looked really ugly.

server/connector/search_sink_writer.hpp
```diff
@@ -0,0 +1,239 @@
...
+class SearchSinkUpdateWriter final : public SinkUpdateWriter,
+                                     public SearchSinkInsertBaseImpl,
+                                     public SearchSinkDeleteBaseImpl {
+ public:
+  SearchSinkUpdateWriter(irs::IndexWriter::Transaction& trx,
+                         velox::memory::MemoryPool& removes_pool)
+    : SearchSinkInsertBaseImpl(trx),
+      SearchSinkDeleteBaseImpl(trx, removes_pool) {}
+
+  void Init(size_t batch_size) override {
```
MBkkt: final for methods, according to current codestyle

server/connector/search_remove_filter.cpp
```diff
@@ -0,0 +1,108 @@
...
+    // might be skipped during flushed segment processing due to ticks (e.g.
+    // remove arrived before insert) but that is not a problem. As if we reached
+    // "tick" limit we anyway should not find anymore valid targets.
+
+    // 3. For segments with sorted field it should also work:
+    // E.G. if we have INSERT PK1 FIELD_SORTED_2 | DELETE PK1 | INSERT PK1
+    // FIELD_SORTED_1 Due to documents are sorted for storing after applying
+    // queries it will still see documents in insertion order.
+
+    auto terms = _pk_field->iterator(irs::SeekMode::RandomOnly);
```
MBkkt: I think we want to create iterator once time for this loop

Also we maybe want to have this iterator in the filter?

server/connector/search_remove_filter.cpp
```diff
@@ -0,0 +1,108 @@
...
+          return false;
+        }
+      }
+      return true;
+    };
+
+    auto doc = irs::doc_limits::eof();
+    auto found =
+      _pk_field->read_documents(terms->value(), std::span(&doc, 1), acceptor,
+                                static_cast<void*>(&_doc_masks));
```
MBkkt: this code doesn't correspond to acceptor.

Also I think we're can avoid some of these stuff, because we have actual mask of deleted in caller

server/connector/data_sink.cpp
```diff
@@ -1763,30 +1925,29 @@ void RocksDBDataSink::WriteArrayValue(const velox::BaseVector& input,
...
-  if (!status.ok()) {
-    SDB_THROW(rocksutils::ConvertStatus(status));
+}
+
+template<typename SubWriterType>
+void RocksDBDataSinkBase<SubWriterType>::WriteNull(std::string_view key) {
+  // empty slice denotes NULL value
+  rocksdb::Slice null_slice;
+  std::span<const rocksdb::Slice> null_slices{&null_slice, 1};
+  _data_writer.Write(null_slices, key);
```
MBkkt, line 2037: _data_writer special case, I'm not sure it's good idea. Because we plan to try different type of writer for `copy from` (sst file writer per column)

server/connector/search_remove_filter.hpp
```diff
@@ -0,0 +1,106 @@
...
+  }
+  velox::memory::MemoryPool& _memory_pool;
+  mutable const irs::SubReader* _segment{};
+  mutable std::array<const irs::DocumentMask*, 2> _doc_masks = {};
+  mutable const irs::TermReader* _pk_field{};
+  mutable size_t _pos{0};
+  mutable irs::DocAttr _doc;
+  // We need to store Pk copies as they would be applied on writer commit and
+  // should stay alive until that.
+  mutable ManagedVector<ManagedString> _pks;
```
MBkkt: I doubt we want to store pointer for allocator for each key

server/connector/search_remove_filter.cpp
Dronplane: I guess we don't need this part entirely. It is leftovers from previous implementation with postings reading.  Will just remove

server/connector/search_remove_filter.cpp
Dronplane: What do you meant "in caller"?

MBkkt: in index writer where we call execute

libs/iresearch/include/iresearch/formats/empty_term_reader.hpp
MBkkt: better than void* context (actual bug now)

Dronplane: BTW which bug?

server/connector/search_remove_filter.cpp
Dronplane: yes but here we must decide if we can pop out value or should continue with it in other segments.

server/connector/search_remove_filter.hpp
Dronplane: Good point. Will think about this, maybe make something arena like

Dronplane: Switched to usage of allocator. But need to decide how to have persistent pool until IndexWriterCommit

Dronplane: As discussed - for now use system allocator and created an issue for proper memory accounting later

libs/iresearch/include/iresearch/formats/formats_burst_trie.cpp
```diff
@@ -2767,47 +2767,41 @@ class FieldReaderImpl final : public FieldReader {
...
         SDB_ASSERT(false);
-        return 0;
+        return;
       }
 
-      auto begin = docs.begin();
-
-      for (auto end = docs.end(); begin != end && docs_it->next(); ++begin) {
-        *begin = doc->value;
+      while (docs_it->next() && acceptor(doc->value)) {
```
gnusi: NIT: we can now use `advance()`, e.g.
```
while (auto doc = docs_it->advance()) { 
  if (doc_limits::eof(doc) || !acceptor(doc)) {
    break;
  }
}
```

Dronplane: fixes

server/connector/search_sink_writer.cpp
```diff
@@ -0,0 +1,333 @@
...
+          if (cell_slices.size() == 1 && cell_slices.front().empty()) {
+            _null_field.SetNullValue();
+            return _null_field;
+          }
+          return write_func(full_key, cell_slices, field);
+        };
+    };
+
+  if constexpr (Kind == velox::TypeKind::UNKNOWN) {
+    _current_writer = MakeIndexWriter(
```
gnusi, line 127: Do we really need to use make_nullable_writer_func for NULL?

Dronplane: we don't. And for UNKNOWN nullable wrapper is never used.

server/connector/common.h
```diff
@@ -36,11 +39,16 @@ enum class ValueFlags : uint8_t {
 
 ENABLE_BITMASK_ENUM(ValueFlags);
 
+// TODO(Dronplane) unify with key?
+constexpr std::string_view kStringPrefix{"\0", 1};
```
MBkkt:
```suggestion
inline constexpr std::string_view kStringPrefix{"\0", 1};
```

server/connector/data_sink.hpp
```diff
@@ -31,28 +31,37 @@
...
-  void appendData(velox::RowVectorPtr input) final;
   bool finish() final;
   std::vector<std::string> close() final;
   void abort() final;
   Stats stats() const final;
 
- private:
+ protected:
+  void PrepareKeyBuffers(const velox::RowVectorPtr& input);
+  template<bool SkipPrimaryKeyColumns>
```
MBkkt:
```suggestion
  void PrepareKeyBuffers(const velox::RowVectorPtr& input);

  template<bool SkipPrimaryKeyColumns>
```

server/connector/sink_writer_base.hpp
```diff
@@ -0,0 +1,70 @@
...
+class SinkInsertWriter {
+ public:
+  SinkInsertWriter() = default;
+  virtual ~SinkInsertWriter() = default;
+
+  virtual void Init(size_t batch_size) {}
+
+  // returns true if writer is interested in this column
+  virtual bool SwitchColumn(velox::TypeKind kind, bool have_nulls,
+                            sdb::catalog::Column::Id column_id) {
```
MBkkt, line 39:
```suggestion
                            catalog::Column::Id column_id) {
```

server/connector/search_sink_writer.cpp
```diff
@@ -0,0 +1,333 @@
...
+  kDefaultPoolSize);
+
+void SetNameToBuffer(std::string& name_buffer, Column::Id column_id) {
+  SDB_ASSERT(name_buffer.size() >= sizeof(column_id));
+  absl::big_endian::Store(name_buffer.data(), column_id);
+}
+
+}  // namespace
+
+namespace sdb::connector::search {
```
MBkkt: can you move anon namespace in `sdb::connector::search {`, in such case you probably not needed to write 3 using namespace

server/connector/search_remove_filter.hpp
```diff
@@ -0,0 +1,100 @@
...
+class SearchRemoveFilter : public SearchRemoveFilterBase {
+ public:
+  explicit SearchRemoveFilter(size_t batch_size) { _pks.reserve(batch_size); }
+
+  void reset() {
+    _pos = 0;
+    _pks.clear();
+  }
+
+  irs::doc_id_t advance() override;
```
MBkkt: Can it be final?

server/connector/search_remove_filter.hpp
```diff
@@ -0,0 +1,100 @@
...
+#include <iresearch/analysis/token_attributes.hpp>
+#include <iresearch/search/filter.hpp>
+
+#include "common.h"
+
+namespace sdb::connector::search {
+
+// this matches generated PK column id but it is not stored so should be fine.
+constexpr inline std::string_view kPkFieldName{
+  "\xff\xff\xff\xff\xff\xff\xff\xff", 8};
```
MBkkt: `inline constexpr` instead of `constexpr inline`

Also it's not clear from comment why pk field name should be 8 bytes length, why it can't be single byte?

libs/iresearch/include/iresearch/index/index_writer.cpp
```diff
@@ -211,7 +211,8 @@ void RemoveFromExistingSegment(DocumentMask& deleted_docs,
     return;  // skip invalid prepared filters
   }
 
-  auto itr = prepared->execute({.segment = reader});
+  auto itr =
+    prepared->execute({.segment = reader, .pending_docs_mask = &deleted_docs});
```
MBkkt, line 215: Maybe instead of modify execute we want to make pending docs mask attribute?
So here we get this attribute and if it's not nullptr we set this mask?

server/connector/data_sink.cpp
```diff
@@ -46,48 +45,164 @@
...
+    for (size_t i = 0; i < _all_column_ids.size(); ++i) {
+      if (!absl::c_contains(_column_ids, _all_column_ids[i])) {
+        _rewrite_columns_idxs.push_back(i);
+      }
+    }
+
+    // Sort columns ascending to match rocksdb storage order.
+    // This allows us to use single iterator pass for all columns and with only
+    // forward seeks
+    std::sort(_rewrite_columns_idxs.begin(), _rewrite_columns_idxs.end(),
```
MBkkt:
```suggestion
    absl::c_sort(_rewrite_columns_idxs,
```

server/connector/data_sink.cpp
```diff
@@ -46,48 +45,164 @@
...
+  // TODO(Dronplane) same condition in ctor. Maybe make it inline method?
+  if (_index_writers.empty()) {
+    rows_range.push_back(
+      velox::IndexRange{0, static_cast<velox::vector_size_t>(input->size())});
+  } else {
+    // for same purpose as columns in ctor we sort the keys
+    const auto num_rows = input->size();
+    rows_idxs.resize(num_rows);
+    absl::c_iota(rows_idxs, 0);
+    std::sort(rows_idxs.begin(), rows_idxs.end(),
```
MBkkt:
```suggestion
    absl::c_sort(rows_idxs,
```

server/connector/data_sink.cpp
```diff
@@ -46,48 +45,164 @@
...
 
-}  // namespace
+void RocksDBUpdateDataSink::appendData(velox::RowVectorPtr input) {
+  PrepareKeyBuffers(input);
+  std::vector<velox::IndexRange> rows_range;
+  std::vector<velox::vector_size_t> rows_idxs;
+  // TODO(Dronplane) same condition in ctor. Maybe make it inline method?
+  if (_index_writers.empty()) {
+    rows_range.push_back(
+      velox::IndexRange{0, static_cast<velox::vector_size_t>(input->size())});
```
MBkkt:
```suggestion
    rows_range.emplace_back(0, static_cast<velox::vector_size_t>(input->size()));
```

server/connector/data_sink.cpp
```diff
@@ -46,48 +45,164 @@
...
+    absl::c_iota(rows_idxs, 0);
+    std::sort(rows_idxs.begin(), rows_idxs.end(),
+              [&](velox::vector_size_t left, velox::vector_size_t right) {
+                return _keys_buffers[left] < _keys_buffers[right];
+              });
+    // we must match order of keys in rewritten columns
+    rows_range.reserve(num_rows);
+    for (auto row_idx : rows_idxs) {
+      rows_range.push_back(
+        velox::IndexRange{static_cast<velox::vector_size_t>(row_idx), 1});
```
MBkkt: emplace_back

server/connector/data_sink.cpp
```diff
@@ -46,48 +45,164 @@
...
+    }
 
-namespace sdb::connector {
+    // Index update implies delete + re-insert.
```
MBkkt: can we don't make it generic? because it can be index that supports patch update

server/connector/search_remove_filter.hpp
Dronplane: Actually it can. We need only to never match existing field name. Let's make it one byte then.

server/connector/data_sink.cpp
Dronplane: Close as code here changed.

server/connector/data_sink.cpp
Dronplane: close as code changed after updating to main

server/connector/data_sink.cpp
Dronplane: code changed

server/connector/data_sink.cpp
Dronplane: code changed

libs/iresearch/include/iresearch/index/index_writer.cpp
Dronplane: Let's address general approach to deletes in a separate PR

server/connector/data_sink.cpp
Dronplane: When we have this index - we could refine our interfaces. For now it is now yet clear what and where would be needed

server/connector/data_sink.hpp
Dronplane: code changed

server/connector/search_sink_writer.cpp
Dronplane: well we can reduce from 3 to 1 using. Ok

PR #156, opened 2026-01-14
Introducing of recovery tests

mkornaukhov:
Add mechanism for recovery tests, a single example test and new failure point.
Just run (with enough privileges) `./tests/sqllogic/run_in_docker.sh recovery` and it will launch recovery tests.


tests/sqllogic/recovery/crash_on_packet.test
```diff
@@ -0,0 +1,40 @@
...
+
+
+# Next sent packet will trigger serened crash, so connection will be closed
+query error connection closed
+SELECT 1;
+
+
+# ==== TRY TO RECONNECT ====
+connection new
+query ok retry 10 backoff 500ms
```
Dronplane: Can we have this "reconnect" as some command with parameters set somewhere in config?
Use case is : we have more tests and at some point we start getting flaky failures due to not enough timeout for restart (agents on virtual machines might be slow) and we need a convenient way to increase timeout in general for all tests

mkornaukhov: I think we can pass it through env variables. Good point, I'll fix it.

mkornaukhov: Fixed, see here https://github.com/serenedb/sqllogictest-rs/pull/21

Dronplane (approved):

PR #158, opened 2026-01-15
Update primary keys

### LINKED ISSUE #162: `ConnectionPoolTest.release_multiple_endpoints_one` is flaky
It's periodically failed when related code is not touched at all. Need to investigate.

mkornaukhov:
# Update Primary Keys Support

## Overview

This PR implements support for updating primary key columns in UPDATE statements. Previously, SereneDB required deleting and re-inserting rows to change primary keys - now users can directly update PK columns:

```sql
UPDATE users SET user_id = 12345 WHERE user_id = 999;
UPDATE orders SET order_id = new_id, customer_id = new_customer WHERE order_id = old_id;
```

## Key points
- **Code structure** : Refactored RocksDBDataSink into 3 classes - base class + separate INSERT/UPDATE sinks for cleaner code
- **Column naming**: Uses null byte prefix (upd_\0column) for internal columns to avoid user naming collisions
- **Locking strategy**: Locks both old and new PK rows, deletes old data, copies unchanged columns (without creating Velox vectors), writes updates

## Misc
Increase number of attempts for flaky test, closes #162

### DIFF
--- a/server/connector/serenedb_connector.hpp
+++ b/server/connector/serenedb_connector.hpp
@@ -313,7 +313,7 @@ class SereneDBConnectorInsertTableHandle final
       _update_pk{basics::downCast<RocksDBTable>(*table).IsUsedForUpdatePK()} {
     GetTransaction().AddRocksDBWrite();
     if (_update_pk) {
-      GetTransaction().AddRocksDBRead();  // for update of primary keys
+      GetTransaction().AddRocksDBRead();
     }
   }
 
--- a/tests/server/connector/data_sink_test.cpp
+++ b/tests/server/connector/data_sink_test.cpp
@@ -2603,7 +2603,6 @@ TEST_F(DataSinkTest, test_insertDeleteConflict) {
   rocksdb::WriteOptions wo;
   std::unique_ptr<rocksdb::Transaction> transaction_delete{
     _db->BeginTransaction(wo, trx_opts, nullptr)};
-
   sdb::connector::RocksDBDeleteDataSink delete_sink(
     *transaction_delete, *_cf_handles.front(), row_type, kObjectKey,
     column_ids);


mkornaukhov:
Do not decode + encode extra columns

server/connector/key_utils.hpp
```diff
@@ -72,6 +72,18 @@ inline void SetupColumnForKey(std::string& buf, catalog::Column::Id column_id) {
...
+// 'object_id | reserved for column_id | pk'
+// and fills pk
+inline void SetupPKForKey(const velox::RowVectorPtr& input,
+                          const std::vector<velox::column_index_t>& pk_columns,
+                          velox::vector_size_t row_idx, std::string& buf) {
+  SDB_ASSERT(buf.size() >= sizeof(ObjectId) + sizeof(catalog::Column::Id));
+  buf.resize(sizeof(ObjectId) + sizeof(catalog::Column::Id));
+  primary_key::Create(*input, pk_columns, row_idx,
+                      buf);  // Or just store 2 arrays?
+}
```
MBkkt: it looks wrong

server/connector/serenedb_connector.hpp
```diff
@@ -315,12 +316,17 @@ class SereneDBConnectorInsertTableHandle final
...
+    return _num_of_rows_affected;
+  }
+
  private:
   axiom::connector::ConnectorSessionPtr _session;
   axiom::connector::TablePtr _table;
   axiom::connector::WriteKind _kind;
   query::Transaction& _transaction;
   std::vector<velox::connector::ColumnHandlePtr> _row_id_handles;
+  mutable std::atomic<size_t> _num_of_rows_affected = 0;
```
MBkkt: I don't think we want to have atomic about this

server/pg/sql_analyzer_velox.cpp
```diff
@@ -1706,24 +1712,36 @@ void SqlAnalyzer::ProcessUpdateStmt(State& state, const UpdateStmt& stmt) {
...
+                      ERR_MSG("column \"", target.name, "\" of relation \"",
+                              table_name, "\" does not exist"));
     }
+
     SDB_ASSERT(it->second);
     const auto& column = *(it->second);
+    column_names.emplace_back(
+      pk_column_names.contains(target.name)
+        ? catalog::Column::GenerateUpdateName(target.name)
+        : std::string(target.name));
```
MBkkt: it doesn't look ok, we don't want such logic

server/catalog/table_options.h
```diff
@@ -20,6 +20,7 @@
 
 #pragma once
```
MBkkt, line 21: 1. read all
2. create new columns
ReplaceRowDataSink
4. drop old row
5. create new row

server/connector/data_sink.cpp
```diff
@@ -64,70 +64,230 @@ void WriteNull(rocksdb::Transaction& trx, rocksdb::ColumnFamilyHandle& cf,
...
+  const auto num_rows = input->size();
+  const auto num_columns = input->childrenSize();
+
+  _keys_buffers.clear();
+  _keys_buffers.reserve(num_rows);
+
+  if (_updating_pk) {
+    _updated_keys_buffers.clear();
+    _updated_keys_buffers.reserve(num_rows);
+  }
```
MBkkt: old + new?

server/connector/data_sink.cpp
```diff
@@ -64,70 +64,230 @@ void WriteNull(rocksdb::Transaction& trx, rocksdb::ColumnFamilyHandle& cf,
...
+                         _keys_buffers[l_ind].end()};
+      auto r_pk =
+        std::string_view{_keys_buffers[r_ind].begin() + sizeof(ObjectId) +
+                           sizeof(catalog::Column::Id),
+                         _keys_buffers[r_ind].end()};
+      return l_pk < r_pk;
+    });
+
+    containers::FlatHashSet<catalog::Column::Id> updated_columns(
+      _column_ids.begin() + _key_childs.size(), _column_ids.end());
```
MBkkt: this is unused

server/pg/sql_analyzer_velox.cpp
MBkkt, line 1774: object.SetTableUsedFor....;

server/connector/key_utils.hpp
MBkkt:
```suggestion
```

server/connector/data_sink.cpp
```diff
@@ -64,70 +64,230 @@ void WriteNull(rocksdb::Transaction& trx, rocksdb::ColumnFamilyHandle& cf,
...
+                             _keys_buffers.emplace_back());
+
+    // Prepare new keys
+    if (_updating_pk) {
+      key_utils::MakeColumnKey(input, _updated_key_childs, row_idx, table_key,
+                               lock_row, _updated_keys_buffers.emplace_back());
+    }
+  }
+
+  // Delete all old value
```
MBkkt: In such case `update t set a = a + 1, pk = pk + 1`

now we have
```
table read (pk, a)
projection (new_pk, new_a)
table write (read all from iterator, delete them, write new from input)
```

We want to optimize this to something like this:

```
for all column ids {
   if (update_column) {
     Delete From Input
     Put From Input
   } else {
     Delete From Iterator
     Put From Iterator
   }
}
```

server/connector/data_sink.cpp
```diff
@@ -64,70 +64,230 @@ void WriteNull(rocksdb::Transaction& trx, rocksdb::ColumnFamilyHandle& cf,
...
+
+        auto& updated_key = _updated_keys_buffers[row_idx];
+        key_utils::SetupColumnForKey(updated_key, col_id);
+        status = _transaction.Put(&_cf, updated_key, it->value());
+        if (!status.ok()) {
+          SDB_THROW(rocksutils::ConvertStatus(status));
+        }
+      }
+    }
+
```
MBkkt: pk=> key
pk => key(pk) value(pk)

mkornaukhov:
- Plan test
- Sort Column::ID not as strings -- better linear check for search
- Check with SDB_DEV that rows are sorted
- N (~2? on for column ref, one for uniqueness) symbols 
- Do not use atomic for affected rows count, use transaction code, div by 2 if needed
- In MakeTableWrite (or after make method setting bool) that updating primary key

server/connector/key_utils.hpp
mkornaukhov: I've dropped it

server/connector/serenedb_connector.hpp
mkornaukhov: Removed

server/pg/sql_analyzer_velox.cpp
mkornaukhov: We've decided to leave it IIRC, just used another char for that. I've picked '$' for now. This code is not changed.

server/catalog/table_options.h
mkornaukhov: We've decided to fix current solution, not to use this one, so I resolve this convo.

server/connector/data_sink.cpp
mkornaukhov: Renamed in another way

server/connector/data_sink.cpp
mkornaukhov: Removed

server/pg/sql_analyzer_velox.cpp
mkornaukhov: Done

server/connector/key_utils.hpp
mkornaukhov: Done

server/connector/data_sink.cpp
mkornaukhov: Just annotation, not actual requested change

server/connector/data_sink.cpp

MBkkt (approved):

PR #167, opened 2026-01-21
Fix TableStats

codeworse:
Make TableStats variable unnecessary
issue: https://github.com/serenedb/serenedb/issues/168


mkornaukhov:
Let's add explanation of this warning and why we cannot consistently write stats in issue

PR #170, opened 2026-01-21
Optimize range primary key updates

mkornaukhov:
## Overview
In case of range primary key updates use `Next()` instead of `Seek()` method of RocksDB iterator.
With these updates, we have improved the average latency of about 27%.
Note that if we would try to do one step of `Next()` unconditionally, then we'd loose about 13% latency on average.

## Reproduce
1. In psql setup table:
```sql
create table t1(a integer primary key, b integer, c integer);

insert into t1 select a2.x * 1000 + a3.x, a2.x, a3.x from
  generate_series(1, 1000) as a2(x),
  generate_series(1, 1000) as a3(x);
```
2. Setup in committed script update query. 
- For range query:
```sql
update t1 set a = a;
```
-  For not a range query:
```sql
update t1 set a = a where a % 30 = 1;
```

### DIFF
--- a/server/connector/data_sink.cpp
+++ b/server/connector/data_sink.cpp
@@ -240,20 +240,14 @@ void RocksDBUpdateDataSink::appendData(velox::RowVectorPtr input) {
     std::unique_ptr<rocksdb::Iterator>(_db.NewIterator(read_options, &_cf));
 
   bool is_range = true;
-  auto seek_to_key = [&](std::string_view key, bool seek_allowed_for_range) {
-    if (!it->Valid()) [[unlikely]] {
-      it->Seek(key);
-      return;
-    }
-
-    if (is_range) {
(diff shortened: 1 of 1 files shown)


server/connector/data_sink.cpp
```diff
@@ -239,6 +239,24 @@ void RocksDBUpdateDataSink::appendData(velox::RowVectorPtr input) {
...
+    if (is_range) {
+      it->Next();
+      if (it->key() == key) {
+        return;
+      }
+      is_range = seek_allowed_for_range;
+    }
+
+    it->Seek(key);
+  };
```
MBkkt, line 252:
```suggestion
  bool is_range = true;
  auto seek_to_key = [&](std::string_view key, bool use_seek) {
    if (is_range && !use_seek) {
      it->Next();
      if (it->key() == key) {
        return;
      }
      is_range = false;
    }
    it->Seek(key);
  };
```

PR #174, opened 2026-01-23
Conflict detection on INSERT 

### LINKED ISSUE #77: Support conflict detection on inserts and `do nothing` behavior
Insert without conflict checking => aka replace

Insert with conflict checking:

- get key lock should return error for conflict
- make first column key
- check there's no such key in rocksdb
- if no conflict => just continue
- if do nothing on conflict => clear pk and skip it later
- else throw

### LINKED ISSUE #184: Change the way affected rows are counted
### Overview
For now we use `rocksdb::Transaction`'s number of locked keys (dividing by 2 in case of updating PK). I suggest to change this approach because it's not correct.

### Case 1: multi query statement
Query (line is one statement):
```SQL
create table t1(a int primary key);
insert into t1 values(1); insert into t1 values(2);
```

SereneDB response to the last statement:
```
INSERT 0 1
INSERT 0 2
```

PostgreSQL response:
```
INSERT 0 1
INSERT 0 1
```

### Case 2: multiple statements in transaction
Same as above, but make inserts as separate statements inside transaction ( inside `begin` and `commit`).

### Case 3: `ON CONFLICT DO NOTHING`
In #174 (not merged yet) we introduce support for keeping old values during insert if conflict is present. In such case, we still lock corresponding rocksdb key, but actually the row is not affected. 

### Proposed solution
Case 3 shows that number of locked keys is not appropriate value. Case 1 and 2 shows are that counter shouldn't be per transaction, but per **query**. 
So I suppose that we should store counter inside query/query context and pass reference to it inside data sink.

mkornaukhov:
### Overview
This PR adds write conflict handling for `INSERT` operations. It introduces a `WriteConflictPolicy` enum with three strategies: 
- `EmitError` (default, throws on duplicate keys)
- `DoNothing` (skips conflicting rows)
- `Replace` (upserts existing rows)

The implementation adds conflict detection in `RocksDBSinkWriter`.
A new session variable `sdb_write_conflict_policy` allows users to set the default behavior, and SQL syntax `ON CONFLICT DO NOTHING` is now supported  which overrides the session setting to use `DoNothing` policy.

### Benches
I've measured inserts on the following query
```sql
insert into t1 select a2.x * 1000 + a3.x, a2.x, a3.x from generate_series(1, 1000) as a2(x), generate_series(1, 1000) as a3(x);
```
I have done 4 measurements:
1. Main branch (only replace policy)
2. Current branch with replace policy
3. Current branch with do nothing policy
4. Current branch with emit error policy

Results are following:
- main replace is same as current branch with replace policy
- emit error and do nothing policy is **40%** worse than replace policy
So some optimizations will be made as a part of #185

### Misc
Closes #184 
Closes #77
Followup #185

### DIFF
--- a/server/connector/data_sink.cpp
+++ b/server/connector/data_sink.cpp
@@ -64,6 +64,12 @@ template<typename T>
 std::string VeloxValueToString(T val) {
   if constexpr (sdb::type::kIsOneOf<T, velox::StringView, velox::Timestamp>) {
     return static_cast<std::string>(val);
+  } else if constexpr (std::is_same_v<T, velox::int128_t>) {
+    std::string result;
+    basics::StrResize(result, absl::numbers_internal::kFastToBuffer128Size);
+    char* end = absl::numbers_internal::FastIntToBuffer(val, result.data());
+    result.resize(end - result.data());
+    return result;
   } else {
     return absl::StrCat(val);
   }
@@ -85,8 +91,8 @@ std::string GetPKVeloxValue(velox::VectorPtr vec, velox::vector_size_t idx) {
 std::string FormatKeyValue(const velox::RowVectorPtr& input,
                            velox::column_index_t key_idx,
                            velox::vector_size_t row_idx) {
-  auto vector = input->childAt(key_idx);
-  auto type = input->rowType()->childAt(key_idx);
+  const auto& vector = input->childAt(key_idx);
+  const auto& type = input->rowType()->childAt(key_idx);
 
   if (type->isPrimitiveType()) {
     return VELOX_DYNAMIC_SCALAR_TYPE_DISPATCH(
--- a/server/connector/serenedb_connector.hpp
+++ b/server/connector/serenedb_connector.hpp
@@ -585,7 +585,8 @@ class SereneDBConnector final : public velox::connector::Connector {
         auto handle = table.columnMap().find(real_name);
         SDB_ASSERT(handle != table.columnMap().end(),
                    "RocksDBDataSink: can't find column handle for ", real_name);
-        auto column = basics::downCast<const SereneDBColumn>(handle->second);
+        const auto* column =
+          basics::downCast<const SereneDBColumn>(handle->second);
         columns.emplace_back(column->Id(), column->name());
       }
       return irs::ResolveBool(
@@ -673,7 +674,8 @@ class SereneDBConnector final : public velox::connector::Connector {
         auto handle = table.columnMap().find(col);
         SDB_ASSERT(handle != table.columnMap().end(),
                    "RocksDBDataSink: can't find column handle for ", col);
-        auto column = basics::downCast<const SereneDBColumn>(handle->second);
+        const auto* column =
+          basics::downCast<const SereneDBColumn>(handle->second);
         columns.emplace_back(column->Id(), column->name());
       }
       auto& rocksdb_transaction = transaction.EnsureRocksDBTransaction();


tests/sqllogic/sdb/pg/settings/show_all.test
```diff
@@ -7,6 +7,7 @@ join_order_algorithm	cost	Sets the join algorithm to be used for query execution
 search_path	"$user", public	Sets the schema search order for names that are not schema-qualified.
 extra_float_digits	1	Sets the number of digits displayed for floating-point values.
 bytea_output	hex	Sets the output format for values of type bytea. Valid values are 'hex' (the default) and 'escape' (the traditional PostgreSQL format).
+sdb_write_conflict_policy	error	Sets the write conflict policy. Valid values are 'error' (the default), 'keep_old' (like DO NOTHING) and 'update'
```
MBkkt: I suggest different names: sdb_insert_on_conflict: error, do_nothing, replace

server/pg/sql_analyzer_velox.cpp
```diff
@@ -1581,9 +1581,30 @@ void SqlAnalyzer::ProcessInsertStmt(State& state, const InsertStmt& stmt) {
...
+  auto conflict_policy = config.Get<VariableType::SdbWriteConflictPolicy>(
+    "sdb_write_conflict_policy");
   if (stmt.onConflictClause) {
-    SDB_THROW(ERROR_NOT_IMPLEMENTED,
-              "ON CONFLICT clause is not implemented yet");
+    if (stmt.onConflictClause->action == ONCONFLICT_UPDATE) {
+      SDB_THROW(ERROR_NOT_IMPLEMENTED,
+                "ON CONFLICT DO UPDATE SET action is not implemented yet");
+    }
+    if (list_head(stmt.onConflictClause->targetList) != nullptr) {
```
MBkkt: list head is unnecessary here?

server/pg/sql_analyzer_velox.cpp
```diff
@@ -1581,9 +1581,30 @@ void SqlAnalyzer::ProcessInsertStmt(State& state, const InsertStmt& stmt) {
...
+    if (stmt.onConflictClause->infer != nullptr) {
+      SDB_THROW(ERROR_NOT_IMPLEMENTED,
+                "ON CONFLICT with infer clause is not implemented yet");
+    }
+    if (stmt.onConflictClause->whereClause != nullptr) {
+      SDB_THROW(ERROR_NOT_IMPLEMENTED,
+                "ON CONFLICT WHERE clause is not implemented yet");
+    }
+
+    SDB_ASSERT(stmt.onConflictClause->action == ONCONFLICT_NOTHING, "Wtf?");
```
MBkkt: comment

server/pg/sql_analyzer_velox.cpp
```diff
@@ -1581,9 +1581,30 @@ void SqlAnalyzer::ProcessInsertStmt(State& state, const InsertStmt& stmt) {
...
+    conflict_policy = WriteConflictPolicy::KeepOld;
```
MBkkt: I don't think it should be default behavior

server/connector/serenedb_connector.hpp
```diff
@@ -254,13 +254,22 @@ class RocksDBTable final : public axiom::connector::Table {
...
+  }
+
  private:
   std::vector<std::unique_ptr<SereneDBTableLayout>> _layout_handles;
   std::vector<const axiom::connector::TableLayout*> _layouts;
   velox::RowTypePtr _pk_type;
   ObjectId _table_id;
   query::Transaction& _transaction;
   catalog::TableStats _stats;
+  WriteConflictPolicy _conflict_policy{WriteConflictPolicy::Error};
```
```suggestion
  WriteConflictPolicy _conflict_policy = WriteConflictPolicy::Error;
```

server/connector/rocksdb_sink_writer.hpp
```diff
@@ -47,15 +50,35 @@ class RocksDBSinkWriterBase {
...
+    _row_conflict_mask.resize(num_rows, false);
+  }
+  void ResetRowId() { _row_id = 0; }
+
+ private:
+  void ConfigureReadOptions();
+  bool CheckConflict(const rocksdb::Slice& key_slice);
+
+  WriteConflictPolicy _conflict_policy;
+  boost::dynamic_bitset<> _row_conflict_mask;
```
MBkkt: remove this please, it doesn't make any sense

server/connector/rocksdb_sink_writer.cpp
```diff
@@ -24,11 +24,60 @@
...
+  bool conflict = CheckConflict(key_slice);
+
+  if (conflict) {
+    switch (_conflict_policy) {
+      case WriteConflictPolicy::Update:
+        break;
+      case WriteConflictPolicy::KeepOld:
+        return;
+      case WriteConflictPolicy::Error:
+        SDB_THROW(ERROR_SERVER_UNIQUE_CONSTRAINT_VIOLATED,
```
MBkkt: it should be postgres error

server/connector/rocksdb_sink_writer.cpp
```diff
@@ -24,11 +24,60 @@
...
+                  "TODO: make sql error");
+        break;
+      default:
+        SDB_UNREACHABLE();
+    }
+  }
+
   rocksdb::Status status;
-  SDB_ASSERT(!cell_slices.empty());
```
MBkkt, line 31: why this removed

server/connector/rocksdb_sink_writer.cpp
```diff
@@ -24,11 +24,60 @@
...
+  _read_options.async_io = true;
+  _read_options.snapshot = _transaction.GetSnapshot();
+}
+
+bool RocksDBSinkWriter::CheckConflict(const rocksdb::Slice& key_slice) {
+  if (_use_conflict_mask) {
+    return _row_conflict_mask[_row_id++];
+  }
+
+  if (_conflict_policy == WriteConflictPolicy::Update) {
```
Dronplane: Should this be first check as most lightweight?

server/connector/data_sink.cpp
```diff
@@ -171,12 +172,19 @@ void RocksDBInsertDataSink::appendData(velox::RowVectorPtr input) {
...
   for (velox::column_index_t i = 0; i < num_columns; ++i) {
-    if (_column_ids[i] != catalog::Column::kGeneratedPKId) {
-      WriteInputColumn(_column_ids[i], i, *input, all_rows_range);
+    if (_column_ids[i] == catalog::Column::kGeneratedPKId) {
+      SDB_ASSERT(i + 1 == num_columns,
+                 "RocksDBDataSink: generated primary column should be the last "
+                 "one in the input vectors");
+      continue;
     }
+    _data_writer.UseConflictMask(i > 0);
```
Dronplane: Should that be just `_data_writer.UseConflictMask(true);` after first call to `WriteInputColumn`?

server/connector/rocksdb_sink_writer.cpp
mkornaukhov: I've made an mapping in `pg_comm_task.cpp`

server/connector/rocksdb_sink_writer.cpp
mkornaukhov: fixed

tests/sqllogic/any/pg/simple/update_pk.test
```diff
@@ -62,6 +62,9 @@ Fragment 0:  numWorkers=0:
 statement error multiple assignments to same column "a"
 UPDATE single_pk SET a = 42, a = 100500 WHERE b = 10;
 
+statement error duplicate key value violates unique constraint
+UPDATE single_pk SET a = 42;
```
MBkkt: I think we want to check `replace` and `do nothing` as well

tests/sqllogic/sdb/pg/simple/insert_conflict.test
```diff
@@ -0,0 +1,97 @@
...
+2	20
+3	30
+
+statement count 0
+INSERT INTO t1 VALUES (1, 10), (2, 20), (3, 30) ON CONFLICT DO NOTHING;
+
+statement error duplicate key value violates unique constraint
+INSERT INTO t1 VALUES (1, 10), (2, 20), (3, 30);
+
+statement error duplicate key value violates unique constraint
```
MBkkt: all statement error should be multiline errors at least for postgres errors

tests/sqllogic/sdb/pg/simple/insert_conflict.test
```diff
@@ -0,0 +1,97 @@
...
+SELECT * FROM t1 ORDER BY a;
+----
+a	b
+1	10
+2	20
+3	30
+4	400
+5	500
+
+statement error duplicate key value violates unique constraint
```
MBkkt: same

tests/sqllogic/any/pg/simple/insert_conflict.test
```diff
@@ -0,0 +1,75 @@
...
+
+# We do not support 'Read Your Own Writes' yet
+onlyif serenedb
+query
+SELECT * FROM t2 ORDER BY a;
+----
+a	b
+
+# But still cannot insert into table as values is in transaction
+statement error
```
MBkkt, line 80: same

tests/sqllogic/any/pg/simple/insert_conflict.test
```diff
@@ -0,0 +1,75 @@
...
+SELECT * FROM t1 ORDER BY a;
+----
+a	b
+1	10
+2	20
+3	30
+4	400
+5	500
+
+statement error duplicate key value violates unique constraint
```
MBkkt: same

server/connector/data_sink.cpp
```diff
@@ -58,19 +58,23 @@ RocksDBDataSinkBase<SubWriterType>::RocksDBDataSinkBase(
   velox::memory::MemoryPool& memory_pool, ObjectId object_key,
   std::span<const velox::column_index_t> key_childs,
   std::vector<catalog::Column::Id> column_ids,
+  WriteConflictPolicy conflict_policy, size_t* number_of_rows_affected,
```
MBkkt:
```suggestion
  WriteConflictPolicy conflict_policy, size_t& number_of_rows_affected,
```
I don't think we want to support nullptr

server/connector/data_sink.cpp
```diff
@@ -426,17 +460,20 @@ void RocksDBDataSinkBase<SubWriterType>::WriteBiasedColumn(
       for (const auto& range : ranges) {
         const auto range_end = range.begin + range.size;
         for (velox::vector_size_t idx = range.begin; idx < range_end; ++idx) {
-          const auto& key = SetupRowKey(row_id++, original_idx);
+          auto* key = SetupRowKey(row_id++, original_idx);
+          if (key == nullptr) {
```
MBkkt: codestyle
```suggestion
          const auto* key = SetupRowKey(row_id++, original_idx);
          if (!key) {
```

server/connector/data_sink.cpp
```diff
@@ -396,16 +427,19 @@ void RocksDBDataSinkBase<SubWriterType>::WriteFlatColumn(
     for (const auto& range : ranges) {
       const auto range_end = range.begin + range.size;
       for (velox::vector_size_t idx = range.begin; idx < range_end; ++idx) {
-        const auto& key = SetupRowKey(row_id++, original_idx);
+        auto* key = SetupRowKey(row_id++, original_idx);
+        if (key == nullptr) {
+          continue;
+        }
```
MBkkt: codestyle

server/connector/data_sink.cpp
```diff
@@ -480,7 +517,11 @@ void RocksDBDataSinkBase<SubWriterType>::WriteDictionaryColumn(
                       original_idx.subspan(current, sub_ranges.size()));
           sub_ranges.clear();
         }
-        WriteNull(SetupRowKey(row_id++, original_idx));
+        auto* key = SetupRowKey(row_id++, original_idx);
+        if (key == nullptr) {
+          continue;
+        }
```
MBkkt: codestyle

server/connector/data_sink.cpp
```diff
@@ -505,9 +546,12 @@ void RocksDBDataSinkBase<SubWriterType>::WriteComplexColumn(
     int32_t begin = ranges[i].begin;
     int32_t end = begin + ranges[i].size;
     for (int32_t offset = begin; offset < end; ++offset) {
-      const auto& key = SetupRowKey(row_id++, original_idx);
+      auto* key = SetupRowKey(row_id++, original_idx);
+      if (key == nullptr) {
+        continue;
+      }
```
MBkkt: codestyle

server/connector/data_sink.cpp
```diff
@@ -552,8 +596,11 @@ void RocksDBDataSinkBase<SubWriterType>::WriteConstantColumn(
     for (const auto& range : ranges) {
       const auto end = range.begin + range.size;
       for (int32_t offset = range.begin; offset < end; ++offset) {
-        const auto& key = SetupRowKey(row_id++, original_idx);
-        WriteRowSlices(key);
+        auto* key = SetupRowKey(row_id++, original_idx);
+        if (key == nullptr) {
+          continue;
+        }
```
MBkkt: codestyle

server/connector/data_sink.cpp
```diff
@@ -166,18 +175,38 @@ void RocksDBInsertDataSink::appendData(velox::RowVectorPtr input) {
         }
       },
       _store_keys_buffers.emplace_back());
+
+    // For early conflict detection
+    key_utils::SetupColumnForKey(_store_keys_buffers.back(),
+                                 _column_ids.front());
```
MBkkt: we don't need this for "replace your own writes" or for update not pk

server/connector/rocksdb_sink_writer.cpp
```diff
@@ -24,6 +24,55 @@
...
+
+size_t RocksDBSinkWriter::HandleConflicts(primary_key::Keys& keys) {
+  if (_conflict_policy == WriteConflictPolicy::Replace) {
+    // Optimize out reading
+    return 0;
+  }
+
+  ConfigureReadOptions();
+  size_t skipped_cnt = 0;
+  for (auto& key : keys) {
```
MBkkt:
```suggestion
  for (const auto& key : keys) {
```

server/connector/rocksdb_sink_writer.cpp
```diff
@@ -24,6 +24,55 @@
 
 namespace sdb::connector {
 
+void RocksDBSinkWriter::ConfigureReadOptions() {
+  _read_options.async_io = true;
+  _read_options.snapshot = _transaction.GetSnapshot();
+}
+
+size_t RocksDBSinkWriter::HandleConflicts(primary_key::Keys& keys) {
```
MBkkt:
```suggestion
size_t RocksDBSinkWriter::HandleConflicts(const primary_key::Keys& keys) {
```

server/connector/rocksdb_sink_writer.cpp
```diff
@@ -24,6 +24,55 @@
...
+
+    const bool conflict = status.ok();
+
+    if (conflict) {
+      switch (_conflict_policy) {
+        case WriteConflictPolicy::Replace:
+          SDB_ASSERT(false,
+                     "WriteConflictPolicy::Update should be handled earlier "
+                     "for optimiztion reason");
+          break;
```
MBkkt:
```suggestion
```
this code is obviously unreachable, also SDB_UNREACHABLE have assert(false)

server/connector/data_sink.cpp
Dronplane:
```suggestion
  WriteConflictPolicy conflict_policy, size_t& number_of_rows_affected,
```
If we assert pointer is not null - why not reference?

server/connector/rocksdb_sink_writer.cpp
```diff
@@ -24,6 +24,55 @@
...
+    // See issue #184
+    _lookup_value.Reset();
+
+    if (!status.ok() && !status.IsNotFound()) {
+      SDB_THROW(rocksutils::ConvertStatus(status));
+    }
+
+    const bool conflict = status.ok();
+
+    if (conflict) {
```
MBkkt: maybe
```suggestion
    if (status.ok()) {
       ...
    } else if (!status.IsNotFound()) {
       SDB_THROW(rocksutils::ConvertStatus(status));
    }
```

tests/server/connector/data_source_test.cpp
```diff
@@ -85,9 +85,11 @@ class DataSourceTest : public ::testing::Test,
     std::unique_ptr<rocksdb::Transaction> transaction{
       _db->BeginTransaction(wo, trx_opts, nullptr)};
     ASSERT_NE(transaction, nullptr);
+    size_t rows_affected = 0;
     sdb::connector::RocksDBInsertDataSink sink(
       *transaction, *_cf_handles.front(), *pool_.get(), object_key, pk,
-      column_ids, {});
+      std::move(column_ids), sdb::WriteConflictPolicy::Replace, &rows_affected,
+      {});
```
Dronplane: NIT we can add assertion check for rows_affected here too

server/connector/rocksdb_sink_writer.cpp
```diff
@@ -24,6 +24,55 @@
...
+                     "for optimiztion reason");
+          break;
+        case WriteConflictPolicy::DoNothing:
+          // Mark key: it should be skipped
+          key.clear();
+          skipped_cnt++;
+          break;
+        case WriteConflictPolicy::EmitError:
+          SDB_THROW(ERROR_SERVER_UNIQUE_CONSTRAINT_VIOLATED,
+                    "Primary key already exists");
```
MBkkt: we probably want postgres like error

tests/server/connector/sink_with_search_test.cpp
```diff
@@ -252,10 +254,12 @@ class DataSinkWithSearchTest : public ::testing::Test,
     index_writers.emplace_back(
       std::make_unique<sdb::connector::search::SearchSinkUpdateWriter>(
         index_transaction));
+    size_t rows_affected = 0;
+
     sdb::connector::RocksDBUpdateDataSink sink(
       *data_transaction, *_cf_handles.front(), *pool_.get(), object_key, pk,
       data_column_oids, all_column_oids, update_pk, table_row_type,
-      std::move(index_writers));
+      &rows_affected, std::move(index_writers));
```
Dronplane: NIT we can add assertion check for rows_affected here too

server/connector/rocksdb_sink_writer.hpp
```diff
@@ -22,40 +22,60 @@
...
+  RocksDBSinkWriterBase(
+    rocksdb::Transaction& transaction, rocksdb::ColumnFamilyHandle& cf,
+    WriteConflictPolicy conflict_policy = WriteConflictPolicy::EmitError)
+    : _transaction{transaction}, _cf{cf}, _conflict_policy{conflict_policy} {}
 
   virtual ~RocksDBSinkWriterBase() = default;
 
   rocksdb::Status Lock(std::string_view full_key) {
-    return _transaction.GetKeyLock(&_cf, full_key, false, true);
+    const bool reentrant = _conflict_policy == WriteConflictPolicy::DoNothing;
```
MBkkt: I think it should be, because for replace also ok to lock same key, can you also add test for this?
```suggestion
    const bool reentrant = _conflict_policy != WriteConflictPolicy::EmitError;
```

server/connector/serenedb_connector.hpp
```diff
@@ -329,25 +338,16 @@ class SereneDBConnectorInsertTableHandle final
...
-      _transaction.EnsureRocksDBTransaction().GetNumKeys();
-    if (_update_pk) {
-      // Each affected rows has associated removed key and inserted one.
-      // Update of PK is implemented as delete with old key + insert with new
-      // key
-      return keys_affected / 2;
-    }
-    return keys_affected;
-  }
+  size_t* NumberOfRowsAffected() const noexcept { return &_rows_affected; }
```
```suggestion
  size_t& NumberOfRowsAffected() const noexcept { return _rows_affected; }
```

server/connector/serenedb_connector.hpp
```diff
@@ -329,25 +338,16 @@ class SereneDBConnectorInsertTableHandle final
...
 
  private:
   axiom::connector::ConnectorSessionPtr _session;
   axiom::connector::TablePtr _table;
   axiom::connector::WriteKind _kind;
   query::Transaction& _transaction;
   std::vector<velox::connector::ColumnHandlePtr> _row_id_handles;
-  bool _update_pk{};
+  mutable size_t _rows_affected = 0;
```
```suggestion
  mutable uint64_t _rows_affected = 0;
```

server/pg/pg_comm_task.cpp
```diff
@@ -178,6 +187,12 @@ void PgSQLCommTaskBase::SafeCall(Func&& func) noexcept try {
   }
 } catch (const SqlException& e) {
   SendNotice(PQ_MSG_ERROR_RESPONSE, e.error());
+} catch (const basics::Exception& e) {
+  if (auto as_sql_e = TryAsSqlError(e)) {
+    SendNotice(PQ_MSG_ERROR_RESPONSE, as_sql_e->error());
+  } else {
+    SendError(e.what(), ERRCODE_INTERNAL_ERROR);
+  }
```
MBkkt: remove this please, we want initially throw postgres error

server/pg/sql_analyzer_velox.cpp
```diff
@@ -1581,9 +1581,30 @@ void SqlAnalyzer::ProcessInsertStmt(State& state, const InsertStmt& stmt) {
...
+                "ON CONFLICT with target list is not implemented yet");
+    }
+    if (stmt.onConflictClause->infer != nullptr) {
+      SDB_THROW(ERROR_NOT_IMPLEMENTED,
+                "ON CONFLICT with infer clause is not implemented yet");
+    }
+    if (stmt.onConflictClause->whereClause != nullptr) {
+      SDB_THROW(ERROR_NOT_IMPLEMENTED,
+                "ON CONFLICT WHERE clause is not implemented yet");
+    }
```
MBkkt: does this possible for not update?

server/pg/sql_exception.h
```diff
@@ -47,3 +47,6 @@ class SqlException final : public std::exception {
 
 #define THROW_SQL_ERROR_FROM_DATA(sqlData) \
   throw ::sdb::SqlException { (sqlData), std::source_location::current() }
+
+#define CONSTRUCT_SQL_ERROR_FROM_DATA(sqlData) \
+  ::sdb::SqlException { (sqlData), std::source_location::current() }
```
MBkkt: you can remove this

server/pg/sql_exception_macro.h
```diff
@@ -41,3 +41,6 @@
 
 #define THROW_SQL_ERROR(...) \
   THROW_SQL_ERROR_FROM_DATA(SQL_ERROR_DATA(__VA_ARGS__))
+
+#define CONSTRUCT_SQL_ERROR(...) \
```
MBkkt: you can remove this

server/catalog/types.h
```diff
@@ -65,6 +65,12 @@ enum class IndexSerialization : uint8_t {
   Inventory = 16,
 };
 
+enum class WriteConflictPolicy : uint8_t {
+  Replace,
+  DoNothing,
+  EmitError,
```
MBkkt: it's strange if order in enum is different compared to order in config/etc
```suggestion
  EmitError,
  DoNothing,
  Replace,
```

server/connector/serenedb_connector.hpp
```diff
@@ -254,14 +254,23 @@ class RocksDBTable final : public axiom::connector::Table {
 
   query::Transaction& GetTransaction() const noexcept { return _transaction; }
 
+  WriteConflictPolicy GetConflictPolicy() const noexcept {
+    return _conflict_policy;
+  }
+
+  void SetConflictPolicy(WriteConflictPolicy conflict_policy) {
+    _conflict_policy = conflict_policy;
+  }
```
MBkkt: maybe `GetWriteConflictPolicy()` that return `decltype(auto)` with deducing this?

server/pg/pg_comm_task.cpp
```diff
@@ -149,6 +149,15 @@ constexpr std::array<char, 47> kTimeoutTermination{PQ_MSG_ERROR_RESPONSE,
...
 
+std::optional<sdb::SqlException> TryAsSqlError(const basics::Exception& e) {
+  if (e.code() == ERROR_SERVER_UNIQUE_CONSTRAINT_VIOLATED) {
+    return CONSTRUCT_SQL_ERROR(
+      ERR_CODE(ERRCODE_UNIQUE_VIOLATION),
+      ERR_MSG("duplicate key value violates unique constraint"));
+  }
+  return {};
+}
+
```
MBkkt:
```suggestion
```

server/connector/rocksdb_sink_writer.hpp
```diff
@@ -22,40 +22,60 @@
...
 namespace sdb::connector {
 
 class RocksDBSinkWriterBase {
  public:
-  RocksDBSinkWriterBase(rocksdb::Transaction& transaction,
-                        rocksdb::ColumnFamilyHandle& cf)
-    : _transaction(transaction), _cf(cf) {}
+  RocksDBSinkWriterBase(
+    rocksdb::Transaction& transaction, rocksdb::ColumnFamilyHandle& cf,
+    WriteConflictPolicy conflict_policy = WriteConflictPolicy::EmitError)
```
MBkkt: I don't think we need default value here

server/connector/rocksdb_sink_writer.hpp
```diff
@@ -22,40 +22,60 @@
...
 // directly inside DataSink so no need for virtual calls/default base members
 class RocksDBSinkWriter : public RocksDBSinkWriterBase {
  public:
-  RocksDBSinkWriter(rocksdb::Transaction& transaction,
-                    rocksdb::ColumnFamilyHandle& cf)
-    : RocksDBSinkWriterBase(transaction, cf) {}
-
+  RocksDBSinkWriter(
+    rocksdb::Transaction& transaction, rocksdb::ColumnFamilyHandle& cf,
+    WriteConflictPolicy conflict_policy = WriteConflictPolicy::EmitError)
```
MBkkt: I'm not sure, why do we need default value here?

server/connector/rocksdb_sink_writer.cpp
```diff
@@ -24,6 +24,55 @@
...
+      switch (_conflict_policy) {
+        case WriteConflictPolicy::Replace:
+          SDB_ASSERT(false,
+                     "WriteConflictPolicy::Update should be handled earlier "
+                     "for optimiztion reason");
+          break;
+        case WriteConflictPolicy::DoNothing:
+          // Mark key: it should be skipped
+          key.clear();
+          skipped_cnt++;
```
MBkkt:
```suggestion
          ++skipped_cnt;
```

server/connector/rocksdb_sink_writer.cpp
```diff
@@ -24,6 +24,55 @@
...
+}
+
+size_t RocksDBSinkWriter::HandleConflicts(primary_key::Keys& keys) {
+  if (_conflict_policy == WriteConflictPolicy::Replace) {
+    // Optimize out reading
+    return 0;
+  }
+
+  ConfigureReadOptions();
+  size_t skipped_cnt = 0;
```
MBkkt:
```suggestion
  size_t skipped_count = 0;
```
or just `skipped` if you want short, I think we want to avoid such shortcuts `cnt`/`usr`/`mod`/etc

tests/sqllogic/any/pg/simple/update_pk.test
MBkkt: I'm not sure from code, do we check only locks for "update pk" case? We shouldn't try to read from rocksdb in such case.

UPD: it's probably true because of replace

server/connector/data_sink.hpp
```diff
@@ -171,9 +172,8 @@ class RocksDBDataSinkBase : public velox::connector::DataSink {
   void WriteRowSlices(std::string_view key);
   void WriteNull(std::string_view key);
 
-  const std::string& SetupRowKey(
-    velox::vector_size_t idx,
-    std::span<const velox::vector_size_t> original_idx);
+  std::string* SetupRowKey(velox::vector_size_t idx,
```
MBkkt:
```suggestion
  const std::string* SetupRowKey(velox::vector_size_t idx,
```

server/connector/serenedb_connector.hpp
MBkkt, line 429: can you add assert for such if's (here and in abort) that `number_of_locked_primary_keys` is zero?

server/connector/rocksdb_sink_writer.cpp
```diff
@@ -24,6 +24,55 @@
...
+  for (auto& key : keys) {
+    auto status = _transaction.Get(_read_options, &_cf, key, &_lookup_value);
```
MBkkt: get from snapshot? we already checked transaction

tests/sqllogic/any/pg/simple/update_pk.test
Dronplane: BTW looks like if PK has more than one column we still will need reading form RocksDB
E.g. 
CREATE TABLE multi_pk2 (a INTEGER, b INTEGER, c VARCHAR, PRIMARY KEY(a, b));
INSERT INTO multi_pk2 VALUES (1,1, 'a'), (1,2, 'b');
UPDATE multi_pk2 SET a = 1, b = 2 WHERE a = 1 AND b = 1;
Here we will have two locks in transaction for keys (1,1) and (1,2). They will not collide and reentry would not help.
But we will overwrite 'b' record in table.

MBkkt: > BTW looks like if PK has more than one column we still will need reading form RocksDB

we don't need read from rocksdb for "conflict detection", update with pk is generally replace.

>  Here we will have two locks in transaction for keys (1,1) and (1,2). They will not collide and reentry would not help.

Yes, but I think @mkornaukhov doesn't support emit_error/do_nothing right now, only replace case for update.

If we need to support emit_error/do_nothing, we need to do something more smart

Dronplane: I guess the problem here it that after running provided script we will have one record in the table. And just loose second ('b') one :(

MBkkt: well it depends on postgres behavior, I think in postgres it depends on isolation level, let's check this

Dronplane: With any isolation level we should not drop existing records not even affected by update.
I really doubt postgres allows such behaviour too.

MBkkt: maybe you're right, I'm not sure

server/connector/data_sink.cpp
mkornaukhov: Fixed

tests/sqllogic/sdb/pg/simple/insert_conflict.test
mkornaukhov: What is the reason?
Fixed

server/connector/rocksdb_sink_writer.cpp
mkornaukhov: Key is cleared in case of `DoNothing` policy and presence of this key in transaction (snapshot or write buffer).
I do it instead of having mask.

server/connector/rocksdb_sink_writer.cpp
mkornaukhov: Same here

server/connector/data_sink.cpp
mkornaukhov: I haven't understood

server/pg/sql_analyzer_velox.cpp
mkornaukhov: Yes, except where clause

server/connector/rocksdb_sink_writer.cpp
mkornaukhov: Fixed in another way

tests/sqllogic/any/pg/simple/update_pk.test
mkornaukhov: Fixed

mkornaukhov: I've decided not to rely on locks as it is not really informative, see last statements `update_pk.test`

server/connector/rocksdb_sink_writer.cpp
mkornaukhov: No, see last statements `update_pk.test`

tests/sqllogic/any/pg/simple/update_pk.test
mkornaukhov: Update should not support such policies

server/connector/serenedb_connector.hpp
```diff
@@ -247,15 +247,19 @@ class RocksDBTable final : public axiom::connector::Table {
...
 
-  bool IsUsedForUpdatePK() const noexcept { return _update_pk; }
-
-  void SetUsedForUpdatePK(bool value = true) { _update_pk = value; }
-
   query::Transaction& GetTransaction() const noexcept { return _transaction; }
 
-  void SetBulkInsert() { _bulk_insert = true; }
+  decltype(auto) WriteConflictPolicy(this auto&& self) noexcept {
+    return (self._write_conflict_policy);
```
MBkkt, line 253: remove ()

server/connector/data_sink.hpp
```diff
@@ -205,25 +229,31 @@ class RocksDBInsertDataSink final
...
  public:
   RocksDBInsertDataSink(
-    rocksdb::Transaction& transaction, rocksdb::ColumnFamilyHandle& cf,
-    velox::memory::MemoryPool& memory_pool, ObjectId object_key,
-    std::span<const velox::column_index_t> key_childs,
-    std::vector<catalog::Column::Id> column_ids,
+    std::string_view table_name, rocksdb::Transaction& transaction,
+    rocksdb::ColumnFamilyHandle& cf, velox::memory::MemoryPool& memory_pool,
+    ObjectId object_key, std::span<const velox::column_index_t> key_childs,
+    std::vector<std::pair<catalog::Column::Id, std::string_view>> columns,
```
MBkkt: struct instead of pair

tests/server/connector/data_source_test.cpp
```diff
@@ -72,22 +72,25 @@ class DataSourceTest : public ::testing::Test,
...
-    std::vector<sdb::catalog::Column::Id> column_ids;
+    std::vector<std::pair<sdb::catalog::Column::Id, std::string_view>>
+      column_ids;
     column_ids.reserve(data->type()->asRow().size());
     for (velox::vector_size_t i = 0; i < data->type()->asRow().size(); ++i) {
       if (!data->childAt(i)->mayHaveNulls()) {
         pk.push_back(i);
       }
-      column_ids.emplace_back(i);
+      column_ids.emplace_back(std::pair{i, ""});
```
MBkkt: you don't need std pair ctor

tests/server/connector/data_sink_test.cpp
```diff
@@ -2604,19 +2618,21 @@ TEST_F(DataSinkTest, test_insertDeleteConflict) {
...
   // should be empty
   ASSERT_TRUE(transaction_delete->Commit().ok());
   ASSERT_TRUE(transaction->Commit().ok());
+  ASSERT_EQ(rows_affected, 0);
   // insert should be fully written
   rocksdb::ReadOptions read_options;
   size_t i = 0;
-  auto column_key =
-    sdb::connector::key_utils::PrepareColumnKey(kObjectKey, column_ids[1]);
+  auto column_key = sdb::connector::key_utils::PrepareColumnKey(
```
MBkkt: why do we need full namespace in tests, let's fix this

server/pg/sql_analyzer_velox.cpp
MBkkt: so we can replace it with assert?

server/connector/data_sink.cpp
```diff
@@ -313,21 +527,44 @@ void RocksDBUpdateDataSink::appendData(velox::RowVectorPtr input) {
...
+      WriteInputColumn(_columns_info[i].first, i, *input, all_rows_range);
     }
+    _number_of_rows_affected += num_rows;
     return;
   }
-
-  // Take locks and prepare buffers for both new and old primary keys
+  // Take locks and prepare buffers for both new and
+  // old primary keys
+  containers::FlatHashSet<std::string_view> batch_keys;
```
MBkkt: I doubt we want map of batch keys, also it can be cached to avoid alloc/free in loop

server/connector/data_sink.cpp
```diff
@@ -43,66 +43,253 @@
...
+  SDB_ASSERT(keys.size() == old_keys.size() || old_keys.empty());
+
+  const auto conflict_policy = _write_conflict_policy;
+  if (conflict_policy == WriteConflictPolicy::Replace) {
+    // Optimize out reading
+    return 0;
+  }
+
+  _read_options.async_io = true;
+  _read_options.snapshot = _transaction.GetSnapshot();
```
MBkkt: it doesn't look good that we create it every time

server/connector/serenedb_connector.hpp
mkornaukhov: No, it's required to use () to return reference, otherwise it will be deduced as value, not ref

server/connector/data_sink.cpp
mkornaukhov: We want to check that there are no collisions in batch before writing into transaction

server/connector/data_sink.cpp
```diff
@@ -43,66 +43,251 @@
...
 
 constexpr std::string_view kZeroLengthVector{"\0", 1};
 constexpr std::string_view kOneValueHeader{"\0\1", 2};
 
+template<typename T>
+std::string VeloxValueToString(T val) {
+  if constexpr (std::is_same_v<T, velox::StringView>) {
+    return std::string(val.data(), val.size());
+  } else {
+    return std::to_string(val);
```
MBkkt: StrCat

server/connector/data_sink.cpp
```diff
@@ -43,66 +43,251 @@
...
+std::string GetPKVeloxValue(velox::VectorPtr vec, velox::vector_size_t idx) {
+  SDB_ASSERT(!vec->isNullAt(idx));
+
+  switch (vec->encoding()) {
+    case facebook::velox::VectorEncoding::Simple::CONSTANT:
+      return GetPKVeloxConstantValue<Kind>(vec, idx);
+    case facebook::velox::VectorEncoding::Simple::DICTIONARY:
+      return GetPKVeloxDictValue<Kind>(vec, idx);
+    case facebook::velox::VectorEncoding::Simple::FLAT:
+      return GetPKVeloxFlatValue<Kind>(vec, idx);
```
MBkkt: `SimpleVector<T>::valueAt`

server/connector/data_sink.cpp
```diff
@@ -192,9 +380,10 @@ void RocksDBInsertDataSink::appendData(velox::RowVectorPtr input) {
   const auto num_columns = input->childrenSize();
 
   _store_keys_buffers.clear();
-  _store_keys_buffers.reserve(num_rows);
+  _store_keys_buffers.resize(num_rows);
 
   for (size_t row_idx = 0; row_idx < num_rows; ++row_idx) {
+    auto& key_buffer = _store_keys_buffers[row_idx];
```
MBkkt: maybe emplace_back

server/connector/data_sink.cpp
```diff
@@ -276,17 +486,19 @@ void RocksDBUpdateDataSink::appendData(velox::RowVectorPtr input) {
   const velox::IndexRange all_rows(0, num_rows);
   const folly::Range all_rows_range{&all_rows, 1};
 
-  auto ensure_input_sorted = [](const auto& keys) {
+  auto extract_pk = [](std::string_view key) {
+    return key.substr(sizeof(ObjectId) + sizeof(catalog::Column::Id));
+  };
```
MBkkt: move closer to usage, maybe constant for sizeof + sizeof, instead of lambda function

server/connector/data_sink.cpp
```diff
@@ -313,21 +525,45 @@ void RocksDBUpdateDataSink::appendData(velox::RowVectorPtr input) {
...
+  _batch_keys.reserve(num_rows);
   for (size_t row_idx = 0; row_idx < num_rows; ++row_idx) {
-    key_utils::MakeColumnKey(input, _updated_key_childs, row_idx, table_key,
-                             lock_new_row, _store_keys_buffers.emplace_back());
+    key_utils::MakeColumnKey(
+      input, _updated_key_childs, row_idx, table_key,
+      [&](std::string_view lock_key) {
+        if (auto [_, inserted] =
+              _batch_keys.insert(lock_key.substr(sizeof(ObjectId)));
+            !inserted) {
```
MBkkt:
```suggestion
        if (!_batch_keys.emplace(lock_key.substr(sizeof(ObjectId))).second) {
```

server/connector/serenedb_connector.hpp
```diff
@@ -646,37 +645,42 @@ class SereneDBConnector final : public velox::connector::Connector {
...
+            if (table.BulkInsert()) {
               return std::make_unique<SSTInsertDataSink>(
                 _db, _cf, *connector_query_ctx->memoryPool(), object_key,
-                pk_indices, column_oids, _rocksdb_directory);
+                pk_indices, columns, _rocksdb_directory);
             }
 
             return std::make_unique<RocksDBInsertDataSink>(
+              table.name(),
+
```
```suggestion
```

server/connector/serenedb_connector.hpp
```diff
@@ -646,37 +645,42 @@ class SereneDBConnector final : public velox::connector::Connector {
...
-      column_oids.reserve(table.type()->size());
+      columns.reserve(table.type()->size());
       for (auto& col : table.type()->names()) {
         auto handle = table.columnMap().find(col);
         SDB_ASSERT(handle != table.columnMap().end(),
                    "RocksDBDataSink: can't find column handle for ", col);
-        column_oids.push_back(
-          basics::downCast<const SereneDBColumn>(handle->second)->Id());
+        auto column = basics::downCast<const SereneDBColumn>(handle->second);
+        columns.push_back({column->Id(), column->name()});
```
MBkkt: emplace_back

server/connector/serenedb_connector.hpp
```diff
@@ -577,17 +576,17 @@ class SereneDBConnector final : public velox::connector::Connector {
...
+      columns.reserve(input_type->size());
       for (auto& col : input_type->names()) {
         std::string_view real_name = catalog::Column::ExtractColumnName(col);
         auto handle = table.columnMap().find(real_name);
         SDB_ASSERT(handle != table.columnMap().end(),
                    "RocksDBDataSink: can't find column handle for ", real_name);
-        column_oids.push_back(
-          basics::downCast<const SereneDBColumn>(handle->second)->Id());
+        auto column = basics::downCast<const SereneDBColumn>(handle->second);
+        columns.push_back({.id = column->Id(), .name = column->name()});
```
MBkkt: emplace_back

server/pg/sql_analyzer_velox.cpp
```diff
@@ -1603,9 +1603,30 @@ void SqlAnalyzer::ProcessInsertStmt(State& state, const InsertStmt& stmt) {
...
+  auto conflict_policy = config.Get<VariableType::SdbWriteConflictPolicy>(
+    "sdb_write_conflict_policy");
   if (stmt.onConflictClause) {
-    SDB_THROW(ERROR_NOT_IMPLEMENTED,
-              "ON CONFLICT clause is not implemented yet");
+    if (stmt.onConflictClause->action == ONCONFLICT_UPDATE) {
+      SDB_THROW(ERROR_NOT_IMPLEMENTED,
+                "ON CONFLICT DO UPDATE SET action is not implemented yet");
+    }
+    if (stmt.onConflictClause->targetList != nullptr) {
```
```suggestion
    if (stmt.onConflictClause->targetList) {
```
codestyle

tests/server/connector/data_sink_test.cpp
```diff
@@ -28,11 +28,12 @@
 #include "iresearch/utils/bytes_utils.hpp"
 #include "rocksdb/utilities/transaction_db.h"
 
+using namespace sdb;
 using namespace sdb::connector;
```
MBkkt: namespace sdb::connector {

tests/server/connector/data_source_test.cpp
```diff
@@ -70,33 +71,34 @@ class DataSourceTest : public ::testing::Test,
     std::filesystem::remove_all(_path);
   }
 
-  void MakeRocksDBWrite(velox::RowVectorPtr data, sdb::ObjectId object_key) {
+  void MakeRocksDBWrite(velox::RowVectorPtr data, ObjectId object_key) {
     std::vector<velox::column_index_t> pk;
-    std::vector<sdb::catalog::Column::Id> column_ids;
+    std::vector<::ColumnInfo> column_ids;
```
MBkkt: why global namespace in this file

server/connector/data_sink.cpp
mkornaukhov: It does not support single argument

server/connector/serenedb_connector.hpp
```diff
@@ -577,17 +576,17 @@ class SereneDBConnector final : public velox::connector::Connector {
...
-      column_oids.reserve(input_type->size());
+      columns.reserve(input_type->size());
       for (auto& col : input_type->names()) {
         std::string_view real_name = catalog::Column::ExtractColumnName(col);
         auto handle = table.columnMap().find(real_name);
         SDB_ASSERT(handle != table.columnMap().end(),
                    "RocksDBDataSink: can't find column handle for ", real_name);
-        column_oids.push_back(
-          basics::downCast<const SereneDBColumn>(handle->second)->Id());
+        auto column = basics::downCast<const SereneDBColumn>(handle->second);
```
MBkkt: we don't want to copy column
```suggestion
        const auto& column = basics::downCast<const SereneDBColumn>(*handle->second);
```

server/connector/serenedb_connector.hpp
```diff
@@ -646,37 +645,41 @@ class SereneDBConnector final : public velox::connector::Connector {
...
     if (serene_insert_handle.Kind() == axiom::connector::WriteKind::kDelete) {
-      column_oids.reserve(table.type()->size());
+      columns.reserve(table.type()->size());
       for (auto& col : table.type()->names()) {
         auto handle = table.columnMap().find(col);
         SDB_ASSERT(handle != table.columnMap().end(),
                    "RocksDBDataSink: can't find column handle for ", col);
-        column_oids.push_back(
-          basics::downCast<const SereneDBColumn>(handle->second)->Id());
+        auto column = basics::downCast<const SereneDBColumn>(handle->second);
```
MBkkt: same

server/connector/data_sink.cpp
```diff
@@ -43,66 +44,221 @@
...
+    // TODO(mkornaukhov) support not only simple vectors
+    return "<unsupported>";
+  }
+  return VeloxValueToString(simple_vec->valueAt(idx));
+}
+
+std::string FormatKeyValue(const velox::RowVectorPtr& input,
+                           velox::column_index_t key_idx,
+                           velox::vector_size_t row_idx) {
+  auto vector = input->childAt(key_idx);
```
MBkkt: const auto&

server/connector/serenedb_connector.hpp
mkornaukhov: it wont be copied, it's pointer, I'll highlight it

server/connector/serenedb_connector.hpp
mkornaukhov: same

Issue #178, opened 2026-01-23
Move `#include "postgres.h"` to .cpp files


codeworse:
For now there are several files, where libpg is included in header file. This creates problems with compilation.
For instance, in [iresearch function](https://github.com/serenedb/serenedb/blob/main/libs/iresearch/include/iresearch/search/score_function.hpp#L120) there is a method `Max`:
```cpp
score_t Max() const noexcept;
```

In the same time in `libpg_query/src/postgres/include/c.h` there is a macros `Max`:
```cpp
#define Max(x, y)               ((x) > (y) ? (x) : (y))
```
So, in result we are getting compilation error:
```bash
serenedb/libs/iresearch/include/iresearch/search/score_function.hpp:115:25: error: field has incomplete type 'void'
  115 |   IRS_FORCE_INLINE void Min(score_t arg) const noexcept {
      |                         ^
serenedb/libs/iresearch/include/iresearch/search/score_function.hpp:115:28: error: expected ';' at end of declaration list
  115 |   IRS_FORCE_INLINE void Min(score_t arg) const noexcept {
      |                            ^
serenedb/libs/iresearch/include/iresearch/search/score_function.hpp:120:15: error: too few arguments provided to function-like macro invocation
  120 |   score_t Max() const noexcept;
      |               ^
serenedb/third_party/libpg_query/src/postgres/include/c.h:998:9: note: macro 'Max' defined here
  998 | #define Max(x, y)               ((x) > (y) ? (x) : (y))
      |         ^
5 errors generated.
ninja: build stopped: subcommand failed.
```

The best approach would be to erase all includes from headers. However, it's hard to rewrite `pg_list_utils.h`

PR #180, opened 2026-01-23
Move Min/Max to libpg macro

codeworse:
Move `Min/Max` to libpg macro in order to avoid compilation error(e.g., macros `Min/Max` in `c.h` in postgres and `Min/Max` methods in scorer in iresearch). Also, partially remove unnecessary includes of postgres.h

Related issue: https://github.com/serenedb/serenedb/issues/178


libs/iresearch/include/iresearch/search/score_function.hpp
```diff
@@ -112,12 +112,12 @@ class ScoreFunction : util::Noncopyable {
...
   }
 
-  IRS_FORCE_INLINE void Min(score_t arg) const noexcept {
+  IRS_FORCE_INLINE void(Min)(score_t arg) const noexcept {
     SDB_ASSERT(_min != nullptr);
     _min(_ctx, arg);
   }
 
-  score_t Max() const noexcept;
+  score_t(Max)() const noexcept;
```
MBkkt: revert this please

codeworse: Done

server/pg/functions/json.h
```diff
@@ -337,9 +332,8 @@ struct PgJsonInFunction {
     simdjson::ondemand::document doc;
     auto ec = parser.iterate(padded_input).get(doc);
     if (ec != simdjson::SUCCESS || !ValidateJson(doc)) {
-      THROW_SQL_ERROR(
-        ERR_CODE(ERRCODE_INVALID_JSON_TEXT),
-        ERR_MSG("Invalid input syntax for type json: ", input_view));
+      SDB_THROW(ERROR_BAD_PARAMETER,
```
Dronplane: Wouldn't it break proper error reporting if we use here not-pg error?
Maybe I'm wrong but we still can move function to cpp but keep it as force inlined

codeworse: I guess `functions/json.*` can be stayed untouched, since it is included only into `funcitons.cpp`. So, no compilation error should not occur

PR #182, opened 2026-01-24
Remove single doc iterator

gnusi:
Remove singleton iterator specialization.


libs/iresearch/include/iresearch/formats/formats_impl.hpp
```diff
@@ -3651,11 +3519,6 @@ DocIterator::ptr PostingsReaderImpl<FormatTraits>::Iterator(
     });
   };
 
-  if (std::popcount(iterator_types) != 1 || (iterator_types & (1 << 0))) {
```
MBkkt, line 3654: this cannot be removed because of wanderator

gnusi: Yep, already noticed it.

PR #209, opened 2026-02-05
Support read your own writes

mkornaukhov:
Add 'read your own writes' support and data source improvements.
- Save N RocksDB iterators (where N is the number of columns), which required disabling async IO.
- Use `adaptive_readahead` rocksdb read option to improve reading latency.
- Leverage `iterate_upper_bound` option for the same purpose.
- Get rid of extra code .
- Add support for read-your-own-writes (guaranteed on every isolation level in Postgres): read from the transaction, not just the snapshot. This makes throughput worse by 35%, so it can be disabled with the `sdb_read_your_own_writes` option. We should investigate how to improve performance here.

Current changes improved latency by 12% on queries like `SELECT count (a + b + c) from t1;`, table of 3 int columns with 10kk rows. 

Also I've tried some perspective rocksdb options which turned out to be not very useful
- Setting snapshot to `nullptr`. Improves latency for about 1%, maybe useful for read committed isolation level, but performance gain is expected to be higher.
- Tailing iterator. It also may be useful for read committed isolation level, but it make performance worse, about 5%.
- `total_order_seek, auto_prefix_mode, prefix_same_as_start` and `prefix_seek_opt_in_only` for db option. Looks like they are more appropriate for point look ups. I've tried some combinations, the best one is to enable only `auto_prefix_mode`, it gives about 1% for latency.
- Disabling verify checksums: improves at less than 1%, so better to keep it enabled.

### DIFF
--- a/server/query/transaction.cpp
+++ b/server/query/transaction.cpp
@@ -150,6 +150,7 @@ void Transaction::ApplyTableStatsDiffs() {
       table_shard->UpdateNumRows(delta);
     }
   }
+  _table_rows_deltas.clear();
 }
 
 }  // namespace sdb::query


server/query/transaction.cpp
```diff
@@ -84,7 +84,6 @@ const rocksdb::Snapshot& Transaction::EnsureRocksDBSnapshot() {
 }
 
 rocksdb::Transaction& Transaction::EnsureRocksDBTransaction() {
-  SDB_ASSERT((_state & State::HasRocksDBWrite) != State::None);
```
Dronplane, line 96: Maybe just update assert to something like Write OR RYOW?

tests/sqllogic/any/pg/simple/insert_conflict.test
```diff
@@ -69,12 +69,11 @@ BEGIN;
 statement count 1
 INSERT INTO t2 VALUES (1, 10);
 
-# We do not support 'Read Your Own Writes' yet
-onlyif serenedb
 query
 SELECT * FROM t2 ORDER BY a;
 ----
 a	b
+1	10
```
Dronplane, line 76: Should we also add test with insert(pk1) - delete(pk1) - insert(pk1) - update(pk1 -> pk2) - insert(pk1)  sequence to check that all our inter-transaction machinery works

tests/sqllogic/sdb/pg/simple/read_your_own_writes.test
```diff
@@ -0,0 +1,180 @@
...
+# Should NOT see the deleted row
+query
+SELECT * FROM t_ryow WHERE id = 3;
+----
+id	value
+
+statement ok
+ROLLBACK;
+
+# Test DELETE with read_your_own_writes=off
```
Dronplane, line 129: Should we also have a dedicated test for DELETE not yet committed INSERT with read-own-writes off  - this delete should not delete anything. And same with RYOW=on, but there DELETE should be able to delete not yet committed INSERT. And after COMMIT no data should be added to the table

server/connector/data_source.cpp
```diff
@@ -60,6 +58,83 @@ RocksDBDataSource::RocksDBDataSource(
...
+RocksDBRYOWDataSource::RocksDBRYOWDataSource(
+  velox::memory::MemoryPool& memory_pool, rocksdb::Transaction& transaction,
+  rocksdb::ColumnFamilyHandle& cf, velox::RowTypePtr row_type,
+  std::vector<catalog::Column::Id> column_ids,
+  catalog::Column::Id effective_column_id, ObjectId object_key)
+  : RocksDBDataSource(memory_pool, cf, std::move(row_type),
+                      std::move(column_ids), effective_column_id, object_key,
+                      transaction.GetSnapshot()),
+    _transaction{transaction} {
+  InitIterators([&] {
```
Dronplane: Why do we need here iterators init? I guess we should do this only when we got a split.

server/connector/data_source.cpp
```diff
@@ -60,6 +58,83 @@ RocksDBDataSource::RocksDBDataSource(
...
+  velox::memory::MemoryPool& memory_pool, rocksdb::DB& db,
+  rocksdb::ColumnFamilyHandle& cf, velox::RowTypePtr row_type,
+  std::vector<catalog::Column::Id> column_ids,
+  catalog::Column::Id effective_column_id, ObjectId object_key,
+  const rocksdb::Snapshot* snapshot)
+  : RocksDBDataSource(memory_pool, cf, std::move(row_type),
+                      std::move(column_ids), effective_column_id, object_key,
+                      snapshot),
+    _db{db} {
+  InitIterators([&] {
```
Dronplane: Why do we need here iterators init? I guess we should do this only when we got a split.

server/connector/data_source.cpp
```diff
@@ -60,6 +58,83 @@ RocksDBDataSource::RocksDBDataSource(
...
+                 "RocksDBDataSource: generated PK column is not effective");
+      read_column_id = _effective_column_id;
+    }
+
+    key_utils::AppendColumnKey(key, read_column_id);
+    _column_keys.push_back(key);
+  }
+
+  // Build sorted indices by column_id for sequential RocksDB access
+  _sorted_indices.resize(_row_type->size());
```
Dronplane: Do we really need this? That could make sense if we have single iterator for all columns and seek it on each next. So our seek would be forward only in rocksdb. But if we anyway have one iterator per column it should make no sense.

server/query/transaction.cpp
MBkkt, line 96: This isn't correct

server/connector/data_source.cpp
```diff
@@ -60,6 +58,73 @@ RocksDBDataSource::RocksDBDataSource(
...
+}
+
+RocksDBRYOWDataSource::RocksDBRYOWDataSource(
+  velox::memory::MemoryPool& memory_pool, rocksdb::Transaction& transaction,
+  rocksdb::ColumnFamilyHandle& cf, velox::RowTypePtr row_type,
+  std::vector<catalog::Column::Id> column_ids,
+  catalog::Column::Id effective_column_id, ObjectId object_key)
+  : RocksDBDataSource(memory_pool, cf, std::move(row_type),
+                      std::move(column_ids), effective_column_id, object_key,
+                      transaction.GetSnapshot()),
```
codeworse:
```suggestion
  : RocksDBDataSource{
      memory_pool, cf,
      std::move(row_type),
      std::move(column_ids),
      effective_column_id, object_key,
      transaction.GetSnapshot(),
      },
```

server/connector/data_source.cpp
```diff
@@ -60,6 +58,73 @@ RocksDBDataSource::RocksDBDataSource(
...
+
+RocksDBSnapshotDataSource::RocksDBSnapshotDataSource(
+  velox::memory::MemoryPool& memory_pool, rocksdb::DB& db,
+  rocksdb::ColumnFamilyHandle& cf, velox::RowTypePtr row_type,
+  std::vector<catalog::Column::Id> column_ids,
+  catalog::Column::Id effective_column_id, ObjectId object_key,
+  const rocksdb::Snapshot* snapshot)
+  : RocksDBDataSource(memory_pool, cf, std::move(row_type),
+                      std::move(column_ids), effective_column_id, object_key,
+                      snapshot),
```
codeworse:
```suggestion
  : RocksDBDataSource{memory_pool, cf, std::move(row_type),
                      std::move(column_ids), effective_column_id, object_key,
                      snapshot},
```

server/connector/data_source.cpp
```diff
@@ -70,83 +135,54 @@ void RocksDBDataSource::addSplit(
...
-    SDB_ASSERT(
-      it->key() == column_key + _last_read_key,
-      "RocksDBDataSource: inconsistent snapshot. Last read key mismatch");
-    it->Next();
-    rocksutils::CheckIteratorStatus(*it);
-  }
-
-  if (!it->Valid() || !it->key().starts_with(column_key)) {
-    it.reset();
+template<typename CreateFn>
```
codeworse: `std::invocable` concept?

server/connector/data_source.cpp
```diff
@@ -301,15 +338,11 @@ template<typename Callback>
 uint64_t RocksDBDataSource::IterateColumn(rocksdb::Iterator& it,
                                           uint64_t max_size,
                                           std::string_view column_key,
-                                          const Callback& func,
-                                          std::string* last_key) {
+                                          const Callback& func) {
```
codeworse, line 341: maybe concept too?

server/query/transaction.cpp
```diff
@@ -67,7 +67,7 @@ bool Transaction::HasTransactionBegin() const noexcept {
 }
 
 rocksdb::Transaction* Transaction::GetRocksDBTransaction() const noexcept {
-  SDB_ASSERT((_state & State::HasRocksDBWrite) != State::None);
+  // SDB_ASSERT((_state & State::HasRocksDBWrite) != State::None);
```
codeworse: Why into the comments?

mkornaukhov: Removed at all.
It's nullable, so it's OK to return nullptr for non-write transaction

server/connector/data_source.cpp
mkornaukhov: Fixed

tests/sqllogic/any/pg/simple/insert_conflict.test
mkornaukhov: Added

tests/sqllogic/sdb/pg/simple/read_your_own_writes.test
mkornaukhov: Added

server/query/transaction.cpp
mkornaukhov: Kept the line

server/connector/data_source.cpp
mkornaukhov: Yes, fixed

server/connector/data_source.cpp
mkornaukhov: Yes, fixed

server/connector/data_source.cpp
mkornaukhov: Yes, I've measured, no point for that. It's removed

server/query/transaction.cpp
mkornaukhov: Kept

server/connector/data_source.cpp
mkornaukhov: Fixed

server/connector/data_source.cpp
mkornaukhov: Fixed

server/connector/data_source.cpp
mkornaukhov: Fixed

server/query/transaction.cpp
```diff
@@ -150,7 +150,6 @@ void Transaction::ApplyTableStatsDiffs() {
       table_shard->UpdateNumRows(delta);
     }
   }
-  _table_rows_deltas.clear();
```
MBkkt, line 153: Why is it removed?

server/query/config.h
```diff
@@ -54,6 +54,7 @@ enum class VariableType {
   PgExtraFloatDigits,
   PgByteaOutput,
   SdbWriteConflictPolicy,
+  SdbReadYourOwnWrites,
```
MBkkt: I don't think we want separate type for this, can we just make it `Bool` type?

server/query/config.h
```diff
@@ -139,6 +140,14 @@ class Config : public velox::config::IConfig {
...
                  "sdb_write_conflict_policy is not validated");
       return WriteConflictPolicy::Replace;
+    } else if constexpr (T == VariableType::SdbReadYourOwnWrites) {
+      SDB_ASSERT(key == "sdb_read_your_own_writes");
+      if (absl::EqualsIgnoreCase("on", *value_str)) {
+        return true;
+      }
+      SDB_ASSERT(absl::EqualsIgnoreCase("off", *value_str),
+                 "sdb_read_your_own_writes is not validated");
+      return false;
```
MBkkt: Instead of this needs to implement for T == Bool `absl::SimpleAtob`

mkornaukhov: fixed

server/query/config.h
mkornaukhov: fixed

PR #211, opened 2026-02-06
perf: Remove keys copying in SSTWriters

pashandor789:


server/pg/sql_analyzer_velox.cpp
MBkkt, line 2041: what will be in case of format = 1

server/pg/sql_analyzer_velox.cpp
```diff
@@ -2006,6 +2006,29 @@ class CopyOptionsParser {
...
+    if (_file_path.empty()) {
+      return {};
+    }
+
+    auto dot_pos = _file_path.rfind('.');
+    if (dot_pos == std::string_view::npos || dot_pos + 1 >= _file_path.size()) {
+      return {};
+    }
+
+    auto file_format = _file_path.substr(dot_pos + 1);
```
```
const auto pos = _file_path.rfind('.');
if (npos) { return {}; }
const auto file_format = _file_path.substr(dot_pos + 1);
    if (file_format == "csv" || file_format == "text" | file_format == "parquet" ||
        file_format == "dwrf" || file_format == "orc") {
      return file_format;
    }
    if (file_format == "tsv" || file_format == "txt") {
      return "text";
    }
```

server/connector/key_utils.hpp
```diff
@@ -62,6 +63,16 @@ void MakeColumnKey(const velox::RowVectorPtr& input,
...
   std::memcpy(key_buffer.data(), object_id.data(), sizeof(ObjectId));
+
+  if constexpr (IsInternalSSTKey) {
+    // TODO: if key is fixed size, we can avoid extra alloc here
+    const auto key_size = key_buffer.size();
+    basics::StrAppend(key_buffer,
+                      sizeof(rocksdb::SstFileWriter::kInternalKeyFooter));
+    std::memcpy(key_buffer.data() + key_size,
+                &rocksdb::SstFileWriter::kInternalKeyFooter,
+                sizeof(rocksdb::SstFileWriter::kInternalKeyFooter));
```
MBkkt: just append

PR #216, opened 2026-02-06
Improve block disjunction

MBkkt:


libs/iresearch/include/iresearch/search/disjunction.hpp
```diff
@@ -905,27 +907,26 @@ class BlockDisjunction : public DocIterator, private Merger, private ScoreCtx {
...
-          _buf_offset += BitsRequired<uint64_t>();
-        }
       }
 
-      const size_t offset = std::countr_zero(_cur);
-      irs::UnsetBit(_cur, offset);
+      // https://lemire.me/blog/2018/02/21/iterating-over-set-bits-quickly
+      const auto t = _cur & -_cur;
+      const auto offset = std::countr_zero(_cur);
+      _cur ^= t;
```
gnusi, line 927: This should work faster:
```
const auto offset = std::countr_zero(_cur);
_cur &= _cur - 1;
```

MBkkt: it's not

PR #223, opened 2026-02-10
Search DataSource 1

Dronplane:
PR introduces DataSource for materializing documents found by iresearch filter and Filter building utilities to convert velox filter expression to the iresearch filter.
PR changes handling BETWEEN SYMMETRIC so it matches handling of BETWEEN.

Supported: =, <=>, NOT, LIKE, BETWEEN, IN, IS NULL operations.

Next steps:
1. Move all materialization code from DataSources/Sinks to data materializer once all current improvements are merged
2. Support field cast cases (currently if there is no direct table field access filter building fails)
3. Support failures of materialization for some rows.
4. Keep rejected filters in data source for early pruning?
5.  wire everything up in the connector so it could create search sources when necessary

### DIFF
--- a/server/connector/search_data_source.cpp
+++ b/server/connector/search_data_source.cpp
@@ -36,10 +36,17 @@ SearchDataSource::SearchDataSource(
   std::vector<catalog::Column::Id> column_ids,
   catalog::Column::Id effective_column_id, ObjectId object_key,
   irs::IndexReader& reader, const irs::Filter::Query& query)
-  : Materializer(memory_pool, snapshot, &db, nullptr, cf, row_type,
-                 std::move(column_ids), effective_column_id, object_key),
-    _reader(reader),
-    _query(query) {}
+  : Materializer{memory_pool,
+                 snapshot,
+                 &db,
+                 nullptr,
+                 cf,
+                 row_type,
+                 std::move(column_ids),
+                 effective_column_id,
+                 object_key},
+    _reader{reader},
+    _query{query} {}
 
 void SearchDataSource::addSplit(
   std::shared_ptr<velox::connector::ConnectorSplit> split) {
--- a/server/connector/search_data_source.hpp
+++ b/server/connector/search_data_source.hpp
@@ -21,6 +21,7 @@
 #pragma once
 #include <velox/connectors/Connector.h>
 
+#include <iresearch/analysis/token_attributes.hpp>
 #include <iresearch/search/filter.hpp>
 
 #include "basics/fwd.h"
@@ -41,8 +42,6 @@ class SearchDataSource final : public velox::connector::DataSource,
                    std::vector<catalog::Column::Id> column_ids,
                    catalog::Column::Id effective_column_id, ObjectId object_key,
                    irs::IndexReader& reader, const irs::Filter::Query& query);
-
-  virtual ~SearchDataSource() = default;
 
   void addSplit(std::shared_ptr<velox::connector::ConnectorSplit> split) final;
   std::optional<velox::RowVectorPtr> next(uint64_t size,
--- a/server/connector/search_filter_builder.cpp
+++ b/server/connector/search_filter_builder.cpp
@@ -19,10 +19,13 @@
 ////////////////////////////////////////////////////////////////////////////////
 
 #include "search_filter_builder.hpp"
+
 // NOLINTBEGIN
-// Need this header to stay first to avoid conflict in DCHECK macros
+// Need this header to stay before iresearch/utils/wildcard_utils.hpp to avoid
+// conflict in DCHECK macros see issue #230
 #include "serenedb_connector.hpp"
 // NOLINTEND
+
 #include <velox/expression/ExprConstants.h>
 
 #include <iresearch/analysis/tokenizers.hpp>
@@ -83,7 +86,8 @@ std::optional<velox::Variant> EvaluateConstant(
   if (!expr->isConstantKind()) {
     return std::nullopt;
   }
-  auto const_expr = basics::downCast<velox::core::ConstantTypedExpr>(*expr);
+  const auto& const_expr =
+    basics::downCast<velox::core::ConstantTypedExpr>(*expr);
   if (const_expr.hasValueVector()) {
     SDB_ASSERT(const_expr.valueVector()->size() == 1);
     return const_expr.valueVector()->variantAt(0);
@@ -240,17 +244,18 @@ ComparisonOp InvertComparisonOp(ComparisonOp op) {
 
 template<typename Filter, typename Source>
 Result MakeGroup(Source& parent, const VeloxFilterContext& ctx,
-                 const velox::core::CallTypedExpr* call) {
+                 const velox::core::CallTypedExpr& call) {
   auto sub_ctx = ctx;
   sub_ctx.boost = irs::kNoBoost;
   irs::BooleanFilter* group_root;
-  if (ctx.negated && absl::c_all_of(call->inputs(), [](const auto& input) {
-        auto* call =
-          dynamic_cast<const velox::core::CallTypedExpr*>(input.get());
-        if (!call) {
+  if (ctx.negated && absl::c_all_of(call.inputs(), [](const auto& input) {
+        SDB_ASSERT(input);
+        if (!input->isCallKind()) {
           return false;
         }
-        return GetComparisonOp(call->name()) != ComparisonOp::None;
+        const auto& call =
+          basics::downCast<const velox::core::CallTypedExpr>(*input);
+        return GetComparisonOp(call.name()) != ComparisonOp::None;
       })) {
     // DeMorgan`s law could be used if we negate group of comparisons. As
     // comparisons consume negation by invertion we can reduce number of NOT
@@ -267,7 +272,7 @@ Result MakeGroup(Source& parent, const VeloxFilterContext& ctx,
     sub_ctx.negated = false;
   }
   group_root->boost(ctx.boost);
-  for (const auto& input : call->inputs()) {
+  for (const auto& input : call.inputs()) {
     auto result = FromVeloxExpression(*group_root, sub_ctx, input);
     if (!result.ok()) {
       return result;
@@ -280,7 +285,6 @@ bool IsNotExpr(const std::string& name) {
   return name == "not" || name.ends_with("_not");
 }
 
-// Helper: Check if it's an equality/inequality operator
 bool IsEqualityOp(const std::string& name, bool& not_equal) {
   if (name == "eq" || name.ends_with("_eq")) {
     not_equal = false;
@@ -315,19 +319,19 @@ bool IsLike(std::string_view name) {
 
 Result FromVeloxBinaryEq(irs::BooleanFilter& filter,
                          const VeloxFilterContext& ctx,
-                         const velox::core::CallTypedExpr* call,
+                         const velox::core::CallTypedExpr& call,
                          bool not_equal) {
-  if (call->inputs().size() != 2) {
-    return {ERROR_NOT_IMPLEMENTED, "Equality has ", call->inputs().size(),
+  if (call.inputs().size() != 2) {
+    return {ERROR_NOT_IMPLEMENTED, "Equality has ", call.inputs().size(),
             " inputs but 2 expected"};
   }
-  if (!call->inputs()[0]->isFieldAccessKind()) {
+  if (!call.inputs()[0]->isFieldAccessKind()) {
     return {ERROR_BAD_PARAMETER, "Left input is not field access"};
   }
 
   auto* left_field = static_cast<const velox::core::FieldAccessTypedExpr*>(
-    call->inputs()[0].get());
-  auto value = EvaluateConstant(call->inputs()[1]);
+    call.inputs()[0].get());
+  auto value = EvaluateConstant(call.inputs()[1]);
 
   if (!value.has_value()) {
     return {ERROR_BAD_PARAMETER, "Failed to evaluate right value as constant"};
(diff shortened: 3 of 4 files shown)


server/connector/search_filter_builder.cpp
```diff
@@ -0,0 +1,687 @@
...
+}
+
+template<typename Filter, typename Source>
+Filter& Negate(Source& parent) {
+  return AddFilter<Filter>(AddFilter<irs::Not>(
+    parent.type() == irs::Type<irs::Or>::id() ? AddFilter<irs::And>(parent)
+                                              : parent));
+}
+
+enum class ComparisonOp { KNone, KLt, KLe, KGt, KGe };
```
MBkkt:
```suggestion
enum class ComparisonOp { None, Lt, Le, Gt, Ge };
```

server/connector/search_filter_builder.cpp
```diff
@@ -0,0 +1,687 @@
...
+        std::forward<Args>(args)...);
+    case velox::TypeKind::BIGINT:
+      return std::forward<Func>(func).template
+      operator()<velox::TypeTraits<velox::TypeKind::BIGINT>::NativeType>(
+        std::forward<Args>(args)...);
+    case velox::TypeKind::DOUBLE:
+    case velox::TypeKind::REAL:
+    case velox::TypeKind::HUGEINT:
+      return std::forward<Func>(func).template operator()<double>(
+        std::forward<Args>(args)...);
```
MBkkt: I think we have float/real in iresearch

server/connector/search_filter_builder.cpp
```diff
@@ -0,0 +1,687 @@
...
+      return std::forward<Func>(func).template
+      operator()<velox::TypeTraits<velox::TypeKind::INTEGER>::NativeType>(
+        std::forward<Args>(args)...);
+    case velox::TypeKind::BIGINT:
+      return std::forward<Func>(func).template
+      operator()<velox::TypeTraits<velox::TypeKind::BIGINT>::NativeType>(
+        std::forward<Args>(args)...);
+    case velox::TypeKind::DOUBLE:
+    case velox::TypeKind::REAL:
+    case velox::TypeKind::HUGEINT:
```
MBkkt: hugeint => double is kind of wrong

server/connector/search_filter_builder.cpp
```diff
@@ -0,0 +1,687 @@
...
+
+template<typename T>
+void DoMangle(std::string& field_name) {
+  if constexpr (std::is_same_v<T, bool>) {
+    sdb::search::mangling::MangleBool(field_name);
+  } else if constexpr (std::is_same_v<T, velox::StringView>) {
+    sdb::search::mangling::MangleString(field_name);
+  } else if constexpr (std::is_floating_point_v<T> || std::is_integral_v<T>) {
+    sdb::search::mangling::MangleNumeric(field_name);
+  } else {
```
MBkkt, line 155:
```suggestion
void DoMangle(std::string& field_name) {
  if constexpr (std::is_same_v<T, bool>) {
    mangling::MangleBool(field_name);
  } else if constexpr (std::is_same_v<T, velox::StringView>) {
    sdb::search::mangling::MangleString(field_name);
  } else if constexpr (std::is_floating_point_v<T> || std::is_integral_v<T>) {
    sdb::search::mangling::MangleNumeric(field_name);
  } else {
```

server/connector/search_filter_builder.cpp
```diff
@@ -0,0 +1,687 @@
...
+  basics::StrResize(field_name, sizeof(column_id));
+  absl::big_endian::Store(field_name.data(), column_id);
+  return static_cast<const SereneDBColumn*>(it->second)->type()->kind();
+}
+
+// Maps velox kinds to native types used in iresearch
+template<typename Func, typename... Args>
+Result DispatchValue(velox::TypeKind kind, Func&& func, Args&&... args) {
+  irs::bstring term_value;
+  switch (kind) {
```
MBkkt, line 114: kind isn't full type. For an example, postgres actively use numeric(decimal), it will has kind = BIGINT/HUGEINT

server/connector/search_filter_builder.cpp
Dronplane: Currently we don't have int128 support. So we do this conversion in sink writer and should follow in source for now

server/connector/search_filter_builder.cpp
Dronplane: that "kind" is from column definition. So we just follow what is in the DDL

server/connector/search_data_source.cpp
```diff
@@ -0,0 +1,138 @@
...
+    }
+    if (irs::doc_limits::eof(doc_id)) {
+      _doc.reset();
+    }
+  }
+  if (index_keys.empty()) {
+    _current_split.reset();
+    return nullptr;
+  }
+
```
MBkkt, line 109: query rocksdb snapshot
!=
query search_data_store iresearch snapshot


query rocksdb snapshot
!=
query search_data_store rocksdb snapshot
query search_data_store iresearch snapshot

server/connector/search_filter_builder.cpp
Dronplane: Removed support of HUGEINT in search for now.

server/connector/search_filter_builder.cpp
Dronplane: Added assertions for type to NOT have custom comparison. As an assertions in this PR.
In the next PR InvertedIndex will check column types and fail creation if types can not be indexed.

server/connector/search_data_source.cpp
Dronplane: Added commit on tick

server/connector/search_data_source.cpp
```diff
@@ -0,0 +1,130 @@
...
+  // WriteBatchWithindex or plain DB with snapshot
+  const rocksdb::Snapshot* snapshot, rocksdb::DB& db,
+  rocksdb::ColumnFamilyHandle& cf, velox::RowTypePtr row_type,
+  std::vector<catalog::Column::Id> column_ids,
+  catalog::Column::Id effective_column_id, ObjectId object_key,
+  irs::IndexReader& reader, const irs::Filter::Query& query)
+  : Materializer(memory_pool, snapshot, &db, nullptr, cf, row_type,
+                 std::move(column_ids), effective_column_id, object_key),
+    _reader(reader),
+    _query(query) {}
```
MBkkt: codestyle: `{}`

server/connector/search_data_source.hpp
```diff
@@ -0,0 +1,71 @@
...
+                   // use just snapshot for now. But maybe we will need to have
+                   // this class template (or use some wrapper) to work with
+                   // WriteBatchWithindex or plain DB with snapshot
+                   const rocksdb::Snapshot* snapshot, rocksdb::DB& db,
+                   rocksdb::ColumnFamilyHandle& cf, velox::RowTypePtr row_type,
+                   std::vector<catalog::Column::Id> column_ids,
+                   catalog::Column::Id effective_column_id, ObjectId object_key,
+                   irs::IndexReader& reader, const irs::Filter::Query& query);
+
+  virtual ~SearchDataSource() = default;
```
MBkkt:
```suggestion
```
it doesn't make any sense, class is final

server/connector/search_filter_builder.cpp
```diff
@@ -0,0 +1,687 @@
...
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+///
+/// Copyright holder is SereneDB GmbH, Berlin, Germany
+////////////////////////////////////////////////////////////////////////////////
+
+#include "search_filter_builder.hpp"
+// NOLINTBEGIN
+// Need this header to stay first to avoid conflict in DCHECK macros
```
MBkkt: create issue for this plz, it shouldn't be an issue

server/query/transaction.cpp
```diff
@@ -36,13 +36,36 @@ Result Transaction::Begin() {
...
+      search_transaction.second->RegisterFlush();
+    }
+    auto status = _rocksdb_transaction->Commit();
+    if (!status.ok()) {
+      return {ERROR_INTERNAL,
+              "Failed to commit RocksDB transaction: ", status.ToString()};
+    }
+    // id is first write operation seqno in the WAL
+    auto post_commit_seq = _rocksdb_transaction->GetId();
+    SDB_ASSERT(post_commit_seq != 0);
```
MBkkt: 1. why we don't expect 0?
2. why something doesn't work if 0 here?

server/query/transaction.cpp
```diff
@@ -36,13 +36,36 @@ Result Transaction::Begin() {
...
+    // add number of operations to get last operation seqno
+    post_commit_seq += num_ops - 1;
+
+    for (auto& search_transaction : _search_transactions) {
+      search_transaction.second->Commit(post_commit_seq);
+    }
   }
   ApplyTableStatsDiffs();
```
MBkkt: We don't need to apply it if num_ops == 0

server/query/transaction.cpp
```diff
@@ -36,13 +36,36 @@ Result Transaction::Begin() {
...
+      // without it. This could happend if background index commit will start
+      // AFTER RocksDB commit but before transaction commit. But as we told
+      // index writer to wait, we are safe.
+      search_transaction.second->RegisterFlush();
+    }
+    auto status = _rocksdb_transaction->Commit();
+    if (!status.ok()) {
+      return {ERROR_INTERNAL,
+              "Failed to commit RocksDB transaction: ", status.ToString()};
+    }
```
MBkkt, line 71: we need to abort search transaction here/in case of exception

server/connector/search_sink_writer.cpp
```diff
@@ -276,7 +279,7 @@ void SearchSinkInsertBaseImpl::Field::SetNumericValue(T value) {
   if constexpr (std::is_same_v<
                   T, velox::TypeTraits<velox::TypeKind::HUGEINT>::NativeType>) {
     // TODO(Dronplane): Native int128 support
-    nstream.reset(static_cast<double>(value));
+    VELOX_UNSUPPORTED("HUGEINT kind is not supported");
```
MBkkt: SDB_THROW?

server/connector/search_filter_builder.cpp
```diff
@@ -0,0 +1,687 @@
...
+    }
+    return FromVeloxIsNull(filter, sub_ctx, call);
+  }
+
+  // Handle AND
+  if (call->name() == velox::expression::kAnd) {
+    return MakeGroup<irs::And>(filter, ctx, call);
+  }
+
+  // Handle OR
```
MBkkt: This applies to all comments in this function:

Avoid redundant comments that simply repeat what the code already says.
If needed prefer - comment explains WHY, not WHAT.

```suggestion
  if (call->name() == velox::expression::kAnd) {
    return MakeGroup<irs::And>(filter, ctx, call);
  }

```

server/connector/search_filter_builder.cpp
```diff
@@ -0,0 +1,687 @@
...
+  // Try equality/inequality
+
+  if (IsEqualityOp(call->name(), negated)) {
+    return FromVeloxBinaryEq(filter, ctx, call, negated);
+  }
+
+  // ByRange openended.
+  // BETWEEN is currently executed as conjunction of comparisons so it also goes
+  // here.
+  auto comparison_op = GetComparisonOp(call->name());
```
MBkkt:
```suggestion
  const auto comparison_op = GetComparisonOp(call->name());
```

server/connector/search_filter_builder.cpp
```diff
@@ -0,0 +1,687 @@
...
+  return {};
+}
+
+}  // namespace
+
+// Recursive conversion: Handle complex expressions (AND, OR, NOT)
+Result FromVeloxExpression(irs::BooleanFilter& filter,
+                           const VeloxFilterContext& ctx,
+                           const velox::core::TypedExprPtr& expr) {
+  auto* call = dynamic_cast<const velox::core::CallTypedExpr*>(expr.get());
```
MBkkt: they have kind btw
```
enum class ExprKind : int32_t {
  kInput = 0,
  kFieldAccess = 1,
  kDereference = 2,
  kCall = 3,
  kCast = 5,
  kConstant = 6,
  kConcat = 7,
  kLambda = 8,
};
```

Prefer to use it + downCast

server/connector/search_filter_builder.cpp
```diff
@@ -0,0 +1,687 @@
...
+  auto* field_typed = static_cast<const velox::core::FieldAccessTypedExpr*>(
+    call->inputs()[0].get());
+
+  std::string field_name;
+  auto kind = ExtractFieldName(ctx, *field_typed, field_name);
+  if (kind != velox::TypeKind::VARCHAR) {
+    return {ERROR_BAD_PARAMETER, "LIKE field is not VARCHAR"};
+  }
+  mangling::MangleString(field_name);
+  auto& wild_filter = ctx.negated ? Negate<irs::ByWildcard>(filter)
```
MBkkt:
```suggestion
  auto& wildcard_filter = ctx.negated ? Negate<irs::ByWildcard>(filter)
```

server/query/transaction.cpp
Dronplane: Good point. Completely forgot about scope exit with aborts

server/query/transaction.cpp
Dronplane: That was copied from old code. But indeed even if it is 0, that makes no harm here. Removed assert

server/connector/search_filter_builder.cpp
Dronplane: https://github.com/serenedb/serenedb/issues/230

PR #228, opened 2026-02-12
Make inverted index writers aware of columns

Dronplane:
+ Add indexing only specified columns
 + Add column validation for inverted index
 + Add column validation for PK columns
 + Add index writers for delete operation
 + Fix bug  - generated PK column should not be rewritten during PK update. 
 + Fix bug - all column ids was not filled for updates with index

### DIFF
--- a/server/connector/serenedb_connector.hpp
+++ b/server/connector/serenedb_connector.hpp
@@ -73,12 +73,12 @@ std::unique_ptr<SinkIndexWriter> MakeInvertedIndexWriter(
   if constexpr (Kind == axiom::connector::WriteKind::kInsert) {
     return std::make_unique<search::SearchSinkInsertWriter>(transaction,
                                                             columns);
-  }
-  if constexpr (Kind == axiom::connector::WriteKind::kUpdate) {
+  } else if constexpr (Kind == axiom::connector::WriteKind::kUpdate) {
     return std::make_unique<search::SearchSinkUpdateWriter>(transaction,
                                                             columns);
-  }
-  if constexpr (Kind == axiom::connector::WriteKind::kDelete) {
+  } else {
+    static_assert(Kind == axiom::connector::WriteKind::kDelete,
+                  "Unexpected WriteKind");
     return std::make_unique<search::SearchSinkDeleteWriter>(transaction);
   }
 }


server/connector/search_sink_writer.cpp
```diff
@@ -59,15 +59,25 @@ void SetNameToBuffer(std::string& name_buffer, catalog::Column::Id column_id) {
...
+  SDB_ASSERT(!_columns.empty());
 }
 
 bool SearchSinkInsertBaseImpl::SwitchColumnImpl(const velox::Type& type,
                                                 bool have_nulls,
                                                 catalog::Column::Id column_id) {
+  if (!_columns.contains(column_id)) {
+    _current_writer = nullptr;
+    return false;
+  }
```
MBkkt: Don't we want to move this logic to some generic place?
For an example I don't think we want to write it second time for secondary or other index

server/connector/search_sink_writer.hpp
```diff
@@ -193,8 +196,9 @@ class SearchSinkUpdateWriter final : public SinkUpdateWriter,
                                      public SearchSinkInsertBaseImpl,
                                      public SearchSinkDeleteBaseImpl {
  public:
-  SearchSinkUpdateWriter(irs::IndexWriter::Transaction& trx)
-    : SearchSinkInsertBaseImpl(trx), SearchSinkDeleteBaseImpl(trx) {}
+  SearchSinkUpdateWriter(irs::IndexWriter::Transaction& trx,
+                         std::span<const catalog::Column::Id> columns)
+    : SearchSinkInsertBaseImpl(trx, columns), SearchSinkDeleteBaseImpl(trx) {}
```
MBkkt: codestyle: `{}`

server/connector/search_sink_writer.hpp
```diff
@@ -154,8 +156,9 @@ class SearchSinkDeleteBaseImpl {
 class SearchSinkInsertWriter final : public SinkInsertWriter,
                                      public SearchSinkInsertBaseImpl {
  public:
-  SearchSinkInsertWriter(irs::IndexWriter::Transaction& trx)
-    : SearchSinkInsertBaseImpl(trx) {}
+  SearchSinkInsertWriter(irs::IndexWriter::Transaction& trx,
+                         std::span<const catalog::Column::Id> columns)
+    : SearchSinkInsertBaseImpl(trx, columns) {}
```
MBkkt: codestyle: `{}`

server/connector/search_sink_writer.cpp
```diff
@@ -86,7 +96,10 @@ bool SearchSinkInsertBaseImpl::SwitchColumnImpl(const velox::Type& type,
 
 void SearchSinkInsertBaseImpl::WriteImpl(
   std::span<const rocksdb::Slice> cell_slices, std::string_view full_key) {
-  SDB_ASSERT(_current_writer);
+  if (!_current_writer) {
+    // not indexing current column
+    return;
+  }
```
MBkkt: Maybe switch column should create kind of span that will be used for write?
In other words, do we want to have runtime check about this here?

server/catalog/local_catalog.cpp
```diff
@@ -1418,6 +1427,24 @@ Result LocalCatalog::CreateFunction(ObjectId database_id,
...
+    // PK must be default sortable or we can not guarantee table scan order
+    if (col->type->providesCustomComparison()) {
+      return {
+        ERROR_BAD_PARAMETER, "Column ", col->name,
+        " has type with custom comparison and can not be part of primary key"};
+    }
+    // this is current limitation of our pirmary key builder. And might be
+    // lifted some day.
+    if (!col->type->isPrimitiveType()) {
+      return {ERROR_BAD_PARAMETER, "Column ", col->name,
```
MBkkt, line 1443: I think same checks on write should became asserts?

server/connector/serenedb_connector.hpp
```diff
@@ -57,24 +57,82 @@
...
+  return absl::c_any_of(index_columns, [&](auto index_col) {
+    return updated_columns.contains(index_col);
+  });
+}
+
 template<typename Writer>
 std::vector<std::unique_ptr<Writer>> CreateIndexWriters(
-  ObjectId table_id, query::Transaction& transaction) {
+  ObjectId table_id, query::Transaction& transaction,
+  std::span<ColumnInfo> updated_columns = {}, bool pk_updated = false) {
```
```suggestion
  std::span<const ColumnInfo> updated_columns = {}, bool pk_updated = false) {
```

server/connector/serenedb_connector.hpp
```diff
@@ -57,24 +57,82 @@
 namespace sdb::connector {
 
 namespace {
+
+inline bool HasColumnOverlap(
+  std::span<const catalog::Column::Id> index_columns,
+  containers::FlatHashSet<catalog::Column::Id> updated_columns) {
```
```suggestion
  const containers::FlatHashSet<catalog::Column::Id>& updated_columns) {
```

server/connector/serenedb_connector.hpp
MBkkt: Do we need default values here? If it's not really needed maybe better to remove them?

server/connector/serenedb_connector.hpp
```diff
@@ -667,7 +725,12 @@ class SereneDBConnector final : public velox::connector::Connector {
 
           if constexpr (IsUpdate) {
             std::vector<catalog::Column::Id> all_column_oids;
-            if (table.UsedForUpdatePK()) {
+            auto update_sinks = CreateIndexWriters<SinkUpdateWriter>(
+              object_key, transaction,
+              std::span(columns.begin() + table.PKType()->size(),
+                        columns.end()),
```
```suggestion
              std::span{columns.begin() + table.PKType()->size(),
                        columns.end()},
```

server/query/transaction.h
```diff
@@ -98,15 +99,28 @@ class Transaction : public Config {
...
 
-  template<
-    InvocableWith<void(irs::IndexWriter::Transaction&, const IndexShard&),
-                  void(rocksdb::Transaction&, const IndexShard&)>
-      Visit>
-  void EnsureIndexesTransactions(ObjectId table_id, Visit&& visit) {
+  template<InvocableWith<void(irs::IndexWriter::Transaction&,
+                              std::span<const catalog::Column::Id>),
+                         void(rocksdb::Transaction&,
+                              std::span<const catalog::Column::Id>)>
```
MBkkt, line 105: Can we remove this concept or at least make it static_assert?

server/catalog/local_catalog.cpp
Dronplane: They already are. That was introduced in previous PR.

server/query/transaction.h
Dronplane: It was there before. I don't see any harm in it.  Why do you think it should be removed now?

server/connector/search_sink_writer.cpp
Dronplane: Done

server/connector/search_sink_writer.cpp
Dronplane: Done

server/connector/serenedb_connector.hpp
Dronplane: It's convenient as we do need this two arguments only for update, but not for insert and delete. I'd keep defaults here.

PR #232, opened 2026-02-13
Repeatable read and Read Commited support

mkornaukhov:
This PR adds:
- Syntax support for `BEGIN` and `SET TRANSACTION`
- Repeatable Read and Read Committed isolation levels
- Better anomalies sqllogic testing for supported isolation levels
- Transactions are recovery resistant

Follow ups:
- Decrease recovery test timings https://github.com/serenedb/serenedb/issues/242

### DIFF
--- a/server/connector/serenedb_connector.hpp
+++ b/server/connector/serenedb_connector.hpp
@@ -770,9 +770,16 @@ class SereneDBConnector final : public velox::connector::Connector {
               CreateIndexWriters<axiom::connector::WriteKind::kInsert>(
                 object_key, transaction);
             if (table.BulkInsert()) {
-              return std::make_unique<SSTInsertDataSink>(
-                _db, _cf, *connector_query_ctx->memoryPool(), object_key,
-                pk_indices, columns, std::move(insert_sinks));
+              const bool is_generated_pk = pk_indices.empty();
+              if (is_generated_pk) {
+                return std::make_unique<SSTInsertDataSink<true>>(
+                  _db, _cf, *connector_query_ctx->memoryPool(), object_key,
+                  pk_indices, columns, std::move(insert_sinks));
+              } else {
+                return std::make_unique<SSTInsertDataSink<false>>(
+                  _db, _cf, *connector_query_ctx->memoryPool(), object_key,
+                  pk_indices, columns, std::move(insert_sinks));
+              }
             }
 
             return std::make_unique<RocksDBInsertDataSink>(
--- a/server/connector/data_sink.cpp
+++ b/server/connector/data_sink.cpp
@@ -255,44 +255,53 @@ RocksDBInsertDataSink::RocksDBInsertDataSink(
     _conflict_resolver{transaction, cf, conflict_policy, table_name},
     _number_of_rows_affected{number_of_rows_affected} {}
 
-SSTInsertDataSink::SSTInsertDataSink(
+template<bool IsGeneratedPK>
+SSTInsertDataSink<IsGeneratedPK>::SSTInsertDataSink(
   rocksdb::DB& db, rocksdb::ColumnFamilyHandle& cf,
   velox::memory::MemoryPool& memory_pool, ObjectId object_key,
   std::span<const velox::column_index_t> key_childs,
   std::vector<ColumnInfo> columns,
   std::vector<std::unique_ptr<SinkIndexWriter>>&& index_writers)
-  : RocksDBDataSinkBase<SSTSinkWriter>(
-      SSTSinkWriter{db, cf, columns}, memory_pool, object_key, key_childs,
-      std::move(columns), std::move(index_writers)) {}
+  : Base(SSTSinkWriter<IsGeneratedPK>{object_key, db, cf, columns, memory_pool},
+         memory_pool, object_key, key_childs, std::move(columns),
+         std::move(index_writers)) {}
 
-void SSTInsertDataSink::appendData(velox::RowVectorPtr input) {
+template<bool IsGeneratedPK>
+void SSTInsertDataSink<IsGeneratedPK>::appendData(velox::RowVectorPtr input) {
   SDB_ASSERT(input->encoding() == velox::VectorEncoding::Simple::ROW);
-  SDB_ASSERT(input->type()->size() == _columns_info.size(),
-             "RocksDBDataSink: column oids size ", _columns_info.size(),
+  SDB_ASSERT(input->type()->size() == this->_columns_info.size(),
+             "RocksDBDataSink: column oids size ", this->_columns_info.size(),
              " doesn't match input type size ", input->type()->size());
   SDB_ASSERT(input->type()->kind() == velox::TypeKind::ROW);
 
-  const std::string table_key = key_utils::PrepareTableKey(_object_key);
   const auto num_rows = input->size();
   const auto num_columns = input->childrenSize();
 
-  _store_keys_buffers.clear();
-  _store_keys_buffers.reserve(num_rows);
-
+  this->_store_keys_buffers.clear();
+  this->_store_keys_buffers.reserve(num_rows);
   for (size_t row_idx = 0; row_idx < num_rows; ++row_idx) {
-    key_utils::MakeColumnKey<true>(
-      input, _key_childs, row_idx, table_key, [&](std::string_view row_key) {},
-      _store_keys_buffers.emplace_back());
+    auto& key_buffer = this->_store_keys_buffers.emplace_back();
+    if (!this->_key_childs.empty()) {
+      primary_key::Create(*input, this->_key_childs, row_idx, key_buffer);
+    } else {
+      const auto generated_pk =
+        std::bit_cast<int64_t>(RevisionId::create().id());
+      primary_key::AppendSigned(key_buffer, generated_pk);
+    }
   }
 
   velox::IndexRange all_rows(0, num_rows);
   const folly::Range all_rows_range{&all_rows, 1};
   for (velox::column_index_t i = 0; i < num_columns; ++i) {
-    if (_columns_info[i].id != catalog::Column::kGeneratedPKId) {
-      WriteInputColumn(_columns_info[i].id, i, *input, all_rows_range);
+    if (this->_columns_info[i].id != catalog::Column::kGeneratedPKId) {
+      this->WriteInputColumn(this->_columns_info[i].id, i, *input,
+                             all_rows_range);
     }
   }
 }
+
+template class SSTInsertDataSink<true>;
+template class SSTInsertDataSink<false>;
 
 RocksDBUpdateDataSink::RocksDBUpdateDataSink(
   std::string_view table_name, rocksdb::Transaction& transaction,
@@ -2415,7 +2424,9 @@ const std::string* RocksDBDataSinkBase<DataWriterType>::SetupRowKey(
   if (row_key.empty()) {
     return nullptr;
   }
-  key_utils::SetupColumnForKey(row_key, _column_id);
+  if constexpr (!kIsSSTSinkWriter<DataWriterType>) {
+    key_utils::SetupColumnForKey(row_key, _column_id);
+  }
 
   return &row_key;
 }
--- a/server/connector/data_sink.hpp
+++ b/server/connector/data_sink.hpp
@@ -294,7 +294,11 @@ class RocksDBUpdateDataSink final
   bool _update_pk{};
 };
 
-class SSTInsertDataSink final : public RocksDBDataSinkBase<SSTSinkWriter> {
+template<bool IsGeneratedPK>
+class SSTInsertDataSink final
+  : public RocksDBDataSinkBase<SSTSinkWriter<IsGeneratedPK>> {
+  using Base = RocksDBDataSinkBase<SSTSinkWriter<IsGeneratedPK>>;
+
  public:
   SSTInsertDataSink(
     rocksdb::DB& db, rocksdb::ColumnFamilyHandle& cf,
@@ -306,15 +310,18 @@ class SSTInsertDataSink final : public RocksDBDataSinkBase<SSTSinkWriter> {
   void appendData(velox::RowVectorPtr input) final;
 
   bool finish() final {
-    _data_writer.Finish();
+    this->_data_writer.Finish();
     return true;
   }
 
   void abort() final {
-    _data_writer.Abort();
-    RocksDBDataSinkBase<SSTSinkWriter>::abort();
+    this->_data_writer.Abort();
+    Base::abort();
   }
 };
+
+extern template class SSTInsertDataSink<true>;
+extern template class SSTInsertDataSink<false>;
 
 class RocksDBDeleteDataSink : public velox::connector::DataSink {
  public:
--- a/server/connector/key_utils.hpp
+++ b/server/connector/key_utils.hpp
@@ -42,7 +42,7 @@ std::string PrepareColumnKey(ObjectId id, catalog::Column::Id column_oid);
 void AppendColumnKey(std::string& key, catalog::Column::Id column_oid);
 
 // Prepare buffer for column key and call 'row_key_handle' on row_key
-template<bool IsInternalSSTKey = false, typename Func>
+template<typename Func>
 void MakeColumnKey(const velox::RowVectorPtr& input,
                    const std::vector<velox::column_index_t>& pk_columns,
                    velox::vector_size_t row_idx, std::string_view object_id,
@@ -63,15 +63,6 @@ void MakeColumnKey(const velox::RowVectorPtr& input,
   row_key_handle(std::string_view{
     key_buffer.begin() + sizeof(catalog::Column::Id), key_buffer.end()});
   std::memcpy(key_buffer.data(), object_id.data(), sizeof(ObjectId));
-
-  if constexpr (IsInternalSSTKey) {
-    const auto key_size = key_buffer.size();
-    basics::StrAppend(key_buffer,
-                      sizeof(rocksdb::SstFileWriter::kInternalKeyFooter));
-    std::memcpy(key_buffer.data() + key_size,
-                &rocksdb::SstFileWriter::kInternalKeyFooter,
-                sizeof(rocksdb::SstFileWriter::kInternalKeyFooter));
-  }
 }
 
 // Takes buffer in format
--- a/server/connector/sst_sink_writer.cpp
+++ b/server/connector/sst_sink_writer.cpp
@@ -27,7 +27,11 @@
 
 #include "basics/assert.h"
 #include "basics/random/random_generator.h"
+#include "basics/string_utils.h"
+#include "basics/system-compiler.h"
 #include "catalog/table_options.h"
+#include "iresearch/utils/bytes_utils.hpp"
+#include "key_utils.hpp"
 #include "rocksdb/options.h"
 #include "rocksdb/table.h"
 #include "rocksdb_engine_catalog/rocksdb_engine_catalog.h"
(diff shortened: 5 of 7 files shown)


server/query/transaction.cpp
```diff
@@ -29,15 +29,27 @@
...
+    ApplyTableStatsDiffs();
+    CommitVariables();
+    Destroy();
+  };
+
+  // RocksDB transaction is lazy initialized, so it may be nullptr here. It
+  // means no rocksdb transaction actually started, for example, sequential
+  // BEGIN and COMMIT statements.
+  if (!_rocksdb_transaction) {
+    cleanup();
```
Dronplane: Do we need to call ApplyTableStatsDiffs (from cleanup) here? As it looks like there is nothing to update if there were no rocksdb transaction. Maybe just move abort_guard before that condition and do just return.

server/query/transaction.cpp
```diff
@@ -29,15 +29,27 @@
...
+    return {};
```
Dronplane: Let's add an assert here that there is no _search_transactions either. As search transactions have implicit commit in destructor so it would be good to make sure there is no such things here.

server/query/transaction.cpp
```diff
@@ -77,26 +94,29 @@ Result Transaction::Commit() {
     for (auto& search_transaction : _search_transactions) {
       search_transaction.second->Commit(post_commit_seq);
     }
-    std::move(abort_guard).Cancel();
-    ApplyTableStatsDiffs();
   }
+  std::move(abort_guard).Cancel();
+  cleanup();
```
Dronplane: same here - if num_ops == 0 we do not need to call ApplyTableStatsDiffs in cleanup.

mkornaukhov: Fixed

server/query/transaction.cpp
mkornaukhov: Fixed

server/query/transaction.cpp
mkornaukhov: Fixed

server/query/transaction.cpp
```diff
@@ -149,7 +173,6 @@ void Transaction::CreateStorageSnapshot() {
 }
 
 void Transaction::CreateRocksDBTransaction() {
-  SDB_ASSERT(!_rocksdb_snapshot);
```
Dronplane, line 152: Why this was removed?  We should not have snapshot updates here if I get it right.

server/connector/serenedb_connector.hpp
```diff
@@ -575,16 +575,20 @@ class SereneDBConnector final : public velox::connector::Connector {
 
     auto* rocksdb_transaction = transaction.GetRocksDBTransaction();
 
+    const rocksdb::Snapshot* snapshot =
+      transaction.GetIsolationLevel() == IsolationLevel::ReadCommitted
+        ? nullptr
+        : &transaction.EnsureRocksDBSnapshot();
```
Dronplane: Can we move this logic inside transaction.EnsureRocksDBSnapshot? So it will give us nullptr snapshot if isolation level is set as read committed. And add transaction.GetSnapshot with similar logic. Main idea is that transaction object should carry logic of snapshot management not the code outside.

server/query/transaction.cpp
```diff
@@ -165,6 +188,7 @@ void Transaction::CreateRocksDBTransaction() {
 
 void Transaction::Destroy() noexcept {
   _state = State::None;
+  _isolation_level = IsolationLevel::RepeatableRead;
```
Dronplane: does it make any sense to update value to default here? object is not usable anyway after destroy.

server/query/transaction.cpp
mkornaukhov: For example, when in transaction we firstly do reads, than writes, we may have snapshot for reads, and initiating rocksdb transaction only when we want write.

server/query/transaction.cpp
mkornaukhov: Object is reassigned after destroy, in `Begin()` method. It has isolation level as argument, so I'll remove it

server/query/transaction.cpp
Dronplane: If we store snapshot here  - that means we have RepeatableRead. But in that case we should not have different snapshots for different queries in transaction.

server/connector/serenedb_connector.hpp
mkornaukhov: Fixed

server/query/transaction.cpp
mkornaukhov: Fixed

server/query/transaction.cpp
```diff
@@ -149,22 +162,21 @@ void Transaction::CreateStorageSnapshot() {
 }
 
 void Transaction::CreateRocksDBTransaction() {
-  SDB_ASSERT(!_rocksdb_snapshot);
   SDB_ASSERT(!_rocksdb_transaction);
+  SDB_ASSERT(!_rocksdb_snapshot);
```
Dronplane: NIT better to keep old order of asserts to reduce changeset

server/connector/serenedb_connector.hpp
```diff
@@ -644,16 +644,16 @@ class SereneDBConnector final : public velox::connector::Connector {
 
     auto* rocksdb_transaction = transaction.GetRocksDBTransaction();
 
+    const auto* snapshot = &transaction.EnsureRocksDBSnapshot();
     if (read_your_own_writes && rocksdb_transaction) {
       return std::make_unique<RocksDBRYOWDataSource>(
         *connector_query_ctx->memoryPool(), *rocksdb_transaction, _cf,
         output_type, column_oids, serene_table_handle.GetEffectiveColumnId(),
-        object_key);
+        object_key, snapshot);
```
Dronplane: I guess changes in RocksDBRYOWDataSource are not necessary. We only need to call transaction.EnsureRocksDBSnapshot() before like it is done here in this PR and rocksdb_transaction will already have desired snapshot. So code inside RocksDBRYOWDataSource can just use transaction.GetSnapshot like before.

server/query/transaction.cpp
```diff
@@ -136,6 +145,10 @@ rocksdb::Transaction& Transaction::EnsureRocksDBTransaction() {
   if (!_rocksdb_transaction) {
     CreateRocksDBTransaction();
   }
+  if (!_rocksdb_snapshot &&
+      _isolation_level == IsolationLevel::RepeatableRead) {
+    SetTransactionSnapshot();
```
Dronplane: BTW does it makes sense to lazy-init transaction snapshot? 
If we are starting to "bootstrap" snapshot from data sources - EnsureRocksDBSnapshot() would be called anyway and snapshot would be set with all isolation levels.  Even If we start from  data sink with read committed -  we will call EnsureRocksDBTransaction() which may leave snapshot  uinitialized but next creation of data source will still set snapshot in rocksdb transaction. So the only case when we will run without snapshot  is  first statement in transaction if it is INSERT that do not require DataSource (e.g. INSERT VALUES(...) ). But even in this case running second same INSERT will trigger  OnNewStatement() and again set snapshot into transaction.

mkornaukhov: It's Postgres behavior, I've added test for that.
It should take snapshot on first read/write operation, not on the `begin` statement.

It hypothetically decreases probability of serialization conflict.

server/query/transaction.cpp
mkornaukhov: Fixed

server/connector/serenedb_connector.hpp
mkornaukhov: Yes, fixed

tests/sqllogic/sdb/pg/settings/show_all.test
```diff
@@ -12,6 +12,7 @@ sdb_read_your_own_writes	true	Controls whether queries can see uncommitted write
 client_encoding	UTF8	Sets the client character set encoding.
 datestyle	ISO, MDY	Sets the display format for date and time values.
 application_name	(empty)	Sets the application name to be reported in statistics and logs.
+default_transaction_isolation	repeatable read	Sets the transaction isolation level of each new transaction.
```
MBkkt, line 15: is this our or postgres setting?
```suggestion
default_transaction_isolation	repeatable read	Sets the default (implicit) transaction isolation level of each new transaction.
```

server/query/cursor.cpp
```diff
@@ -55,6 +55,7 @@ void Cursor::RequestCancel() {
 
 Cursor::Process Cursor::Next(velox::RowVectorPtr& batch) {
   const auto& query_ctx = _query.GetContext();
+  query_ctx.transaction->OnNewStatement();
```
MBkkt: wrong place

server/query/config.h
```diff
@@ -54,13 +54,19 @@ enum class VariableType {
   PgExtraFloatDigits,
   PgByteaOutput,
   SdbWriteConflictPolicy,
+  PgTransactionIsolation,
```
MBkkt: We use `Pg` for pg settings. Is it pg settings?

server/pg/commands/transaction.cpp
```diff
@@ -32,15 +33,44 @@ yaclib::Future<Result> Transaction(ExecContext& context,
...
+          }
+        });
+        return yaclib::MakeFuture(conn_ctx.Begin(isolation_level));
       } else {
         conn_ctx.AddNotice(SQL_ERROR_DATA(
           ERR_CODE(ERRCODE_ACTIVE_SQL_TRANSACTION),
           ERR_MSG("there is already a transaction in progress")));
       }
       break;
+    }
```
MBkkt:
```suggestion
    } break;
```

libs/basics/system-compiler.h
```diff
@@ -49,3 +49,8 @@
 #else
 #define SDB_PRETTY_FUNCTION __func__
 #endif
+
+// TODO(mkornaukhov) use `__builtin_trap()` for clang and gcc,
+// but it makes recovery tests two times longer.
+// immediate program termination
```
MBkkt: comment is misleading, it doesn't  use builtin trap

tests/sqllogic/sdb/pg/settings/show_all.test
MBkkt: if it's postgres, write test plz

server/query/config.h
MBkkt: If it's pg settings name should be `PgDefaultTransactionIsolation`
or it should be `SdbOurName`

server/pg/commands/variable_set.cpp
```diff
@@ -121,6 +122,40 @@ yaclib::Future<Result> VariableSet(ExecContext& ctx,
     conn_ctx.ResetAll();
     return {};
   }
+
+  if (stmt.kind == VAR_SET_MULTI) {
+    std::string_view name = stmt.name;
+    if (name == "TRANSACTION" || name == "SESSION CHARACTERISTICS") {
```
MBkkt: what should be if it's called inside transaciton

Add more tests

server/pg/commands/variable_set.cpp
```diff
@@ -189,7 +224,7 @@ yaclib::Future<Result> VariableSet(ExecContext& ctx,
       r = {ERROR_NOT_IMPLEMENTED, "SET ... TO CURRENT is not implemented"};
       break;
     case VAR_SET_MULTI:
-      r = {ERROR_NOT_IMPLEMENTED, "SET ... TO MULTI is not implemented"};
+      SDB_UNREACHABLE();
       break;
     default:
```
MBkkt: just default?

server/pg/commands/transaction.cpp
```diff
@@ -32,15 +33,44 @@ yaclib::Future<Result> Transaction(ExecContext& context,
   auto& conn_ctx = basics::downCast<ConnectionContext>(context);
   switch (stmt.kind) {
     case TRANS_STMT_BEGIN:
-    case TRANS_STMT_START:
+    case TRANS_STMT_START: {
       if (!conn_ctx.HasTransactionBegin()) {
-        return yaclib::MakeFuture(conn_ctx.Begin());
+        auto isolation_level =
```
MBkkt: maybe?

Begin();
SetConfig(default_transaction_isolation, "")



`BEGIN TRANSACTION repeatable read`
=> 
`BEGIN TRANSACTION`
+
`SET CONFIG repeatable read`

server/query/transaction.h
```diff
@@ -144,11 +155,15 @@ class Transaction : public Config {
   void CreateStorageSnapshot();
   void CreateRocksDBTransaction();
   void ApplyTableStatsDiffs();
+  void SetTransactionSnapshot() const;
 
   State _state = State::None;
+  IsolationLevel _isolation_level =
```
MBkkt: create connection
set isolation level
// destroy and it probably reads isolation level from config?
next query

server/query/transaction.h
```diff
@@ -68,7 +68,14 @@ class Transaction : public Config {
   }
 #endif
 
-  Result Begin();
+  void OnNewStatement() const {
+    if (_isolation_level == IsolationLevel::ReadCommitted &&
+        _rocksdb_transaction) {
+      SetTransactionSnapshot();
+    }
```
MBkkt: maybe we want reset snapshot here?

to make it more similar between first and next statements of transaction

server/connector/serenedb_connector.hpp
```diff
@@ -642,18 +642,19 @@ class SereneDBConnector final : public velox::connector::Connector {
...
       transaction.Get<VariableType::Bool>("sdb_read_your_own_writes");
 
-    auto* rocksdb_transaction = transaction.GetRocksDBTransaction();
-
-    if (read_your_own_writes && rocksdb_transaction) {
+    const auto* snapshot = &transaction.EnsureRocksDBSnapshot();
+    if (read_your_own_writes && transaction.HasRocksDBWrite()) {
+      auto* rocksdb_transaction = transaction.GetRocksDBTransaction();
+      SDB_ASSERT(rocksdb_transaction);
+      SDB_ASSERT(snapshot == rocksdb_transaction->GetSnapshot());
```
MBkkt: Test about this logic (no read_your_own_writes for begin/commit, if there wasn't write operations)

```
set fail_if_read_your_own_writes to true;
begin
read // we checked that this read doesn't use read own writes
insert
set fail_if_read_your_own_writes to false;
set fail_if_not_read_your_own_writes to false; // optionally
read // we checked that this read does use read own writes
abort/commit
```

server/query/transaction.cpp
```diff
@@ -64,7 +65,11 @@ Result Transaction::Commit() {
       // index writer to wait, we are safe.
       search_transaction.second->RegisterFlush();
     }
+
```
MBkkt, line 65: probably move abort to here

server/query/transaction.cpp
```diff
@@ -123,11 +131,16 @@ const rocksdb::Snapshot& Transaction::EnsureRocksDBSnapshot() {
   SDB_ASSERT((_state & State::HasRocksDBRead) != State::None);
   if (!_rocksdb_snapshot) {
     if ((_state & State::HasRocksDBWrite) != State::None) {
```
MBkkt: HasRocksdbWrite/etc

server/query/transaction.cpp
```diff
@@ -123,11 +131,16 @@ const rocksdb::Snapshot& Transaction::EnsureRocksDBSnapshot() {
...
-      CreateRocksDBTransaction();
+      if (!_rocksdb_transaction) {
+        CreateRocksDBTransaction();
+      }
+      SetTransactionSnapshot();
     } else {
       CreateStorageSnapshot();
```
MBkkt: this code can be executed only for single read-only statement (not inside transaction)

server/query/transaction.cpp
Dronplane: What bothers me is that we now create rocksdb transaction in Begin but do not set snapshot. And that transaction might be unused if there are no write operations. I see the point of postponing taking snapshot like postgres does and agree with that. My point is that we should not have the case when we have created transaction but not taken snapshot.  Anyway all "Ensure***" methods again check if transaction present and create it if necessary. And if Ensure*** method called that means we are already executing something so we should take rocksdb snapshot regardless of isolation level. Wither drop this snapshot or not after executing query is indeed determined by isolation as it is implemented here.  So I suggest that Begin should just set the HaveTransaction flag and lock isolation level. BTW there was an attempt to make transaction object optional in this PR - but for some reason it was rollbacked.

MBkkt: Implementation will be different, no point to discuss this place, here will be just set snapshot (member of our transaction class) to nullptr

mkornaukhov: > So I suggest that Begin should just set the HaveTransaction flag and lock isolation level

No, that is not possible, I've written comment about that.

server/query/config.h
mkornaukhov: Same type is used for 2 variables, I've renamed into `SdbTransactionIsolation`

libs/basics/system-compiler.h
mkornaukhov: Fixed comment, it's about intention

server/query/transaction.cpp
mkornaukhov: fixed

server/query/transaction.cpp
mkornaukhov: Fixed

server/query/transaction.cpp
mkornaukhov: Fixed

tests/sqllogic/sdb/pg/settings/show_all.test
mkornaukhov: It's postres'. We cannot test description normally without `pg_settings` system table/view

server/pg/commands/variable_set.cpp
mkornaukhov: Fixed

server/pg/commands/transaction.cpp
mkornaukhov: For now it looks approximately like that

server/query/transaction.h
mkornaukhov: Fixed

server/pg/commands/variable_set.cpp
mkornaukhov: Added

server/connector/serenedb_connector.hpp
mkornaukhov: Added

server/connector/serenedb_connector.hpp
```diff
@@ -642,18 +642,32 @@ class SereneDBConnector final : public velox::connector::Connector {
...
+#ifdef SDB_FAULT_INJECTION
+      // failpoints are per process so we make unique name to allow multiple
+      // sqlogic tests run in parallel without interference of failpoints
+      // TODO(mkornaukhov): Find a better way. Maybe make failpoints database
+      // bindable to allow parallel execution.
+      auto table_ptr =
+        transaction.GetCatalogSnapshot()->GetObject<catalog::Table>(object_key);
+      SDB_ASSERT(table_ptr);
+      auto fail_on_ryow = absl::StrCat(table_ptr->GetName(), "_fail_on_ryow");
+      SDB_IF_FAILURE(fail_on_ryow) { SDB_IMMEDIATE_ABORT(); }
```
MBkkt: throw instead of crash?

server/query/config_variables.cpp
```diff
@@ -117,6 +117,24 @@ constexpr auto kVariableDescription =
         "",
       },
     },
+    // TODO(mkornaukhov) test description. Looks like <slt:ignore> for certain
+    // level is requires
```
MBkkt: > Looks like <slt:ignore> for certain level is requires

What does this mean?

server/pg/commands/variable_set.cpp
```diff
@@ -121,6 +128,87 @@ yaclib::Future<Result> VariableSet(ExecContext& ctx,
...
+  if (stmt.kind == VAR_SET_MULTI) {
+    std::string_view name = stmt.name;
+
+    auto get_isolation_level = [](const VariableSetStmt& stmt) {
+      std::string_view level;
+      VisitNodes(stmt.args, [&](const DefElem& option) {
+        std::string_view opt_name = option.defname;
+        if (opt_name == "transaction_isolation") {
+          level = strVal(&castNode(A_Const, option.arg)->val);
+          if (!ValidateValue(VariableType::SdbTransactionIsolation, level)) {
```
MBkkt: for not supported we need special function that will use validate?

server/pg/commands/variable_set.cpp
```diff
@@ -121,6 +128,87 @@ yaclib::Future<Result> VariableSet(ExecContext& ctx,
...
+        } else if (opt_name == "transaction_read_only") {
+          THROW_SQL_ERROR(
+            ERR_CODE(ERRCODE_FEATURE_NOT_SUPPORTED),
+            ERR_MSG("transaction READ WRITE | READ ONLY is not supported yet"));
+        } else if (opt_name == "transaction_deferrable") {
+          THROW_SQL_ERROR(
+            ERR_CODE(ERRCODE_FEATURE_NOT_SUPPORTED),
+            ERR_MSG("transaction DEFERRABLE is not supported yet"));
+        }
+        SDB_UNREACHABLE();
```
MBkkt: throw

server/pg/commands/variable_set.cpp
```diff
@@ -86,6 +87,12 @@ Result ProcessValues(const List* list, std::string_view name, VariableType type,
     return {};
   }
 
+  if (context == Config::VariableContext::Session &&
+      name == "transaction_isolation") {
+    // Silently ignore
+    return {};
+  }
```
MBkkt: move ignore below

server/pg/commands/variable_set.cpp
MBkkt, line 101: We want to set transaction_isolation if right now we're setting default_transaction_isolation

MBkkt, line 275: if (transaction) ProcessTransactionVar
if (default_transaction) ProcessDefaultTransactionVar

case Multi:
  if (...) ProcessTransactionVar
  else if (...) ProcessDefaultTransactionVar

server/query/config_variables.cpp
mkornaukhov: Made issue https://github.com/serenedb/serenedb/issues/251

server/pg/commands/variable_set.cpp
mkornaukhov: I've added SDB_ASSERT, but it is not arbitrary string, it's from grammar

server/pg/commands/variable_set.cpp
mkornaukhov: Get rid of this at all

server/connector/serenedb_connector.hpp
mkornaukhov: Fixed

server/pg/commands/variable_set.cpp
mkornaukhov: fixed

PR #255, opened 2026-02-24
feat: Support CREATE TABLE USING EXTERNAL WITH (FILE)

pashandor789:


server/query/utils.cpp
```diff
@@ -49,4 +49,17 @@ bool Equals(const axiom::logical_plan::Expr* lhs,
   return !lhs && !rhs;
 }
 
+std::string_view ToAlias(std::string_view name) {
+  return name.substr(0, name.find_last_of(kColumnSeparator));
+}
+
+std::vector<std::string> ToAliases(std::span<const std::string> names) {
```
MBkkt, line 56: Why do we need to allocate strings for return values?

Issue #217, opened 2026-02-07
Better rocksdb data source


MBkkt:
We should support following cases (for primary key and unique secondary indexes):

1) multiple point lookup, should be handled via MultiGet (where keys is number of columns * number of points)
2) single range lookup, should be handled via iterator seek. Iterator per column (with optionally parallel iterator reads). prefix also is range
3) combination of multiple point lookups and single range lookups, probably easier to separate such things to multiple Velox splits, in such case they can be read in parallel and at the same time complexity of implementation will be lower.
4) we can try to implement multiple range lookup
5) support filter pushdown for wildcard/some random queries/etc, everything that is ok to compute at one key/value (range queries should be computed with range iterator + sometimes post filtering)
6) support lazy columns

We need to implement Wait for iterator, to be able cache iterator and use async_io for it.

PR #267, opened 2026-03-02
Point lookup data source

mkornaukhov:
Adds point lookup data source, see #217
Also add possibility to execute pushed down filters inside data source

### DIFF
--- a/server/connector/data_source.cpp
+++ b/server/connector/data_source.cpp
@@ -63,18 +63,18 @@ void SetResultValue(std::string_view value, size_t idx,
 
 template<velox::TypeKind Kind>
 velox::VectorPtr ReadPointColumnValues(
-  size_t sort_rank, size_t num_points,
+  size_t sort_rank, size_t read_points_count, size_t write_points_count,
   const std::vector<rocksdb::Status>& statuses,
   const std::vector<std::string>& values,
   velox::memory::MemoryPool& memory_pool) {
   using T = typename velox::TypeTraits<Kind>::NativeType;
 
   auto result = velox::BaseVector::create<velox::FlatVector<T>>(
-    velox::Type::create<Kind>(), num_points, &memory_pool);
+    velox::Type::create<Kind>(), write_points_count, &memory_pool);
 
   size_t written = 0;
-  for (size_t read_idx = 0; read_idx < num_points; ++read_idx) {
-    const size_t idx = sort_rank * num_points + read_idx;
+  for (size_t read_idx = 0; read_idx < read_points_count; ++read_idx) {
+    const size_t idx = sort_rank * read_points_count + read_idx;
     const auto& status = statuses[idx];
     const auto& value = values[idx];
 
@@ -87,7 +87,7 @@ velox::VectorPtr ReadPointColumnValues(
     SDB_ENSURE(status.ok(), rocksutils::ConvertStatus(status));
     SetResultValue(value, written++, *result);
   }
-  result->resize(written);
+  SDB_ASSERT(written == write_points_count);
   return result;
 }
 
@@ -578,7 +578,9 @@ std::optional<velox::RowVectorPtr> RocksDBPointLookupDataSource<Source>::next(
   // rows are present. This may allow us to reduce number of multiget calls
   BuildKeys(batch_size, num_columns);
   PerformMultiGet(batch_size * num_columns);
-  CheckAndCountFound(batch_size * num_columns);
+  size_t found_count = CheckAndCountFound(batch_size * num_columns);
+  SDB_ASSERT(found_count % num_columns == 0);
+  found_count /= num_columns;
 
   std::vector<velox::VectorPtr> columns;
   columns.reserve(num_columns);
@@ -586,10 +588,9 @@ std::optional<velox::RowVectorPtr> RocksDBPointLookupDataSource<Source>::next(
     const auto& type = _read_type->childAt(column_idx);
     columns.emplace_back(VELOX_DYNAMIC_SCALAR_TYPE_DISPATCH(
       ReadPointColumnValues, type->kind(), _col_rank[column_idx], batch_size,
-      _statuses, _raw_values, _memory_pool));
+      found_count, _statuses, _raw_values, _memory_pool));
   }
 
-  const size_t found_count = columns[0]->size();
   SDB_ASSERT(
     absl::c_all_of(columns,
                    [&](const auto& col) { return col->size() == found_count; }),
--- a/tests/sqllogic/sdb/pg/data_source/choice.test
+++ b/tests/sqllogic/sdb/pg/data_source/choice.test
@@ -146,6 +146,20 @@ EXPLAIN (ANALYZE) SELECT * FROM t_choice_cpk WHERE (a=1 AND b=2) OR (a=3 AND b=4
 QUERY PLAN
 <slt:ignore>[t_choice_cpk, scan=point]<slt:ignore>
 
+query
+SELECT * FROM t_choice_cpk WHERE a=1 AND (b=2 OR b=4) ORDER BY a, b;
+----
+a	b	val
+1	2	10
+1	4	11
+
+onlyif serenedb
+query
+EXPLAIN (ANALYZE) SELECT * FROM t_choice_cpk WHERE a=1 AND (b=2 OR b=4) ORDER BY a, b;
+----
+QUERY PLAN
+<slt:ignore>[t_choice_cpk, scan=point]<slt:ignore>
+
 
 # Cartesian product of {a:1},{b:2} x {a:2},{b:4}:
 #   {a:1} and {a:2} = contradiction


tests/sqllogic/recovery/faults.test
MBkkt, line 1: this already committed in main?

server/connector/serenedb_connector.hpp
```diff
@@ -706,7 +724,15 @@ class SereneDBConnector final : public velox::connector::Connector {
         search_snapshot.reader, *serene_table_handle.GetSearchQuery());
     }
 
-    return std::make_unique<RocksDBSnapshotDataSource>(
+    auto points = TryGetPoints(filter, serene_table_handle.GetPKType(),
+                               connector_query_ctx->memoryPool());
+    if (points) {
```
MBkkt: This is bad, we want this single time per query, datasource can be create many times per single query

server/connector/serenedb_connector.cpp
```diff
@@ -135,11 +139,13 @@ SereneDBTableLayout::createTableHandle(
                                              std::move(remaining_filter));
   }
 
+  const auto& pk_type = basics::downCast<RocksDBTable>(table()).PKType();
+  auto filter = ParseFilters(filters, pk_type->names());
   rejected_filters = std::move(filters);
```
MBkkt: I think we want to pushdown all filters. And if possible create datasource that isn't full scan.

See HiveDataSource to understand how to execute filters inside DataSource

tests/sqllogic/sdb/pg/data_source/choice.test
```diff
@@ -0,0 +1,270 @@
...
+1	6	12
+2	2	21
+3	2	30
+3	4	20
+3	6	32
+5	6	30
+
+# OrFilter of two complete PK conjunctions -> 2 specific points -> MultiGet.
+query
+SELECT * FROM t_choice_cpk WHERE (a=1 AND b=2) OR (a=3 AND b=4) ORDER BY a, b;
```
MBkkt, line 136: We are also very want to support BETWEEN and IN

server/connector/data_source.cpp
```diff
@@ -31,16 +31,71 @@
...
+      // Filtered point is not found, it's OK, just skip
+      continue;
+    }
+
+    if (!status.ok()) {
+      auto res = rocksutils::ConvertStatus(status);
+      SDB_THROW(std::move(res));
+    }
+
+    if constexpr (std::is_same_v<T, velox::StringView>) {
```
Dronplane: Looks like copy&paste from FullScan datasource. Could we move this part to some helper method?

server/connector/data_source.cpp
```diff
@@ -352,4 +407,82 @@ uint64_t RocksDBDataSource::IterateColumn(rocksdb::Iterator& it,
...
+  if (!_current_split) {
+    return nullptr;
+  }
+
+  SDB_ASSERT(_values);
+  SDB_ASSERT(_values->size() > 0,
+             "Case of empty filters should be processed in connector");
+
+  const auto num_columns = _row_type->size();
+  SDB_ASSERT(num_columns > 0);
```
Dronplane: Could we optimize COUNT(*) with points? Handle 0 columns same way as for full scan.

server/connector/data_source.cpp
```diff
@@ -352,4 +407,82 @@ uint64_t RocksDBDataSource::IterateColumn(rocksdb::Iterator& it,
...
+    return nullptr;
+  }
+
+  const size_t total_keys = batch_size * num_columns;
+
+  _keys.resize(total_keys);
+  for (size_t point_idx = 0; point_idx < batch_size; ++point_idx) {
+    for (size_t col_idx = 0; col_idx < num_columns; ++col_idx) {
+      auto& key = _keys[point_idx * num_columns + col_idx];
+      key.clear();
```
Dronplane: "Table" part of the key seems constant for every point in every column. Maybe it would be faster here to check if key is not empty and if it is then just resize it down to sizeof(ObjectId) and if key is empty then indeed put table key. But need some benchmarking - maybe having here a condition would make it worse.

server/connector/data_source.cpp
```diff
@@ -352,4 +407,82 @@ uint64_t RocksDBDataSource::IterateColumn(rocksdb::Iterator& it,
...
+             "Case of empty filters should be processed in connector");
+
+  const auto num_columns = _row_type->size();
+  SDB_ASSERT(num_columns > 0);
+
+  const auto total_points = static_cast<size_t>(_values->size());
+  const auto batch_size =
+    std::min(static_cast<size_t>(size), total_points - _offset);
+
+  if (batch_size == 0) {
```
Dronplane: This condition seems redundant. batch_size could be 0 in three cases. When size == 0 but that is asserted out. When _offset  == total_points but that is also impossible because check at the end of the method will reset split and set offset to 0 in that case. So this could only happen if total_points == 0 but that means _values->size() == 0 but that is also asserted out.

server/connector/data_source.hpp
```diff
@@ -21,21 +21,72 @@
...
 class SereneDBConnectorSplit;
 
-class RocksDBDataSource : public velox::connector::DataSource {
+class MultiGetContext {
+ public:
+  // TODO benchmark and choose best threshold
+  static constexpr size_t kThreshold = 1;
+  static constexpr size_t kBatchSize = 32;
+
+  template<
```
Dronplane: Maybe just pass to MultiGet method a data source as templated argument. Source could be DB or Transaction.
Both has Get/MultiGet method with same signatures.
So instead of passing 2 wrappers for calling Get/MultiGet on corresponding source we will just pass source. And RYOW/Snapshot will pass transaction/db accordingly.
And instead of batch_fn/single_fn just call
src.MultiGet/src.Get

server/connector/data_source.hpp
```diff
@@ -96,36 +149,160 @@ class RocksDBDataSource : public velox::connector::DataSource {
...
+  std::vector<catalog::Column::Id> _column_ids;
+  ObjectId _object_key;
+  std::shared_ptr<velox::connector::ConnectorSplit> _current_split;
+  uint64_t _produced = 0;
+  size_t _offset = 0;
+  velox::RowVectorPtr _values;
+  std::vector<std::string> _keys;
+  std::vector<rocksdb::Slice> _key_slices;
+  std::vector<std::string> _raw_values;
+  std::vector<rocksdb::Status> _statuses;
```
Dronplane, line 325: why not just std::array? Batch size is limited by compile time value

server/connector/rocksdb_filter.cpp
```diff
@@ -0,0 +1,367 @@
...
+  return std::make_unique<EqFilterNode>(field_access->name(), const_val,
+                                        pk_names);
+}
+
+std::unique_ptr<FilterNode> ParseIn(const velox::core::CallTypedExpr* func_call,
+                                    std::span<const std::string> pk_names) {
+  if (func_call->inputs().size() != 2) {
+    return nullptr;
+  }
+  if (func_call->inputs()[0]->kind() != velox::core::ExprKind::kFieldAccess ||
```
Dronplane: NIT
```suggestion
  if (!func_call->inputs()[0]->isFieldAccessKind() ||
```

server/connector/rocksdb_filter.cpp
```diff
@@ -0,0 +1,367 @@
...
+      func_call->inputs()[1]->kind() != velox::core::ExprKind::kConstant) {
```
Dronplane: NIT

```suggestion
      !func_call->inputs()[1]->isConstantKind()) {
```

server/connector/data_source.cpp
mkornaukhov: It never happens for now because for lookup points we need a filter that accesses columns of a table, so for now it will never will be called

tests/sqllogic/sdb/pg/data_source/choice.test
mkornaukhov: Between logic is for range queries, I think we want it in another PR.
Tests with IN are present in this file

server/connector/data_source.cpp
mkornaukhov: As row type is not changed, even column id is stable.
I suppose better to have separate code blocks for that

server/connector/rocksdb_filter.hpp
```diff
@@ -0,0 +1,147 @@
...
+  explicit OrFilterNode(std::vector<std::unique_ptr<FilterNode>> filters);
+
+  [[nodiscard]] std::vector<Point> NextPoints() final;
+
+ private:
+  std::vector<std::unique_ptr<FilterNode>> _children;
+  size_t _current = 0;
+};
+
+// Returns nullptr for expressions that cannot be parsed as a specific filter.
```
MBkkt: (c1 = 1 or c2 = 2) and (c1 = 3 or c2 = 4)


-- hashmap<columnid, vector<Range>> conjunctions;

vector<Struct<hashmap<columnid, range>, TypedExprPtr>>

struct State {
};

walkOver(TypedExprPtr&, State&) {
   state
   
   rem 
}

c1: "A",     "B",  "D"
c2:  ANY,  "C",  "P"

c2 = "A" && c3 = "B" && kek LIKE "X"


vector<> newBooleanArgs;
if



pk: c1, c2, c3


hashmap<c2, c3>





1. visit to fill hashmaps (each hashmap is conjunct)
2. remove hashmap that cannot be converted to points (or ranges in future)
3. visit to replace filters that was converted to hashmaps

server/connector/data_source.cpp
```diff
@@ -172,7 +247,49 @@ void RocksDBDataSource::InitIterators(CreateFn&& create_iter) {
   }
 }
 
-std::optional<velox::RowVectorPtr> RocksDBDataSource::next(
+velox::RowVectorPtr RocksDBFullScanDataSource::ApplyRemainingFilter(
```
MBkkt: 1. _remainingFilterExprSet exists => create batch without mangled names => when filter => when remapping to mangled names and necessary columns
2. _remainingFilterExprSet not exists => create batch with mangled names

server/connector/data_source.cpp
```diff
@@ -352,4 +488,82 @@ uint64_t RocksDBDataSource::IterateColumn(rocksdb::Iterator& it,
...
+      key_utils::AppendTableKey(key, _object_key);
+      key_utils::AppendColumnKey(key, _column_ids[col_idx]);
+      primary_key::Create(*_values, _offset + point_idx, key);
+    }
+  }
+
+  _key_slices.resize(total_keys);
+  for (size_t i = 0; i < total_keys; ++i) {
+    _key_slices[i] = _keys[i];
+  }
```
MBkkt: we want to sort whole keys.
also we don't want rocksdb to sort our batches

tests/sqllogic/recovery/faults.test
mkornaukhov: yep

server/connector/serenedb_connector.hpp
mkornaukhov: fixed

server/connector/data_source.cpp
mkornaukhov: Did it. 
1) Points are sorted as velox values
2) Column iteration is according to column_id
So totally we have sorted keys

server/connector/rocksdb_filter.hpp
mkornaukhov: Did like this.
But list is a little bit misleading. If after step 1 we have any non-specific hashmap, it means that point lookup is not suitable, so step 3 is not going to be executed, but all other things are same as described.

server/connector/data_source.cpp
mkornaukhov: Now it's become possible, implemented as in full scan

server/connector/data_source.cpp
mkornaukhov: This looks a little overcomplicated, in point 1 we might need to do 2: mangled in output type -> unmangled -> mangle again. For now I just just references inside remaining filter so that if some field is present in output type, than use this mangled reference

server/connector/data_source.hpp
mkornaukhov: Which constant do you mean?

Dronplane: static constexpr size_t kBatchSize = 32;

mkornaukhov: I allocate keys for all points inside point lookup datasource, but 32 is constant only for batching rocksdb calls.

server/connector/data_source.cpp
```diff
@@ -31,43 +32,100 @@
...
+    velox::Type::create<Kind>(), num_points, &memory_pool);
+
+  for (size_t point_idx = 0; point_idx < num_points; ++point_idx) {
+    const size_t idx = sort_rank * num_points + point_idx;
+    const auto& status = statuses[idx];
+    const auto& value = values[idx];
+
+    if (status.IsNotFound()) {
+      // Filtered point is not found, it's OK, just skip
+      continue;
```
Dronplane, line 83: Could we just skip incrementing idx in vector for writing in this case? Then we will not need wrapping in dictionary for result vector. We write only exisiting values sequentially. At the end  we will check found count and if it is lower than expected just "resize" our vector to cut off tail.

server/connector/data_source.hpp
Dronplane: I see. What bothers me in this approach is that we store&allocate whole read  batch values/statuses but MultiGet still works by `kBatchSize` window. So it could be processed and put to result vector in same small window without additional allocations.

server/connector/data_source.cpp
```diff
@@ -352,4 +408,228 @@ uint64_t RocksDBDataSource::IterateColumn(rocksdb::Iterator& it,
...
+
+    _produced += found_count;
+    FinalizeOffset(batch_size, total_points);
+
+    return velox::BaseVector::create<velox::RowVector>(_read_type, found_count,
+                                                       &_memory_pool);
+  }
+
+  BuildKeys(batch_size, num_columns);
+  PerformMultiGet(batch_size * num_columns);
```
Dronplane, line 580: As an optimization idea: Can we first read single column and filter out non-existing points? And read from other columns only existing ones. As we always write row in one-or-nothing  manner - it can be safely assumed that if key is missing from first column it would be missing from all others.

server/connector/data_source.cpp
```diff
@@ -352,4 +408,228 @@ uint64_t RocksDBDataSource::IterateColumn(rocksdb::Iterator& it,
...
+
+    for (size_t i = 1; i < batch_size; ++i) {
+      SDB_ASSERT(_keys[i] > _keys[i - 1], "Failed on ", i);
+    }
+
+    PerformMultiGet(batch_size);
+    const size_t found_count = CountFound(batch_size);
+
+    // Ensure statuses
+    if (auto it = absl::c_find_if(_statuses,
```
Dronplane: Could we move this check to CountFound? We anyway throw on read errors. Just that we do same check in ReadPointColumnValues in case of actual columns read. I suggets to consolidate this check in a single place. As a bonus we will avoid second loop over results.

mkornaukhov: CountFound is used before reading column (num_columns > 0 case), there is a status check, too.
I thought about this, but I think we do not want to check 2 times statuses

Dronplane: We should remove error status check ReadPointColumnValues in this case of course. That is what I meant as "consolidation". As if there is at least one error status it makes all read doomed and  we can bail out without even starting ReadPointColumnValues

mkornaukhov: Fixed

server/connector/data_source.cpp
mkornaukhov: Added todo

server/connector/data_source.cpp
mkornaukhov: Fixed

server/connector/data_source.cpp
```diff
@@ -352,4 +406,205 @@ uint64_t RocksDBDataSource::IterateColumn(rocksdb::Iterator& it,
...
+  BuildKeys(batch_size, num_columns);
+  PerformMultiGet(batch_size * num_columns);
+  CheckAndCountFound(batch_size * num_columns);
+
+  std::vector<velox::VectorPtr> columns;
+  columns.reserve(num_columns);
+  for (size_t column_idx = 0; column_idx < num_columns; ++column_idx) {
+    const auto& type = _read_type->childAt(column_idx);
+    columns.emplace_back(VELOX_DYNAMIC_SCALAR_TYPE_DISPATCH(
+      ReadPointColumnValues, type->kind(), _col_rank[column_idx], batch_size,
```
Dronplane, line 590: why not use here result of CheckAndCountFound to allocate resulting vectors? That will remove need of resizing vector at the end and we will not over-allocate at the beginning.

mkornaukhov: fixed

server/connector/serenedb_connector.hpp
```diff
@@ -706,9 +829,22 @@ class SereneDBConnector final : public velox::connector::Connector {
...
-    return std::make_unique<RocksDBSnapshotDataSource>(
-      *connector_query_ctx->memoryPool(), _db, _cf, output_type, column_oids,
-      serene_table_handle.GetEffectiveColumnId(), object_key, snapshot);
+    const auto& points = serene_table_handle.GetPoints();
+    if (!points.empty()) {
+      return std::make_unique<RocksDBSnapshotPointLookupDataSource>(
+        *connector_query_ctx->memoryPool(), _db, _cf, read_type, column_oids,
+        object_key, snapshot,
+        PointsToRowVector(points, serene_table_handle.GetPKType(),
+                          connector_query_ctx->memoryPool()),
```
MBkkt, line 838: 1. I doubt we want points to be velox vector
2. If it's really needed it can be created single time, and shared across different datasources

server/connector/serenedb_connector.hpp
```diff
@@ -657,11 +704,77 @@ class SereneDBConnector final : public velox::connector::Connector {
...
+        name_mapping.emplace(
+          std::string{original},
+          std::make_shared<velox::core::FieldAccessTypedExpr>(
+            output_type->childAt(i), mangled));
+      }
+
+      // Detect filter-only columns: referenced in filter but absent from
+      // output. Extend column_oids and read_type so ApplyRemainingFilter can
+      // read them.
+      absl::flat_hash_set<std::string_view> filter_field_names;
```
MBkkt, line 732: containers::FlatHashSet

server/connector/serenedb_connector.hpp
```diff
@@ -212,6 +255,10 @@ class SereneDBConnectorTableHandle final
   query::Transaction& _transaction;
   irs::Filter::Query::ptr _search_query;
   ObjectId _index_id = ObjectId::none();
+  velox::RowTypePtr _pk_type;
+  std::vector<Point> _points;
+  velox::core::TypedExprPtr _remaining_filter;
+  absl::flat_hash_map<std::string, FilterColumn> _table_column_map;
```
MBkkt, line 261: containers::FlatHashMap

server/connector/serenedb_connector.hpp
```diff
@@ -177,14 +192,31 @@ class SereneDBColumnHandle final : public velox::connector::ColumnHandle {
...
 
+  std::string toString() const final {
+    // TODO(mkornaukhov) implement filter printing
+    if (_search_query) {
+      return absl::StrCat(_name, ", scan=search");
+    }
+    if (!_points.empty()) {
+      return absl::StrCat(_name, ", scan=point");
+    }
+    return absl::StrCat(_name, ", scan=full");
```
MBkkt, line 217:
```suggestion
    if (!_points.empty()) {
      return absl::StrCat(_name, ", type=rocksdb_point_loopup");
    }
    return absl::StrCat(_name, ", type=rocksdb_full_scan");
```
+ something special for file scan?

server/connector/serenedb_connector.hpp
```diff
@@ -177,14 +192,31 @@ class SereneDBColumnHandle final : public velox::connector::ColumnHandle {
...
+    velox::core::TypedExprPtr remaining_filter);
 
   bool supportsIndexLookup() const final { return false; }
 
   const std::string& name() const final { return _name; }
 
+  std::string toString() const final {
+    // TODO(mkornaukhov) implement filter printing
+    if (_search_query) {
+      return absl::StrCat(_name, ", scan=search");
```
MBkkt, line 212:
```suggestion
      return absl::StrCat(_name, ", type=search_lookup");
```

server/connector/serenedb_connector.cpp
```diff
@@ -135,11 +155,36 @@ SereneDBTableLayout::createTableHandle(
...
+
+    if (!res.points.empty()) {
+      points = std::move(res.points);
+      SortPoints(points, *pk_type);
+      remaining_filter = std::move(res.remaining_filter);
+    }
+  }
+
+  if (!kExecuteFiltersInTableScan && remaining_filter) {
+    rejected_filters = {std::move(remaining_filter)};
```
MBkkt, line 180: if (remaining_filter is AND) {
   rejected_filters.push_back(each child of remaining_filter)
}

server/connector/rocksdb_filter.hpp
```diff
@@ -0,0 +1,100 @@
...
+
+  // Returns nullopt when the two points are contradictory (e.g. a=1 AND a=2).
+  [[nodiscard]] static std::optional<Point> Intersect(const Point& lhs,
+                                                      const Point& rhs);
+
+ private:
+  std::span<const std::string> _pk_names;
+  absl::flat_hash_map<std::string, velox::core::ConstantTypedExprPtr>
+    _column_filters;
+  absl::flat_hash_set<const velox::core::ITypedExpr*> _source_exprs;
```
MBkkt, line 73: containers::FlatHashMap/Set

server/connector/rocksdb_filter.cpp
```diff
@@ -0,0 +1,348 @@
...
+
+  return {std::move(pts), RewriteExpr(expr, sources)};
+}
+
+void SortPoints(std::vector<Point>& points, const velox::RowType& pk_type) {
+  absl::c_sort(points, [&](const Point& lhs, const Point& rhs) {
+    for (std::string_view col_name : pk_type.names()) {
+      const auto* lhs_filter = lhs.FindFilter(col_name);
+      const auto* rhs_filter = rhs.FindFilter(col_name);
+      SDB_ASSERT(lhs_filter && rhs_filter);
```
MBkkt, line 338:
```suggestion
      SDB_ASSERT(lhs_filter);
      SDB_ASSERT(rhs_filter);
```

server/connector/primary_key.hpp
```diff
@@ -47,6 +47,8 @@ void Create(const velox::RowVector& data,
 void Create(const velox::RowVector& data, velox::vector_size_t idx,
             std::string& key);
 
+void AppendKeyValue(std::string& key, const velox::BaseVector& column,
+                    velox::vector_size_t idx);
 template<typename T>
```
MBkkt, line 52:
```suggestion
                    velox::vector_size_t idx);

template<typename T>
```

server/connector/data_source.hpp
```diff
@@ -92,40 +199,177 @@ class RocksDBDataSource : public velox::connector::DataSource {
...
+                                   std::move(values),
+                                   output_column_count,
+                                   std::move(remaining_filter),
+                                   transaction.GetSnapshot(),
+                                   evaluator,
+                                   transaction} {}
+};
+
+class RocksDBSnapshotPointLookupDataSource final
+  : public RocksDBPointLookupDataSource<rocksdb::DB> {
```
MBkkt, line 353: why rocksdb::DB? I thought it should be something like snapshot

MBkkt, line 353: using?

server/connector/data_source.hpp
```diff
@@ -92,40 +199,177 @@ class RocksDBDataSource : public velox::connector::DataSource {
...
+                                velox::core::ExpressionEvaluator* evaluator);
 
   void addSplit(std::shared_ptr<velox::connector::ConnectorSplit> split) final;
 
  private:
   rocksdb::Transaction& _transaction;
 };
 
-class RocksDBSnapshotDataSource final : public RocksDBDataSource {
+class RocksDBSnapshotFullScanDataSource : public RocksDBFullScanDataSource {
```
MBkkt, line 223: Make it using  + common template?

server/connector/data_source.cpp
```diff
@@ -352,4 +406,206 @@ uint64_t RocksDBDataSource::IterateColumn(rocksdb::Iterator& it,
...
+    const size_t found_count = CheckAndCountFound(batch_size);
+    _produced += found_count;
+    FinalizeOffset(batch_size);
+
+    return velox::BaseVector::create<velox::RowVector>(_read_type, found_count,
+                                                       &_memory_pool);
+  }
+
+  // TODO(mkornaukhov) try to load only the first column and get mask of which
+  // rows are present. This may allow us to reduce number of multiget calls
```
MBkkt, line 578: I think it can be done via old batching approach, because first column end is special case

server/connector/rocksdb_filter.hpp
MBkkt, line 73: This is ok for temporary struct, but I doubt we want to keep all these maps during whole query execution

server/connector/rocksdb_filter.cpp
```diff
@@ -0,0 +1,348 @@
...
+    const auto* func_call = expr->asUnchecked<velox::core::CallTypedExpr>();
+    if (func_call->name() == "presto_eq") {
+      pts = ExtractFilterEq(func_call, pk_names);
+    } else if (func_call->name() == "in") {
+      pts = ExtractFilterIn(func_call, pk_names);
+    } else if (func_call->name() == "and") {
+      pts = ExtractFilterAnd(func_call, pk_names);
+    } else if (func_call->name() == "or") {
+      pts = ExtractFilterOr(func_call, pk_names);
+    } else {
```
MBkkt, line 309: you need to handle "not" and "presto_ne", pushdown "not" to "and"/"or"/"not"/"presto_ne"

Also instead of literals you should use velox literals and for something like presto_eq use Dronplane approach from search:
name == "eq" or ends_with("_eq")

server/connector/rocksdb_filter.cpp
```diff
@@ -0,0 +1,348 @@
...
+ExtractAndRewriteResult ExtractAndRewriteFilterExpr(
+  const velox::core::TypedExprPtr& expr,
+  std::span<const std::string> pk_names) {
+  auto pts = ExtractFilterExpr(expr, pk_names);
+  if (!absl::c_all_of(pts, [](const Point& p) { return p.IsSpecific(); })) {
+    return {{}, expr};
+  }
+
+  // Collect the unique source sub-expressions to get rid of them.
+  absl::flat_hash_set<const velox::core::ITypedExpr*> sources;
```
MBkkt, line 325: containers::

server/connector/serenedb_connector.hpp
```diff
@@ -212,6 +255,10 @@ class SereneDBConnectorTableHandle final
   query::Transaction& _transaction;
   irs::Filter::Query::ptr _search_query;
   ObjectId _index_id = ObjectId::none();
+  velox::RowTypePtr _pk_type;
+  std::vector<Point> _points;
```
MBkkt, line 259: We want to store something that can be shared between multiple data sources.

It can be velox vector but something simpler will be nice

server/connector/data_source.hpp
mkornaukhov: there is api like `options.snapshot = ...; db.get(options, ...);`, so `get()/multiget()` are called on rocksdb::DB

mkornaukhov: clarify please

server/connector/serenedb_connector.hpp
mkornaukhov: > something special for file scan?

It has special connector table handle, @ pashandor789 implemented special this method for it.

PR #280, opened 2026-03-06
feat: Add boost query

MBkkt:
It's mostly used by lucene parser: for queries like `+open some stuff` for such queries we need to return all docs that has open, but for score also "boost" docs that have `some` or `stuff`


pashandor789 (approved):
Thanks ! (c) a great guy

PR #283, opened 2026-03-06
feat: Add tests for S3, options framework

pashandor789:


tests/sqllogic/sdb/pg/index/basic.test
```diff
@@ -18,7 +14,7 @@ statement ok
 INSERT INTO t VALUES ('a'), ('b'), ('c')
 
 
-query ok
+query ??
```
MBkkt: shouldn't be like this?

server/pg/commands/create_function.cpp
```diff
@@ -47,94 +48,104 @@ namespace sdb::pg {
 
 namespace {
 
+using namespace std::string_view_literals;
+
+inline constexpr EnumOptionInfo<catalog::FunctionLanguage> kLanguage{
```
MBkkt: inline constexpr not in header doesn't make sense

PR #287, opened 2026-03-06
feat: ADD create table as select

pashandor789:


server/catalog/local_catalog.cpp
```diff
@@ -1520,6 +1531,29 @@ Result LocalCatalog::DropTable(ObjectId db_id, std::string_view schema_name,
...
+  }
+  auto table_id =
+    _snapshot->GetObjectId<ResolveType::Relation>(*schema_id, name);
+  if (!table_id) {
+    return Result{ERROR_SERVER_ILLEGAL_NAME};
+  }
+  auto object = _snapshot->GetObject(*table_id);
+  if (object) {
+    auto& schema_obj = basics::downCast<SchemaObject>(*object);
+    schema_obj.SetTombstoned(false);
```
MBkkt, line 1551: this write should be atomic?

server/catalog/object.h
```diff
@@ -168,6 +168,9 @@ class SchemaObject : public DatabaseObject {
   // TODO(gnusi): remove it after schema management is done
   void SetSchemaId(ObjectId schema_id) noexcept { _schema_id = schema_id; }
 
+  bool Tombstoned() const noexcept { return _tombstoned; }
```
MBkkt, line 171: this old should be atomic?

PR #298, opened 2026-03-11
Add TERM_xx functions for SEARCH

Dronplane:
Added TERM_.. functions to explicitly filter indexed strings by verbatim terms. This allows querying inverted index with any tokenizer used. It is up to user to provide particular filter terms.
Search function prototype tightened to move validation logic to the SQL parser/analyzer whenever possible.

### DIFF
--- a/server/search/functions.hpp
+++ b/server/search/functions.hpp
@@ -24,14 +24,14 @@
 
 namespace sdb::search::functions {
 
-constexpr inline std::string_view kPhrase = "sdb_phrase";
-constexpr inline std::string_view kTermEq = "sdb_term_eq";
-constexpr inline std::string_view kTermLt = "sdb_term_lt";
-constexpr inline std::string_view kTermLe = "sdb_term_lte";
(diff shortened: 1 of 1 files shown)


server/search/functions.hpp
```diff
@@ -25,6 +25,13 @@
 namespace sdb::search::functions {
 
 constexpr inline std::string_view kPhrase = "sdb_phrase";
+constexpr inline std::string_view kTermEq = "sdb_term_eq";
+constexpr inline std::string_view kTermLt = "sdb_term_lt";
+constexpr inline std::string_view kTermLe = "sdb_term_lte";
+constexpr inline std::string_view kTermGe = "sdb_term_gte";
+constexpr inline std::string_view kTermGt = "sdb_term_gt";
+constexpr inline std::string_view kTermIn = "sdb_term_in";
+constexpr inline std::string_view kTermLike = "sdb_term_like";
```
MBkkt: inline constexpr same as static constexpr

PR #334, opened 2026-03-14
perf: Try 256 elements per block, based on gnusi changes

MBkkt:
It's quite same, maybe a little slower.

I think larger block is generally good, but the decompression itself is probably slower, because worse delta (prefix sum) compute


third_party/simdcomp/include/avxintegratedbitpacking.h
```diff
@@ -0,0 +1,64 @@
...
+inline __m256i avx_delta(__m256i curr, __m256i prev) {
+  __m256i shifted = _mm256_slli_si256(curr, 4);
+  const __m256i elem3 = _mm256_shuffle_epi32(curr, _MM_SHUFFLE(3, 3, 3, 3));
+  const __m256i cross = _mm256_permute2x128_si256(elem3, elem3, 0x08);
+  shifted = _mm256_blend_epi32(shifted, cross, 0x10);
+  const __m256i prev_bcast = _mm256_set1_epi32(_mm256_extract_epi32(prev, 7));
+  shifted = _mm256_blend_epi32(shifted, prev_bcast, 0x01);
+
+  return _mm256_sub_epi32(curr, shifted);
+}
```
MBkkt, line 43: something like this also don't help

```inline __m256i avx_delta(__m256i curr, __m256i prev) {
  __m256i left_shift = _mm256_slli_si256(curr, 4);
  __m256i curr_shift = _mm256_srli_si256(curr, 12);
  __m256i curr_right_only =
    _mm256_permute2f128_si256(curr_shift, curr_shift, 8);
  __m256i prev_shift = _mm256_srli_si256(prev, 12);
  __m256i sub_left =
    _mm256_permute2f128_si256(prev_shift, prev_shift, 3 | (8 << 4));
  __m256i diff =
    _mm256_or_si256(left_shift, _mm256_or_si256(curr_right_only, sub_left));
  return _mm256_sub_epi32(curr, diff);
}
```

PR #365, opened 2026-03-18
Add text dictionary definition for inverted index

Dronplane:
VARCHAR fields should now have opclass defining text dictionary.
Dictionary has indexes as dependencies so it can not be dropped while used except during cascade drop of entire schema.
Opclass options yet not supported.
Default text dictionary yet not supported.


libs/iresearch/include/iresearch/analysis/analyzers.cpp
```diff
@@ -204,7 +204,9 @@ bool MakeAnalyzer(vpack::Slice input, Analyzer::ptr& output) {
     return false;
   }
   if (type == StringTokenizer::type_name()) {
-    output = {};
+    // TODO(Dronplane): maybe we want to handle this case outside? But returning
+    // true + nullptr looks not very robust.
+    output = std::make_unique<StringTokenizer>();
     return true;
   }
```
MBkkt, line 211: It's identity case? Before this we had global pool of such objects

server/catalog/catalog.h
```diff
@@ -172,7 +174,7 @@ struct LogicalCatalog {
     std::shared_ptr<catalog::Tokenizer> tokenizer) = 0;
   virtual ResultOr<std::shared_ptr<Index>> RegisterIndex(
     ObjectId database_id, ObjectId schema_id, ObjectId id, ObjectId relation_id,
-    IndexBaseOptions options) = 0;
+    IndexBaseOptions options, vpack::Slice impl_options) = 0;
```
MBkkt: It's not ok, we should have vpack here, instead struct without vpack should be used I think

Dronplane: you mean "should not"?
The issue is that we do not have common structure for concrete index options. 
So "impl" properties are parsed after index type is determined. I was thinking of having some base class for all concrete index options and have parsing outside. But that would be anyway some function in the index.cpp
So I think it would be simple just to pass down this slice to the MakeIndex and let indexes deal with it.

libs/iresearch/include/iresearch/analysis/analyzers.cpp
Dronplane: Yes, there was some code that handle this special case (e.g. true + null). I thing it would be more robust to have Identity just as regular Tokenizer (which has a pool inside)

tests/sqllogic/sdb/pg/index/term_ops.test
```diff
@@ -50,6 +59,9 @@ SELECT a FROM test_term_ops_idx WHERE TERM_EQ(c, 'd');
 a
 4
 
+query error
+SELECT a FROM test_term_ops_idx c = 'd';
```
MBkkt: error text?

tests/server/connector/search_test_utils.cpp
```diff
@@ -0,0 +1,41 @@
...
+#include <iresearch/formats/formats.hpp>
+#include <iresearch/search/scorers.hpp>
+
+namespace sdb::connector::test {
+
+void RegisterSearchEntities() {
+  static bool gInited = false;
+
+  if (!gInited) {
+    irs::analysis::analyzers::Init();
```
MBkkt, line 33: move to test main? in such case no need gInited

server/pg/sql_analyzer_velox.cpp
```diff
@@ -2565,6 +2565,10 @@ void SqlAnalyzer::ProcessIndexStmt(State& state, const IndexStmt& stmt) {
   FillColumnsInfo(table_state, pk, table_type, column_names, column_exprs);
 
   VisitNodes(stmt.indexParams, [&](const IndexElem& index_elem) {
+    if (!index_elem.name) {
```
MBkkt, line 2568: when this is possible?

server/connector/search_sink_writer.hpp
```diff
@@ -78,8 +78,8 @@ class SearchSinkInsertBaseImpl : public ColumnSinkWriterImplBase {
 
     irs::Tokenizer& GetTokens() const noexcept {
       SDB_ASSERT(analyzer || string_analyzer);
-      SDB_ASSERT((analyzer == nullptr) || (string_analyzer == nullptr));
-      return analyzer ? *analyzer : *string_analyzer;
+      SDB_ASSERT((analyzer == nullptr) || !string_analyzer.has_value());
```
MBkkt: without has_value plz, codestyle

server/connector/search_sink_writer.hpp
```diff
@@ -78,8 +78,8 @@ class SearchSinkInsertBaseImpl : public ColumnSinkWriterImplBase {
...
+      return analyzer ? *analyzer : *string_analyzer.value().get();
```
MBkkt: **string_abalyzer?

libs/iresearch/include/iresearch/analysis/analyzers.cpp
MBkkt: Ok, I think we can improve perf of this later if it will be needed

server/pg/sql_analyzer_velox.cpp
Dronplane: See test:
```
statement error db error: ERROR: Unsupported index column definition
CREATE INDEX index_ts on tstest USING inverted(ts_lexize('test_english', c));
```

tests/server/connector/search_test_utils.cpp
Dronplane: we use gtest_main and do not have own test main implementation right now.
I guess we can make it and put IcuInitializer and RegisterSearchEntities.
But better as a separate small PR just for clarity.

MBkkt: Create an issue/PR please

Dronplane: https://github.com/serenedb/serenedb/issues/371

PR #369, opened 2026-03-19
Fix asan and tsan problems

mkornaukhov:
So after this PR we expect that asan and tsan launches should not emit sanitizer errors. If they do, it means something has broken.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -145,9 +145,6 @@
 [submodule "third_party/aws-sdk-cpp"]
 	path = third_party/aws-sdk-cpp
 	url = git@github.com:serenedb/aws-sdk-cpp.git
-[submodule "third_party/wasm-micro-runtime"]
-	path = third_party/wasm-micro-runtime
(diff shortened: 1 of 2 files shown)


.gitmodules
```diff
@@ -118,42 +106,54 @@
...
+	url = git@github.com:serenedb/aws-checksums.git
+[submodule "third_party/aws-crt-cpp"]
+	path = third_party/aws-crt-cpp
+	url = git@github.com:serenedb/aws-crt-cpp.git
+[submodule "third_party/aws-sdk-cpp"]
+	path = third_party/aws-sdk-cpp
+	url = git@github.com:serenedb/aws-sdk-cpp.git
+[submodule "third_party/wasm-micro-runtime"]
+	path = third_party/wasm-micro-runtime
+	url = git@github.com:serenedb/wasm-micro-runtime.git
```
MBkkt: Can we remove this and also remove from cmake/etc?

PR #402, opened 2026-03-23
fix: Describe the next statement for the simple query

pashandor789:


third_party/sqllogictest-rs
MBkkt, line 1: you break submodule another time

PR #414, opened 2026-03-24
feat: Allow setting log level

gnusi:


server/query/config_variables.cpp
```diff
@@ -28,6 +28,20 @@ namespace sdb {
 constexpr std::pair<std::string_view, VariableDescription>
   kVariableDescription[] = {
     // serenedb specific variables
+    {
+      "sdb_log_level",
```
MBkkt: reuse constexpr constant for this variable

PR #419, opened 2026-03-24
Some improvements of data source

mkornaukhov:
Some improvements of connector and data source
- Print filters for rocksdb and search data sources
- Get rid of extra `O(batch size)` vectors, use constexpr size windows for building output velox vector
- Reduce number of keys for multiget: store mask of present rows after getting values for first column
- Parametrize rocksdb data sources with source: transaction or snapshot (database + snapshot really)

Also fix some follow ups from #267

### DIFF
--- a/server/connector/search_filter_printer.hpp
+++ b/server/connector/search_filter_printer.hpp
@@ -48,8 +48,4 @@ void AbslStringify(Sink& sink, const Filter& filter) {
   sink.Append(ToString(filter));
 }
 
-inline std::ostream& operator<<(std::ostream& os, const Filter& filter) {
-  return os << ToString(filter);
-}
-
 }  // namespace irs


server/connector/search_filter_printer.cpp
```diff
@@ -0,0 +1,357 @@
...
+        break;
+      case sdb::search::mangling::kAnalyzer:
+        mangle_name = "analyzer";
+        break;
+      case sdb::search::mangling::kNested:
+        mangle_name = "nested";
+        break;
+    }
+  }
+  return absl::StrCat("col:", col_id, "(", mangle_name, ")");
```
Dronplane: Could we make a full decoding - e.g. columnid->field name?
If right now it is complicated let's create an issue.

tests/sqllogic/sdb/pg/index/inverted_index.test
```diff
@@ -120,7 +120,7 @@ QUERY PLAN
 EXECUTION PLAN:
 Fragment 0:  numWorkers=0:
 -- Aggregation[1][SINGLE count := presto_count()] -> "count":BIGINT
-  -- TableScan[0][vedernikoff, scan=search] ->
+  -- TableScan[0][vedernikoff, type=search_lookup, filter=AND[PHRASE[col(string) = <Term:pudge(0, 0); >] && PHRASE[col(string) = <Term:game(0, 0); >]]] ->
```
Dronplane: Shouldn't it be something like "AND[PHRASE[col:<slt:ignore>(string) =" ?

server/connector/data_source.cpp
```diff
@@ -543,68 +480,96 @@ std::optional<velox::RowVectorPtr> RocksDBPointLookupDataSource<Source>::next(
...
-  BuildKeys(batch_size, num_columns);
-  PerformMultiGet(batch_size * num_columns);
-  size_t found_count = CheckAndCountFound(batch_size * num_columns);
-  SDB_ASSERT(found_count % num_columns == 0);
-  found_count /= num_columns;
+  // For each column, iterate mask with chunks of
+  // kMultiGetBatchSize, fetching only the found_count present points.
+  // _in_batch_offset is reset per column.
+  std::vector<velox::VectorPtr> columns(num_columns);
+  for (size_t col_idx : _sorted_col_indices) {
```
Dronplane: While building `_present_rows_batch` we have already "fetched" data for `_sorted_col_indices[0]`. And as we already know if we need column data or not we could read "presence" column while filling `_present_rows_batch`  and avoid re-fetching it here again.

tests/sqllogic/sdb/pg/index/inverted_index.test
mkornaukhov: Fixed, `col=<slt:ignore>` is used

server/connector/data_source.cpp
mkornaukhov: fixed

server/connector/search_filter_printer.cpp
mkornaukhov: Yes, fixed

server/connector/search_filter_printer.hpp
```diff
@@ -20,13 +20,36 @@
...
+
 template<typename Sink>
-void AbslStringify(Sink& sink, const Filter& filter);
+void AbslStringify(Sink& sink, const Filter& filter) {
+  sink.Append(ToString(filter));
+}
+
+inline std::ostream& operator<<(std::ostream& os, const Filter& filter) {
+  return os << ToString(filter);
+}
```
MBkkt: Can we remove it?

server/connector/rocksdb_filter.cpp
```diff
@@ -222,6 +206,12 @@ velox::variant ToVariant(const velox::core::ConstantTypedExpr& expr) {
     ExtractScalarVariant, expr.type()->kind(), *expr.valueVector());
 }
 
+bool IsCallOf(const velox::core::CallTypedExpr* call, std::string_view suffix) {
+  SDB_ASSERT(suffix[0] == '_');
+  auto name = suffix.substr(1);
+  return call->name() == name || call->name().ends_with(suffix);
+}
```
MBkkt:
```suggestion
bool IsCallOf(const velox::core::CallTypedExpr* call, std::string_view suffix) {
  SDB_ASSERT(suffix[0] == '_');
  auto expected = suffix.substr(1);
  const auto& actual = call->name();
  return actual == expected || actual.ends_with(suffix);
}
```
Also it will be nice to reuse this code in search filter factory, so it can be some generic helper

mkornaukhov: fixed

server/connector/search_filter_printer.hpp
mkornaukhov: yes, removed

PR #432, opened 2026-03-26
Fix index backfill creation commits

Dronplane:
+ Fixed flushing during index creation.  Intermediate Flush/Commit should not happen inside write batch. As it would leave Document orphaned and lead to concurrent write and commit in the same SegmentWriter. That will end up with various assert firing randomly.

 + Fixed final commit at index backfill finish. It is now synchronous commit operation. Not just scheduling regular commit task. Scheduling regular task does not guarantee full commit. If other background commit is already running task will just re-schedule itself leaving "tail" uncommitted. If query would immediately follow INDEX CREATE operation it might miss tail if snapshot would be taken before scheduled task would have a chance to actually run the commit.

 + Fixed VACUUM parameter check. If index name is specified server would still try cast it to Table and assert out. Now Table getter checks the object type.

### DIFF
--- a/tests/sqllogic/sdb/pg/index/inverted_index.test_slow
+++ b/tests/sqllogic/sdb/pg/index/inverted_index.test_slow
@@ -120,3 +120,67 @@ SELECT * FROM clickclack_pk_facts WHERE PHRASE(detail, 'cutelooked row') ORDER B
 
 statement ok
 DROP TABLE clickclack_pk;
+
+control substitution on
+
+# index on external text/csv table with large dataset
+
+statement count 100000
+CREATE TABLE csv_idx_src AS
+SELECT
+  CASE
+    WHEN vedernikoff_team.crook % 2 = 0 THEN
+      'team member vedernikoff has ' || (crook / 10)::text || ' stocks'
+    ELSE
+      'team member cutelooked has ' || (boss * 1000)::text || ' stocks'
+  END AS fact,
+  CASE
+    WHEN vedernikoff_team.crook % 3 = 0 THEN
+      'department alpha'
+    ELSE
+      'department beta'
+  END AS dept
+FROM
+  generate_series(1, 1000) as vedernikoff_team(crook),
+  generate_series(1, 100) as cutelooked_team(boss);
+
+statement count 100000
+COPY csv_idx_src TO '/tmp/${__DATABASE__}_csv_idx_slow.csv';
+
+statement ok
+CREATE TABLE csv_idx_ext(
+  fact TEXT,
+  dept TEXT
+) USING EXTERNAL
+WITH (
+  PATH = '/tmp/${__DATABASE__}_csv_idx_slow.csv'
+);
+
+statement ok
+CREATE INDEX csv_idx_ext_facts ON csv_idx_ext USING inverted(fact test_english, dept test_english);
+
+query
+SELECT * FROM csv_idx_ext_facts WHERE PHRASE(fact, 'team member');
+----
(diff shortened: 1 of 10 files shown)


tests/sqllogic/sdb/pg/index/basic.test
```diff
@@ -48,9 +48,11 @@ c
 statement ok
 VACUUM (UPDATE_INDEXES) t;
 
-statement error
+statement error db error: ERROR: relation 'wrong_t_name' not found.
```
MBkkt: multi line errors please

tests/sqllogic/sdb/pg/index/inverted_index.test_slow
```diff
@@ -48,6 +48,20 @@ SELECT * FROM clickclack_facts WHERE PHRASE(fact, 'has 0 stocks');
 ----
 400 values hashing to 281bddebb99a13518781132c8d56e194
 
+statement ok
+SET sdb_fault_segment_1000_docs_max TO DEFAULT;
```
MBkkt, line 52: this makes test not able to run in parallel, right? it probably should have something like index name in fault name, and have unique index name

server/pg/command_executor.cpp
MBkkt, line 1: it doesn't look correct, it shouldn't be blocking

server/connector/search_sink_writer.hpp
```diff
@@ -300,9 +305,11 @@ class SearchSinkBackfillWriter final : public SinkIndexWriter,
   }
 
  private:
-  void Commit() {
+  void Commit(bool final) {
```
MBkkt: let's don't use final word, it's almost keyword, so it has strange highlighting in GitHub and some other tools

server/catalog/local_catalog.cpp
```diff
@@ -498,7 +498,7 @@ class SnapshotImpl : public Snapshot {
   std::shared_ptr<Table> GetTable(ObjectId db_id, std::string_view schema,
                                   std::string_view table) const final {
     auto rel = GetRelation(db_id, schema, table);
-    if (!rel) {
+    if (!rel || rel->GetType() != ObjectType::Table) {
```
MBkkt, line 501: What is the test for this?

tests/sqllogic/sdb/pg/index/inverted_index.test_slow
Dronplane: It should not break any other tests. The worse case for now is that some indexes created in the same time as this failpoint is active would operate a bit slower in case there would be more than 1000 docs in test.
I'd leave it as is and not introduce additional complexity.

server/catalog/local_catalog.cpp
Dronplane:
```
statement error db error: ERROR: relation 'index_t' not found.
VACUUM (UPDATE_INDEXES) index_t;
```
see basic.test in same commit as this change

server/pg/command_executor.cpp
Dronplane: I see. Let's make CommitTask with forced wait then

server/catalog/local_catalog.cpp
MBkkt: Maybe better error for this ?

Dronplane: There could be any "object" name here. Existing or not. Of course we can first check if it is "table", if not lookup if this name present in catalog anywhere and specify exactly that user misspelled the name but rather named object of other type. And if not there tell the user that such entity is not exists.  But not sure if it would be better then just tell user " there is no relation with name X "  as it is now.

PR #433, opened 2026-03-26
Add vector distances

codeworse:
Like in pgvector: https://github.com/pgvector/pgvector?tab=readme-ov-file#querying
added distances:
* `<->` - L2
* `<=>` - Cosine
* `<+>` - L1
* `<#>` - inner product

Add ARRAY as column type


tests/sqllogic/sdb/pg/simple/vector.test
```diff
@@ -0,0 +1,44 @@
+query
```
pashandor789, line 1: not enough vedernikoffs in the test

server/connector/data_source.cpp
```diff
@@ -320,6 +326,153 @@ velox::VectorPtr RocksDBFullScanDataSource<Source>::ReadUnknownColumn(
...
+  const auto vector_size = IterateColumn(
+    it, max_size,
+    [&](uint64_t value_idx, std::string_view key, std::string_view value) {
+      if (value.empty()) {
+        // NULL array
+        if (!null_buf) {
+          null_buf = velox::allocateNulls(max_size, &_memory_pool);
+        }
+        velox::bits::clearBit(null_buf->asMutable<uint64_t>(), value_idx);
+        raw_offsets[value_idx] = elem_offset;
```
Dronplane: I guess for "Nulls" it is fine to leave offsets/sizes undefined

tests/sqllogic/sdb/pg/simple/vector.test
codeworse: added tests with vedernikoff and pablo

pashandor789: where .
also why did u put that bad emoji ? 
i's offended by it

server/connector/data_source.cpp
```diff
@@ -320,6 +326,153 @@ velox::VectorPtr RocksDBFullScanDataSource<Source>::ReadUnknownColumn(
...
+        (flags & ValueFlags::HaveLength) != ValueFlags::None;
+
+      uint32_t length_array_size = 0;
+      if (have_length) {
+        length_array_size = irs::vread<uint32_t>(ptr);
+      }
+
+      if (is_constant) {
+        // Remaining bytes are the single constant value (or none for null).
+        const auto remaining_size = static_cast<size_t>(
```
Dronplane, line 398: Please add an assertion that `have_nulls` flag is not set. With scalar values it should be never set. But for more complex cases when constant is wrapped in the dictionary (that brings in some nulls) which is part of the structure that lies in the array - it could be set to optimize storage and avoid expanding constant into flat vector with nulls at data store.

server/connector/data_source.cpp
```diff
@@ -320,6 +326,153 @@ velox::VectorPtr RocksDBFullScanDataSource<Source>::ReadUnknownColumn(
...
+        } else if constexpr (!velox::TypeTraits<ElemKind>::isFixedWidth) {
+          // Variable-length elements (e.g., VARCHAR).
+          // Layout: [length_array: length_array_size bytes][string data]
+          const uint8_t* lptr = ptr;
+          ptr += length_array_size;
+          const uint8_t* data_ptr = ptr;
+
+          for (uint32_t i = 0; i < elem_count; i++) {
+            const uint32_t len = irs::vread<uint32_t>(lptr);
+            if (elem_nulls && !velox::bits::isBitSet(elem_nulls, i)) {
```
Dronplane: could we make elem_nulls check constexpr via ResolveBool ?

tests/sqllogic/sdb/pg/simple/vector.test
```diff
@@ -0,0 +1,150 @@
...
+query error
+SELECT ARRAY[] <+> ARRAY[];
+
+
+query error
+SELECT ARRAY[] <#> ARRAY[];
+
+
+statement ok
+CREATE TABLE t(name TEXT, emb FLOAT[3]);
```
Dronplane: Please add test cases with NULLs

server/connector/data_source.cpp
codeworse: but it would be practically the same, but with more complex code

Dronplane: I mean put entire loop over elem_count into lambda. So we will have branchless execution for cases without nulls. And check bits only when necessary.

tests/sqllogic/sdb/pg/simple/vector.test
codeworse: added into array.test

server/connector/data_source.cpp
codeworse: added

server/connector/data_source.cpp
codeworse: added assertion

server/connector/data_source.cpp
codeworse: erased

tests/sqllogic/sdb/pg/simple/vector.test
codeworse: more tests after: tests/sqllogic/sdb/pg/simple/vector.test:89
P.S. sorry, bro, no offense, that won't happen again

server/connector/data_source.cpp
```diff
@@ -320,6 +326,161 @@ velox::VectorPtr RocksDBFullScanDataSource<Source>::ReadUnknownColumn(
...
+
+  const auto vector_size = IterateColumn(
+    it, max_size,
+    [&](uint64_t value_idx, std::string_view, std::string_view value) {
+      if (value.empty()) {
+        // NULL array
+        if (!null_buf) {
+          null_buf = velox::allocateNulls(max_size, &_memory_pool);
+        }
+        velox::bits::clearBit(null_buf->asMutable<uint64_t>(), value_idx);
```
Dronplane: NIT  - that would be more clear for the reader
```suggestion
        velox::bits::setNull(null_buf->asMutable<uint64_t>(), value_idx);
```

server/pg/sql_analyzer_velox.cpp
```diff
@@ -4501,6 +4501,11 @@ const OpToFuncMap kOpToFunc{
   {"is", "presto_is"},
   {"IS DISTINCT FROM", "presto_distinct_from"},
   {"count_star", "presto_count"},
+  // https://github.com/pgvector/pgvector?tab=readme-ov-file#querying
+  {"<->", "presto_l2_squared"},
```
MBkkt, line 4505: I think velox (faiss) functions works slower than our, let's create an issue to benchmark

third_party/faiss
MBkkt, line 1: let's disable code with #if 0

third_party/velox
MBkkt, line 1: I didn't look into others, but l1 is ridiculously slow

server/pg/sql_analyzer_velox.cpp
```diff
@@ -5230,7 +5235,9 @@ lp::WindowExprPtr SqlAnalyzer::MaybeWindowFuncCall(
 lp::ExprPtr SqlAnalyzer::ProcessColumnRef(State& state, const ColumnRef& expr) {
   std::string name = NameToStr(expr.fields);
 
-  SDB_ASSERT(state.root);
+  if (!state.root) {
+    ErrorColumnDoesNotExist(name, ExprLocation(&expr));
```
MBkkt, line 5239: test for this?

codeworse: in vector.test:
```sql
statement error db error: ERROR: column "pablo" does not exist
INSERT INTO t VALUES ("pablo", ARRAY[6.5, 1.101]), ("vedernikoff", ARRAY[1.002, 8.098]);
```

PR #465, opened 2026-03-30
Range data source

mkornaukhov:
## Range Scan Support for Primary Key Lookups
Extends the connector layer with the ability to push range predicates on primary key columns down into RocksDB, avoiding full table scans for queries like WHERE a > 5 AND a < 10 or WHERE a = 1 AND b >= 3.

## Filter extraction (rocksdb_filter)
The filter extraction machinery was significantly expanded. Previously it only recognised equality point lookups `(a = 1, a IN (1,2,3))`. It now also extracts single-sided and two-sided range constraints (`<, <=, >, >=, !=, BETWEEN`) and organises them into a `ConstraintKind::Ranges` result alongside the existing `ConstraintKind::Points` path. N-dimensional sweep line algorithm were used. Extracted constraints are represented as `ResolvedRange` — an equality prefix over the leading PK columns plus a ColumnRange on the next column.

## Range data source (data_source)
A new `RocksDBPrefixRangeLookupDataSource` was added to execute range scans. It uses one RocksDB iterator per column, each positioned and bounded to cover exactly the extracted ranges. The data source hierarchy was also refactored: shared iterator-reading machinery was extracted into a new abstract `RocksDBPerColumnIteratorDataSource` base, from which both the existing `RocksDBFullScanDataSource` and the new range source inherit.

## Connector wiring (serenedb_connector)
`SereneDBTableLayout::createTableHandle `now distinguishes three cases: point lookup (unchanged), range scan (new), and full scan (fallback). When the filter extraction yields ranges, a SereneDBConnectorTableHandle carrying the resolved ranges is created and routed to the range data source.

## Follow up
Utilize new data source for secondary index

### DIFF
--- a/server/connector/data_source.hpp
+++ b/server/connector/data_source.hpp
@@ -272,7 +272,7 @@ class PointLookupSKColumnBuilder {
   }
 
   velox::RowVectorPtr Finish(size_t found_count) {
-    return _materializer.ReadRows(_row_keys, nullptr);
+    return _materializer.ReadRows(_row_keys, nullptr, {});
   }
 
   const irs::bitset& PresentRows() const { return _dummy; }
--- a/server/connector/secondary_data_source.hpp
+++ b/server/connector/secondary_data_source.hpp
@@ -90,7 +90,7 @@ class SecondaryIndexDataSource final : public velox::connector::DataSource {
     }
 
     _produced += _row_keys.size();
-    return _materializer.ReadRows(_row_keys, nullptr);
+    return _materializer.ReadRows(_row_keys, nullptr, {});
   }
 
   void addDynamicFilter(velox::column_index_t,
@@ -308,7 +308,7 @@ class SecondaryIndexFullScanDataSource final
     }
 
     _produced += _row_keys.size();
-    return _materializer.ReadRows(_row_keys, nullptr);
+    return _materializer.ReadRows(_row_keys, nullptr, {});
   }
 
   void addDynamicFilter(velox::column_index_t,
--- a/server/connector/serenedb_connector.hpp
+++ b/server/connector/serenedb_connector.hpp
@@ -536,7 +536,8 @@ class IndexTable : public axiom::connector::Table {
   CopyColumns(const axiom::connector::Table& table,
               const catalog::Index& index) {
     std::vector<std::unique_ptr<const axiom::connector::Column>> columns;
-    columns.reserve(table.allColumns().size() + 1);
+    columns.reserve(table.allColumns().size() + index.GetColumnIds().size() +
+                    1);
     for (const auto* col : table.allColumns()) {
       auto* sdb_col = basics::downCast<const SereneDBColumn>(col);
       columns.push_back(std::make_unique<SereneDBColumn>(
@@ -549,6 +550,18 @@ class IndexTable : public axiom::connector::Table {
         catalog::Column::GenerateScoreName(table.type()->names());
       columns.push_back(std::make_unique<SereneDBColumn>(
         score_name, velox::REAL(), catalog::Column::kInvertedIndexScoreId));
+      // Add one virtual offsets column per requested index column.
+      // It is only included in the output type when a corresponding offsets
+      // requested. We can not emit here only requested as index could be
+      // referenced multiple times and first creation of IndexTable object could
+      // be without offsets. So we must ensure we always have such columns for
+      // every indexed column.
+      auto offsets_type = catalog::Column::MakeOffsetsType();
+      for (const auto& column_id : index.GetColumnIds()) {
+        columns.push_back(std::make_unique<SereneDBColumn>(
+          catalog::Column::MakeOffsetsName(column_id), offsets_type,
+          catalog::Column::kInvertedIndexOffsetsId));
+      }
     }
     return columns;
   }
@@ -571,6 +584,7 @@ class IndexTable : public axiom::connector::Table {
   const auto& GetIndex() const noexcept { return _index; }
   const auto& GetTable() const noexcept { return _table; }
   auto& GetTransaction() const noexcept { return _transaction; }
+  const auto& GetOffsetsRequests() const noexcept { return _offsets_requests; }
 
   const std::vector<const axiom::connector::TableLayout*>& layouts()
     const final {
@@ -589,6 +603,13 @@ class IndexTable : public axiom::connector::Table {
     _scorer = std::move(scorer);
   }
 
+  void SetOffsets(std::vector<catalog::Column::OffsetsFieldRequest>&& offsets) {
+    SDB_ASSERT(_index.GetType() == catalog::ObjectType::InvertedIndex);
+    _offsets_requests.insert(_offsets_requests.end(),
+                             std::make_move_iterator(offsets.begin()),
+                             std::make_move_iterator(offsets.end()));
+  }
+
   const std::shared_ptr<const irs::Scorer>& GetScorerPtr() const noexcept {
     SDB_ASSERT(_index.GetType() == catalog::ObjectType::InvertedIndex);
     return _scorer;
@@ -598,6 +619,7 @@ class IndexTable : public axiom::connector::Table {
   query::Transaction& _transaction;
   axiom::connector::TablePtr _table;
   const catalog::Index& _index;
+  std::vector<catalog::Column::OffsetsFieldRequest> _offsets_requests;
   std::vector<std::unique_ptr<SereneDBTableLayout>> _layout_handles;
   std::vector<const axiom::connector::TableLayout*> _layouts;
   std::shared_ptr<const irs::Scorer> _scorer;
@@ -620,6 +642,7 @@ class InvertedIndexTableHandle final
       _search_query{std::move(search_query)},
       _scorer{std::move(scorer)},
       _underlying_table{table.GetTable()},
+      _offsets_requests{table.GetOffsetsRequests()},
       _search_filter_str{std::move(search_filter_str)} {
     const auto& column_map = table.columnMap();
     _effective_column_id = ComputeEffectiveColumnId(column_map);
@@ -661,6 +684,8 @@ class InvertedIndexTableHandle final
     return *_underlying_table;
   }
 
+  const auto& GetOffsetsRequests() const noexcept { return _offsets_requests; }
+
  private:
   std::string _name;
   ObjectId _table_id;
@@ -670,6 +695,7 @@ class InvertedIndexTableHandle final
   irs::Filter::Query::ptr _search_query;
   std::shared_ptr<const irs::Scorer> _scorer;
   axiom::connector::TablePtr _underlying_table;
+  std::vector<catalog::Column::OffsetsFieldRequest> _offsets_requests;
   containers::FlatHashMap<std::string, FilterColumn> _table_column_map;
   std::string _search_filter_str;
 };
@@ -1537,18 +1563,47 @@ class SereneDBConnector final : public velox::connector::Connector {
     const auto& search_snapshot =
       transaction.EnsureSearchSnapshot(handle.GetIndexId());
 
-    // TODO: teach ScanSpec to skip the score column instead of stripping it
-    // from the type.
+    // TODO: teach ScanSpec to skip the score/offsets columns instead of
+    // stripping them from the type.
     std::vector<std::string> reader_names;
     std::vector<velox::TypePtr> reader_types;
     for (size_t i = 0; i < output_type->size(); ++i) {
-      if (column_oids[i] != catalog::Column::kInvertedIndexScoreId) {
+      if (column_oids[i] != catalog::Column::kInvertedIndexScoreId &&
+          column_oids[i] != catalog::Column::kInvertedIndexOffsetsId) {
         reader_names.push_back(output_type->nameOf(i));
         reader_types.push_back(output_type->childAt(i));
       }
     }
     auto reader_type =
       velox::ROW(std::move(reader_names), std::move(reader_types));
+
+    // Select only the offsets requests relevant to this DataSource, in the
+    // order they appear in output_type. The accumulated requests may contain
+    // entries from other sub-queries that reference the same index; column_name
+    // uniquely identifies each request across all sub-queries.
+    std::vector<catalog::Column::OffsetsFieldRequest> offsets;
+    const auto& all_offsets = handle.GetOffsetsRequests();
+    for (size_t i = 0; i < output_type->size(); ++i) {
+      if (column_oids[i] != catalog::Column::kInvertedIndexOffsetsId) {
+        continue;
+      }
+      const auto& col_name = output_type->nameOf(i);
+      auto it =
+        std::find_if(all_offsets.begin(), all_offsets.end(),
+                     [&](const catalog::Column::OffsetsFieldRequest& r) {
+                       return r.column_name == col_name;
+                     });
+      SDB_ENSURE(it != all_offsets.end(), ERROR_INTERNAL,
+                 "CreateSearchInvertedIndexDataSource: offsets request not "
+                 "found for column ",
+                 col_name);
+      SDB_ASSERT(
+        std::find_if(std::next(it), all_offsets.end(),
+                     [&](const catalog::Column::OffsetsFieldRequest& r) {
+                       return r.column_name == col_name;
+                     }) == all_offsets.end());
+      offsets.push_back(*it);
+    }
 
     return CreateWithMaterializer(
       handle, output_type, reader_type, column_oids, column_handles, pool,
@@ -1558,7 +1613,7 @@ class SereneDBConnector final : public velox::connector::Connector {
         using Mat = decltype(mat);
         return std::make_unique<SearchDataSource<Mat>>(
           pool, std::move(mat), search_snapshot.reader, handle.GetSearchQuery(),
-          handle.GetScorer());
+          handle.GetScorer(), std::move(offsets));
       });
   }
 
--- a/.github/workflows/build-manual.yml
+++ b/.github/workflows/build-manual.yml
@@ -42,8 +42,12 @@ on:
         default: '{}'
 
 permissions:
-  statuses: write
+  checks: write
   contents: write
+
+concurrency:
+  group: build-manual-${{ github.ref }}
+  cancel-in-progress: true
 
 jobs:
   test-unit:
@@ -129,22 +133,25 @@ jobs:
           RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
         run: ./scripts/send_notification.bash
 
-      - name: Update PR commit status
-        if: contains(inputs.BUILD_CONFIG, 'PR_SHA')
+      - name: Update PR check run
+        if: contains(inputs.BUILD_CONFIG, 'CHECK_RUN_ID')
         uses: actions/github-script@v8
         with:
           script: |
             const config = JSON.parse(process.env.BUILD_CONFIG);
-            const state = process.env.JOB_STATUS === 'success' ? 'success' : 'failure';
+            const conclusion = process.env.JOB_STATUS === 'success' ? 'success' : 'failure';
             const runUrl = `${context.serverUrl}/${context.repo.owner}/${context.repo.repo}/actions/runs/${context.runId}`;
-            await github.rest.repos.createCommitStatus({
+            await github.rest.checks.update({
               owner: context.repo.owner,
               repo: context.repo.repo,
-              sha: config.PR_SHA,
-              state: state,
-              description: state === 'success' ? 'All checks passed' : 'Some checks failed',
-              context: config.PR_STATUS_CONTEXT || 'serenedb | build & test',
-              target_url: runUrl,
+              check_run_id: config.CHECK_RUN_ID,
+              status: 'completed',
+              conclusion: conclusion,
+              details_url: runUrl,
+              output: {
+                title: conclusion === 'success' ? 'All checks passed' : 'Some checks failed',
+                summary: `[View Run](${runUrl})`,
+              },
             });
         env:
           BUILD_CONFIG: ${{ inputs.BUILD_CONFIG }}
--- a/.github/workflows/pr-build-trigger.yml
+++ b/.github/workflows/pr-build-trigger.yml
@@ -69,7 +69,7 @@ jobs:
       actions: write
       contents: read
       pull-requests: write
-      statuses: write
+      checks: write
     steps:
       - name: Get PR info
         id: pr-info
@@ -104,14 +104,34 @@ jobs:
             const enableMsan = isChecked('MSAN');
 
             const { owner, repo } = context.repo;
-            const statusContext = 'serenedb | build & test';
+            const checkName = 'serenedb | build & test';
 
-            await github.rest.repos.createCommitStatus({
-              owner, repo, sha,
-              state: 'pending',
-              description: 'Build queued',
-              context: statusContext,
+            const { data: check } = await github.rest.checks.create({
+              owner, repo,
+              head_sha: sha,
+              name: checkName,
+              status: 'in_progress',
+              output: { title: 'Build queued', summary: 'Waiting for workflow to start...' },
             });
+
+            // Cancel in-progress runs on this branch
+            const { data: runs } = await github.rest.actions.listWorkflowRuns({
+              owner, repo, workflow_id: 'build-manual.yml', branch,
+              status: 'in_progress', per_page: 10,
+            });
+            for (const run of runs.workflow_runs) {
+              core.info(`Cancelling run #${run.id} (${run.status})`);
+              await github.rest.actions.cancelWorkflowRun({ owner, repo, run_id: run.id }).catch(() => {});
+            }
+            // Also cancel queued/waiting runs
+            const { data: queued } = await github.rest.actions.listWorkflowRuns({
+              owner, repo, workflow_id: 'build-manual.yml', branch,
+              status: 'queued', per_page: 10,
+            });
+            for (const run of queued.workflow_runs) {
+              core.info(`Cancelling queued run #${run.id}`);
+              await github.rest.actions.cancelWorkflowRun({ owner, repo, run_id: run.id }).catch(() => {});
+            }
 
             // Snapshot latest run ID before dispatch
             const { data: before } = await github.rest.actions.listWorkflowRuns({
@@ -132,7 +152,7 @@ jobs:
                 BUILD_CONFIG: JSON.stringify({
                   TRIGGERED_BY: context.actor,
                   PR_SHA: sha,
-                  PR_STATUS_CONTEXT: statusContext,
+                  CHECK_RUN_ID: check.id,
                 }),
               },
             });
@@ -151,6 +171,15 @@ jobs:
               }
             }
 
+            // Update check run with the Actions run URL
+            await github.rest.checks.update({
+              owner, repo,
+              check_run_id: check.id,
+              status: 'in_progress',
+              details_url: runUrl,
+              output: { title: 'Build in progress', summary: `[View Run](${runUrl})` },
+            });
+
             const opts = [
               enableAsan && 'asan',
               enableTsan && 'tsan',
@@ -167,15 +196,15 @@ jobs:
             // Results are stored as JSON in a hidden HTML comment for reliable parsing
             const dataRe = /<!-- ci-data:(.*?) -->/;
             const dataMatch = updatedBody.match(dataRe);
-            const runs = dataMatch ? JSON.parse(dataMatch[1]) : [];
-            runs.push({ opts: optsStr, url: runUrl, time: timestamp });
+            const history = dataMatch ? JSON.parse(dataMatch[1]) : [];
+            history.push({ opts: optsStr, url: runUrl, time: timestamp });
 
             // Build visible output
-            const latest = runs[runs.length - 1];
+            const latest = history[history.length - 1];
             let resultsBlock = `\n\n**Latest:** ${latest.opts} · [View Run](${latest.url}) · ${latest.time}`;
 
-            if (runs.length > 1) {
-              const oldRuns = runs.slice(0, -1).reverse().map(r =>
+            if (history.length > 1) {
+              const oldRuns = history.slice(0, -1).reverse().map(r =>
                 `| ${r.opts} | [View Run](${r.url}) | ${r.time} |`
               );
               resultsBlock += [
@@ -193,7 +222,7 @@ jobs:
 
             // Replace or append results section
             const sectionRe = /\n\n<!-- ci-data:.*? -->[\s\S]*$/;
-            const newData = `\n\n<!-- ci-data:${JSON.stringify(runs)} -->${resultsBlock}`;
+            const newData = `\n\n<!-- ci-data:${JSON.stringify(history)} -->${resultsBlock}`;
             if (sectionRe.test(updatedBody)) {
               updatedBody = updatedBody.replace(sectionRe, newData);
             } else {
--- a/libs/app/options/parameters.h
+++ b/libs/app/options/parameters.h
@@ -217,7 +217,7 @@ struct NumericParameter : public Parameter {
     } else if constexpr (std::is_same_v<ValueType, double>) {
       return "double";
     } else {
-      static_assert("unsupported ValueType");
+      static_assert(false, "unsupported ValueType");
     }
   }
 
--- a/libs/iresearch/include/iresearch/search/phrase_iterator.hpp
+++ b/libs/iresearch/include/iresearch/search/phrase_iterator.hpp
@@ -406,7 +406,7 @@ class FixedPhraseFrequency {
   }
 
   IRS_FORCE_INLINE uint32_t NextPosition() {
-    if constexpr (HasIntervals) {
+    if constexpr (HasIntervals || Offs) {
       return NextPositionGeneric();
     } else {
       return NextPositionOptimized();
--- a/libs/iresearch/include/iresearch/search/phrase_query.cpp
+++ b/libs/iresearch/include/iresearch/search/phrase_query.cpp
@@ -182,7 +182,9 @@ DocIterator::ptr FixedPhraseQuery::ExecuteWithOffsets(
       auto back = std::prev(std::end(phrase_state->terms));
 
       while (term_state != back) {
-        if (!add_iterator(kRequiredFeatures)) {
+        // TODO(Dronplane) Make Adapter accept itarators without offs
+        // We will not need ones from middle iterators.
+        if (!add_iterator(kRequireOffs)) {
           return DocIterator::empty();
         }
       }
--- a/server/catalog/table_options.h
+++ b/server/catalog/table_options.h
@@ -99,6 +99,7 @@ struct Column {
     std::numeric_limits<uint64_t>::max() - 1'000'000;
   static constexpr Id kGeneratedPKId = kMaxRealId + 1;
   static constexpr Id kInvertedIndexScoreId = kMaxRealId + 2;
+  static constexpr Id kInvertedIndexOffsetsId = kMaxRealId + 3;
 
   static constexpr std::string_view GetUpdateNamePrefix() {
     static constexpr std::string_view kUpdatePrefix = "upd$";
@@ -137,6 +138,33 @@ struct Column {
 
   static std::string GenerateScoreName(
     std::span<const std::string> column_names);
+
+  // Prefix used in virtual offsets column names. Ends with kReservedSymbol so
+  // it can never collide with a user-defined column name.
+  static constexpr std::string_view kOffsetsNamePrefix =
+    "sdb_inverted_index_offsets$";
+
+  // Encodes the searched column's catalog ID into the virtual column name.
+  static std::string MakeOffsetsName(Id column_id) {
+    static_assert(kOffsetsNamePrefix.ends_with(query::kReservedSymbol));
+    return absl::StrCat(kOffsetsNamePrefix, column_id);
+  }
+
+  // ARRAY(BIGINT) -- flat offsets column: interleaved start,end pairs.
+  static velox::TypePtr MakeOffsetsType() {
+    return velox::ARRAY(velox::BIGINT());
+  }
+
+  // Request for a single OFFSETS() column: which catalog column to extract
+  // offsets for, and how many offset pairs to return per document at most.
+  // column_name is the mangled output column name produced by NextColumnName;
+  // it is globally unique across all sub-queries and used to match this request
+  // to the correct DataSource during SearchDataSource construction.
+  struct OffsetsFieldRequest {
+    Id column_id;
+    size_t limit = pg::Objects::kDefaultOffsetsLimit;
+    std::string column_name;
+  };
 
   Id id;
   velox::TypePtr type;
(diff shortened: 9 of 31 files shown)


server/connector/serenedb_connector.hpp
```diff
@@ -319,7 +323,8 @@ class SereneDBConnectorTableHandle final
   catalog::Column::Id _effective_column_id;
   query::Transaction& _transaction;
   velox::RowTypePtr _pk_type;
-  std::vector<SpecificPoint> _points;
+  std::vector<ResolvedPoint> _points;
+  std::vector<ResolvedRange> _ranges;
```
MBkkt, line 327: I don't really like old name and can say same about new.
But I think ok

server/connector/serenedb_connector.hpp
```diff
@@ -1111,6 +1116,16 @@ class SereneDBConnector final : public velox::connector::Connector {
           PointLookupPKColumnBuilder{});
       }
 
+      const auto& ranges_ryow = serene_table_handle.GetRanges();
+      if (!ranges_ryow.empty()) {
```
```suggestion
     
      if ( const auto& ranges = serene_table_handle.GetRanges(); !ranges.empty()) {
```

server/connector/serenedb_connector.hpp
```diff
@@ -1128,6 +1143,15 @@ class SereneDBConnector final : public velox::connector::Connector {
         PointLookupPKColumnBuilder{});
     }
 
+    const auto& ranges = serene_table_handle.GetRanges();
+    if (!ranges.empty()) {
```
```suggestion
    
    if (const auto& ranges = serene_table_handle.GetRanges(); !ranges.empty()) {
```

server/connector/serenedb_connector.hpp
```diff
@@ -740,7 +745,7 @@ class SecondaryIndexTableHandle final
   query::Transaction& _transaction;
   catalog::Column::Id _effective_column_id;
   ObjectId _shard_id;
-  std::vector<SpecificPoint> _points;
+  std::vector<ResolvedPoint> _points;
```
MBkkt, line 774: Also should be ranges? Or in separate PR, will you add this?

server/connector/serenedb_connector.cpp
```diff
@@ -372,6 +382,38 @@ std::string SereneDBConnectorTableHandle::toString() const {
...
+        }
+        absl::StrAppend(&format_str, names[prefix_size], "=",
+                        sr.range_column.toString());
+      }
+      return absl::StrCat("{", format_str, "}");
+    };
+
+    std::vector<std::string> parts;
+    for (const auto& sr : _ranges) {
+      if (!sr.IsContradictory()) {
```
MBkkt: really?
```suggestion
      if (!sr.IsConflicting()) {
```

server/connector/rocksdb_filter.hpp
```diff
@@ -39,60 +39,292 @@
...
+// A possibly-bounded interval on a single column value.
+struct ColumnRange {
+  static constexpr uint8_t kZero = 0x0;
+  static constexpr uint8_t kLeftBounded = 0x1;
+  static constexpr uint8_t kLeftInclusive = 0x2;
+  static constexpr uint8_t kRightBounded = 0x4;
+  static constexpr uint8_t kRightInclusive = 0x8;
+  // Contradictory: the range is empty (no value satisfies it, e.g. a>5 AND
+  // a<3). A contradictory ColumnRange is disjoint from every other range.
+  static constexpr uint8_t kContradictory = 0x10;
```
MBkkt: Conflicting or just Empty, it's way more intuitive.

Also why do we need to keep such ranges? Make comment about why or remove

server/connector/rocksdb_filter.hpp
```diff
@@ -39,60 +39,292 @@
 
 namespace sdb::connector {
 
-// A point in PK-space. Absent column in _column_filters means "any value".
-class Point {
+enum class ComparisonOp { None, Lt, Le, Gt, Ge };
+
+// A possibly-bounded interval on a single column value.
+struct ColumnRange {
+  static constexpr uint8_t kZero = 0x0;
```
MBkkt: why?

server/connector/rocksdb_filter.hpp
MBkkt: Maybe enum class/enum?

server/connector/rocksdb_filter.hpp
```diff
@@ -39,60 +39,292 @@
...
  private:
+  explicit KeyBounds(std::span<const std::string> pk_names)
+    : _pk_names{pk_names} {}
+
   std::span<const std::string> _pk_names;
-  containers::FlatHashMap<std::string, velox::core::ConstantTypedExprPtr>
-    _column_filters;
-  containers::FlatHashSet<const velox::core::ITypedExpr*> _source_exprs;
+  bool _contradictory =
+    false;  // true -> contradictory predicate, matches no rows
```
MBkkt:
```suggestion
  // true -> contradictory predicate, matches no rows
    bool _contradictory =
    false;
```

server/connector/secondary_data_source.hpp
MBkkt, line 1: should support ranges too

server/connector/rocksdb_filter.hpp
```diff
@@ -39,60 +39,292 @@
...
+  const velox::core::TypedExprPtr& expr, std::span<const std::string> pk_names,
+  bool negated = false);
+
+enum class ConstraintKind {
+  // All constraints are fully-specified equality points; use point lookup.
+  Points,
+  // At least one constraint is a range; use range scan on the range prefix.
+  Ranges,
+  // No constraints, use full scan.
+  None
```
MBkkt:
```suggestion
  None,
```

server/connector/rocksdb_filter.hpp
```diff
@@ -102,8 +334,8 @@ struct ExtractAndRewriteResult {
   std::span<const std::string> column_names);
 
 // Sorts and deduplicates points in-place by key order. Column order matches
-// the pk_type used during ToSpecificPoints.
-void SortAndDedupPoints(std::vector<SpecificPoint>& points);
+// the pk_type used during ToResolvedPoints.
```
MBkkt, line 319: do we still have ToResolvedPoints?

server/connector/rocksdb_filter.hpp
```diff
@@ -39,60 +39,292 @@
...
+      return IsContradictory() && !other.IsContradictory();
+    }
+    const size_t min_depth = std::min(prefix.size(), other.prefix.size());
+    for (size_t i = 0; i < min_depth; ++i) {
+      if (prefix[i] < other.prefix[i]) {
+        return true;
+      }
+      if (other.prefix[i] < prefix[i]) {
+        return false;
+      }
```
MBkkt: For future, there's good rule about how to write `operator<`, for simple types
It's more clear and simple
```suggestion
      if (prefix[i] != other.prefix[i]) {
        return prefix[i] < other.prefix[i];
      }
```

But this and your code have drawback, it's slower than needed.
Should be one compare/<=> call. Lets add them in our Velox, create an issue please

server/connector/rocksdb_filter.hpp
```diff
@@ -39,60 +39,292 @@
...
+
+    if (prefix.size() == other.prefix.size()) {
+      return range_column.LeftBoundLessThan(other.range_column);
+    }
+
+    // Unequal prefix depths with a matching common prefix.
+    // At the first diverging position, one side has an exact prefix value and
+    // the other has its range_col starting there.
+    const bool range_is_this = prefix.size() < other.prefix.size();
+    const velox::variant& point =
```
MBkkt:
```suggestion
    const auto& point =
```

server/connector/rocksdb_filter.hpp
```diff
@@ -39,60 +39,292 @@
...
+      range_is_this ? other.prefix[prefix.size()] : prefix[other.prefix.size()];
+    const ColumnRange& column_range =
```
MBkkt:
```suggestion
    const auto& column_range =
```

server/connector/rocksdb_filter.hpp
```diff
@@ -39,60 +39,292 @@
...
+      range_is_this ? range_column : other.range_column;
+
+    if (!column_range.HasLeft() || column_range.LeftValue() < point) {
+      return range_is_this;
+    }
+    if (point < column_range.LeftValue()) {
+      return !range_is_this;
+    }
+    return !range_is_this && !column_range.IsLeftInclusive();
```
MBkkt: looks quite fragile, maybe improve it with separate of cases?

server/connector/data_source.hpp
```diff
@@ -359,15 +375,50 @@ class RocksDBPointLookupDataSource : public RocksDBBaseDataSource {
...
+    velox::memory::MemoryPool& memory_pool, Source& source,
+    rocksdb::ColumnFamilyHandle& cf, velox::RowTypePtr read_type,
+    std::vector<catalog::Column::Id> column_ids,
+    catalog::Column::Id effective_column_id, ObjectId object_key,
+    size_t output_column_count, const rocksdb::Snapshot* snapshot,
+    std::vector<ResolvedRange> ranges, velox::RowTypePtr pk_type,
+    velox::core::TypedExprPtr remaining_filter = nullptr,
+    velox::core::ExpressionEvaluator* evaluator = nullptr);
+
+  // NOLINTNEXTLINE(readability-identifier-naming)
```
MBkkt, line 395: can we remove such things?
```suggestion
```

server/connector/rocksdb_filter.cpp
```diff
@@ -21,74 +21,231 @@
...
+
+velox::variant ToVariant(const velox::core::ConstantTypedExpr& expr) {
+  if (!expr.hasValueVector()) {
+    return expr.value();
+  }
+  if (expr.isNull()) {
+    return velox::variant(expr.type()->kind());
+  }
+  return VELOX_DYNAMIC_SCALAR_TYPE_DISPATCH(
+    ExtractScalarVariant, expr.type()->kind(), *expr.valueVector());
```
MBkkt: I thought they have something like this

server/connector/rocksdb_filter.cpp
```diff
@@ -21,74 +21,231 @@
...
 namespace sdb::connector {
 namespace {
 
-std::vector<Point> AnyPoint(const std::span<const std::string> names) {
-  return {Point(names)};
+template<velox::TypeKind Kind>
+velox::variant ExtractScalarVariant(const velox::BaseVector& vec) {
+  using T = typename velox::TypeTraits<Kind>::NativeType;
+  const auto& cv = static_cast<const velox::ConstantVector<T>&>(vec);
+  return velox::variant(cv.valueAt(0));
```
MBkkt: basics::downCast + assert on size

server/connector/rocksdb_filter.cpp
```diff
@@ -21,74 +21,231 @@
...
+      return ComparisonOp::Ge;
+    case ComparisonOp::Le:
+      return ComparisonOp::Gt;
+    case ComparisonOp::None:
+      return ComparisonOp::None;
+  }
+  SDB_UNREACHABLE();
+}
+
+// Helper: produces two KCs for a != v -> a < v OR a > v.
```
MBkkt: two what?

PS. after more reading it's probably key constraints, but still from code that return KeyBounds looks strange

server/connector/rocksdb_filter.cpp
```diff
@@ -197,152 +429,822 @@ velox::core::TypedExprPtr RewriteExpr(
...
   }
 
-  return {std::move(pts), RewriteExpr(expr, sources)};
+  velox::core::TypedExprPtr remaining = RewriteExpr(expr, sources);
+  return {ConstraintKind::Ranges, std::move(constraints), std::move(remaining)};
 }
-
-void SortAndDedupPoints(std::vector<SpecificPoint>& points) {
-  absl::c_sort(points, [](const SpecificPoint& lhs, const SpecificPoint& rhs) {
+void SortAndDedupPoints(std::vector<ResolvedPoint>& points) {
```
MBkkt, line 1217: blank lines between large functions implementation is a common sense

server/connector/rocksdb_filter.cpp
```diff
@@ -197,152 +429,822 @@ velox::core::TypedExprPtr RewriteExpr(
...
+    // (e.g. OR(b<2, b>2) instead of (a<2 AND b<2) OR (a>2 AND b>2)).
+    // Keep the original expression as remaining_filter in this case.
+    const bool has_beyond_prefix_constraints = absl::c_any_of(
+      pk_names.subspan(prefix), [&](const std::string& column_name) {
+        return contraint.FindColumnRange(column_name) != nullptr;
+      });
+    if (!has_beyond_prefix_constraints) {
+      for (size_t i = 0; i < prefix; ++i) {
+        const auto& col_exprs = contraint.GetSourceExprs(pk_names[i]);
+        sources.insert(col_exprs.begin(), col_exprs.end());
```
MBkkt, line 1208: insert_range, also why not move?

PS. I see why not move, they're just pointers. Maybe pointers on pointers?

server/connector/rocksdb_filter.cpp
```diff
@@ -197,152 +429,822 @@ velox::core::TypedExprPtr RewriteExpr(
...
+      }
+    }
   }
 
-  return {std::move(pts), RewriteExpr(expr, sources)};
+  velox::core::TypedExprPtr remaining = RewriteExpr(expr, sources);
```
MBkkt:
```suggestion
  auto remaining = RewriteExpr(expr, sources);
```

server/connector/rocksdb_filter.cpp
```diff
@@ -197,152 +429,822 @@ velox::core::TypedExprPtr RewriteExpr(
...
+      return {ConstraintKind::None, {}, expr};
+    }
+  }
+
+  // Collect rewrite sources from every column in each constraint's prefix
+  // (all K equality columns + the range column), not just the first PK column.
   containers::FlatHashSet<const velox::core::ITypedExpr*> sources;
-  for (const auto& p : pts) {
-    sources.insert(p.GetSourceExprs().begin(), p.GetSourceExprs().end());
+  for (const auto& contraint : constraints) {
```
MBkkt: typo: should be constraint

server/connector/rocksdb_filter.cpp
```diff
@@ -197,152 +429,822 @@ velox::core::TypedExprPtr RewriteExpr(
...
+    const size_t prefix = contraint.RangePrefixSize();
```
MBkkt:
```suggestion
    const auto prefix = contraint.RangePrefixSize();
```

server/connector/rocksdb_filter.cpp
```diff
@@ -197,152 +429,822 @@ velox::core::TypedExprPtr RewriteExpr(
...
+
+    // Check if any constrains columns beyond its range prefix. If so, the
+    // scan covers a superset of the intended rows, and the beyond-prefix
+    // predicates must be post-filtered.  Because different KCs may have
```
MBkkt:
```suggestion
    // predicates must be post-filtered. Because different KCs may have
```

server/connector/rocksdb_filter.cpp
```diff
@@ -197,152 +429,822 @@ velox::core::TypedExprPtr RewriteExpr(
...
+    // different beyond-prefix conditions, we cannot form a single shared
+    // remaining filter by stripping only the prefix sources -- the stripped
+    // expression would combine conditions from different ranges incorrectly
+    // (e.g. OR(b<2, b>2) instead of (a<2 AND b<2) OR (a>2 AND b>2)).
+    // Keep the original expression as remaining_filter in this case.
```
MBkkt: the main issue with such comments they can lie/became lie. I'm not sure we should keep them. 

I just don't like this and above comment.

server/connector/rocksdb_filter.cpp
```diff
@@ -197,152 +429,822 @@ velox::core::TypedExprPtr RewriteExpr(
...
+        contiguous = false;
+      }
+    }
+    if (contiguous) {
+      return {ConstraintKind::None, {}, expr};
+    }
+  }
+
+  // Collect rewrite sources from every column in each constraint's prefix
+  // (all K equality columns + the range column), not just the first PK column.
```
MBkkt:
```
// Collect source expressions from prefix columns so they can be stripped
// from the remaining filter (the scan will enforce them).
// If a constraint also covers suffix PK columns, skip it -- stripping
// prefix sources would break correlations between prefix and suffix
// conditions, e.g. OR(not_pk<2, not_pk>2) instead of (pk<2 AND not_pk<2) OR (pk>2 AND not_pk>2))
```

server/connector/rocksdb_filter.cpp
MBkkt: I think wording is bad.

server/connector/rocksdb_filter.cpp
```diff
@@ -197,152 +429,822 @@ velox::core::TypedExprPtr RewriteExpr(
...
+
+    // Check if any constrains columns beyond its range prefix. If so, the
+    // scan covers a superset of the intended rows, and the beyond-prefix
+    // predicates must be post-filtered.  Because different KCs may have
+    // different beyond-prefix conditions, we cannot form a single shared
+    // remaining filter by stripping only the prefix sources -- the stripped
+    // expression would combine conditions from different ranges incorrectly
+    // (e.g. OR(b<2, b>2) instead of (a<2 AND b<2) OR (a>2 AND b>2)).
+    // Keep the original expression as remaining_filter in this case.
+    const bool has_beyond_prefix_constraints = absl::c_any_of(
```
MBkkt: has_suffix_constraints/on_suffix

server/connector/rocksdb_filter.cpp
```diff
@@ -197,152 +429,822 @@ velox::core::TypedExprPtr RewriteExpr(
...
+      !first_column_ranges.back() || !first_column_ranges.back()->HasRight();
+    bool contiguous = first_unbounded && last_unbounded;
+    for (size_t i = 1; contiguous && i < first_column_ranges.size(); ++i) {
+      const ColumnRange* prev = first_column_ranges[i - 1];
+      const ColumnRange* curr = first_column_ranges[i];
+      // prev must have a right bound and curr a left bound with the same
+      // value; together they must not leave the shared point uncovered
+      // (i.e. not both exclusive).
+      if (!prev || !prev->HasRight() || !curr || !curr->HasLeft() ||
+          !(prev->RightValue() == curr->LeftValue()) ||
```
MBkkt:
```suggestion
          prev->RightValue() != curr->LeftValue() ||
```

server/connector/rocksdb_filter.cpp
```diff
@@ -197,152 +429,822 @@ velox::core::TypedExprPtr RewriteExpr(
...
+    }
+
+    const bool first_unbounded =
+      !first_column_ranges.front() || !first_column_ranges.front()->HasLeft();
+    const bool last_unbounded =
+      !first_column_ranges.back() || !first_column_ranges.back()->HasRight();
+    bool contiguous = first_unbounded && last_unbounded;
+    for (size_t i = 1; contiguous && i < first_column_ranges.size(); ++i) {
+      const ColumnRange* prev = first_column_ranges[i - 1];
+      const ColumnRange* curr = first_column_ranges[i];
```
MBkkt:
```suggestion
      const auto* prev = first_column_ranges[i - 1];
      const auto* curr = first_column_ranges[i];
```

server/connector/rocksdb_filter.cpp
```diff
@@ -197,152 +429,822 @@ velox::core::TypedExprPtr RewriteExpr(
...
+  }
+
+  // Normalize: the sweep may produce multiple KCs that share the same scan
+  // prefix (equality columns 0..K-1 plus the range column K) but differ only
+  // in trailing columns beyond K.  All such KCs map to the identical RocksDB
+  // range scan; keep one representative and let remaining_filter handle the
+  // rest.  Two KCs are "scan-equivalent" iff their RangePrefixSize() agrees
+  // and the ColumnRange for every column in 0..prefix_size-1 is equal.
+  {
+    static const ColumnRange kUnconstr{};
```
MBkkt: constexpr?
```suggestion
    static const ColumnRange kUnconstr;
```
is it needed to be static if not possible constexpr

MBkkt: bad name

server/connector/rocksdb_filter.cpp
```diff
@@ -197,152 +429,822 @@ velox::core::TypedExprPtr RewriteExpr(
...
+    auto scan_equivalent = [&](const KeyBounds& a, const KeyBounds& b) {
+      const size_t prefix_size = a.RangePrefixSize();
+      if (prefix_size != b.RangePrefixSize()) {
+        return false;
+      }
+      for (size_t i = 0; i < prefix_size; ++i) {
+        const ColumnRange* col_a_ptr = a.FindColumnRange(pk_names[i]);
+        const ColumnRange* col_b_ptr = b.FindColumnRange(pk_names[i]);
+        const ColumnRange& col_a = col_a_ptr ? *col_a_ptr : kUnconstr;
+        const ColumnRange& col_b = col_b_ptr ? *col_b_ptr : kUnconstr;
```
MBkkt:
```suggestion
        const auto* col_a_ptr = a.FindColumnRange(pk_names[i]);
        const auto* col_b_ptr = b.FindColumnRange(pk_names[i]);
        const auto& col_a = col_a_ptr ? *col_a_ptr : kUnconstr;
        const auto& col_b = col_b_ptr ? *col_b_ptr : kUnconstr;
```

server/connector/rocksdb_filter.cpp
```diff
@@ -197,152 +429,822 @@ velox::core::TypedExprPtr RewriteExpr(
...
+        if (!(col_a == col_b)) {
```
MBkkt: !=, C++ compiler autogenerate them

server/connector/rocksdb_filter.cpp
```diff
@@ -197,152 +429,822 @@ velox::core::TypedExprPtr RewriteExpr(
...
+
+  // Normalize: the sweep may produce multiple KCs that share the same scan
+  // prefix (equality columns 0..K-1 plus the range column K) but differ only
+  // in trailing columns beyond K.  All such KCs map to the identical RocksDB
+  // range scan; keep one representative and let remaining_filter handle the
+  // rest.  Two KCs are "scan-equivalent" iff their RangePrefixSize() agrees
+  // and the ColumnRange for every column in 0..prefix_size-1 is equal.
+  {
+    static const ColumnRange kUnconstr{};
+    auto scan_equivalent = [&](const KeyBounds& a, const KeyBounds& b) {
```
MBkkt: prefer `l` and `r` instead of random `a`/`b`

server/connector/rocksdb_filter.cpp
```diff
@@ -197,152 +429,822 @@ velox::core::TypedExprPtr RewriteExpr(
...
+    for (const auto& point : constraints) {
+      for (const auto& [column_name, source_exprs] : point.GetSourceExprs()) {
+        sources.insert(source_exprs.begin(), source_exprs.end());
+      }
+    }
+    return {ConstraintKind::Points, std::move(constraints),
+            RewriteExpr(expr, sources)};
+  }
+
+  // Normalize: the sweep may produce multiple KCs that share the same scan
```
MBkkt: I think KC is key constraint? Maybe just "constraints"

server/connector/rocksdb_filter.cpp
```diff
@@ -197,152 +429,822 @@ velox::core::TypedExprPtr RewriteExpr(
...
+  // prefix (equality columns 0..K-1 plus the range column K) but differ only
+  // in trailing columns beyond K.  All such KCs map to the identical RocksDB
```
MBkkt:
```suggestion
  // in trailing columns beyond K. All such KCs map to the identical RocksDB
```
llms really bad at this, replace this in other places too

server/connector/rocksdb_filter.cpp
```diff
@@ -197,152 +429,822 @@ velox::core::TypedExprPtr RewriteExpr(
...
+  // range scan; keep one representative and let remaining_filter handle the
+  // rest.  Two KCs are "scan-equivalent" iff their RangePrefixSize() agrees
+  // and the ColumnRange for every column in 0..prefix_size-1 is equal.
+  {
```
MBkkt: // Normalize: multiple constraints may share the same prefix ranges but
// differ in suffix PK columns. They produce identical RocksDB range scans,
// so keep one representative and let remaining_filter handle the rest.
// Two constraints are scan-equivalent iff their prefix size matches and
// every column range in the prefix is equal.

server/connector/rocksdb_filter.cpp
```diff
@@ -197,152 +429,822 @@ velox::core::TypedExprPtr RewriteExpr(
...
+      std::unique(constraints.begin(), constraints.end(), scan_equivalent),
+      constraints.end());
+  }
+
+  // If the constraints' first-column ranges together cover (-inf, +inf) with
+  // no gap, the range scan reads every row -- equivalent to a full scan.
+  // Detect this by checking that sorted disjoint ranges tile the whole axis:
+  // first has no left bound, last has no right bound, and every consecutive
+  // pair is adjacent (right of prev and left of next share a value with
+  // complementary or equal-inclusive endpoints, leaving no uncovered point).
```
MBkkt: this is written in the code

server/connector/rocksdb_filter.cpp
```diff
@@ -197,152 +429,822 @@ velox::core::TypedExprPtr RewriteExpr(
...
+      }
+      return true;
+    };
+    constraints.erase(
+      std::unique(constraints.begin(), constraints.end(), scan_equivalent),
+      constraints.end());
+  }
+
+  // If the constraints' first-column ranges together cover (-inf, +inf) with
+  // no gap, the range scan reads every row -- equivalent to a full scan.
```
MBkkt: // If the prefix ranges together cover (-inf, +inf) with no gaps,
// the scan reads every row -- just do a full scan instead.

server/connector/rocksdb_filter.cpp
```diff
@@ -197,152 +429,822 @@ velox::core::TypedExprPtr RewriteExpr(
...
+      !first_column_ranges.front() || !first_column_ranges.front()->HasLeft();
+    const bool last_unbounded =
+      !first_column_ranges.back() || !first_column_ranges.back()->HasRight();
+    bool contiguous = first_unbounded && last_unbounded;
+    for (size_t i = 1; contiguous && i < first_column_ranges.size(); ++i) {
+      const ColumnRange* prev = first_column_ranges[i - 1];
+      const ColumnRange* curr = first_column_ranges[i];
+      // prev must have a right bound and curr a left bound with the same
+      // value; together they must not leave the shared point uncovered
+      // (i.e. not both exclusive).
```
MBkkt: // Adjacent ranges must meet at the same value with no gap
// (at least one endpoint must include the shared point).

server/connector/rocksdb_filter.cpp
```diff
@@ -197,152 +429,822 @@ velox::core::TypedExprPtr RewriteExpr(
...
+    } else if (IsCallOf(func_call, "_gt") || IsCallOf(func_call, "_gte") ||
+               IsCallOf(func_call, "_lt") || IsCallOf(func_call, "_lte")) {
+      pts = ExtractFilterComparison(func_call, pk_names, negated);
     } else if (IsCallOf(func_call, "_isnull") ||
                IsCallOf(func_call, "_is_null")) {
       // TODO: NOT NULL
       pts = ExtractFilterIsNull(func_call, pk_names);
     } else if (func_call->name() == velox::expression::kAnd) {
-      pts = ExtractFilterAnd(func_call, pk_names);
+      // De Morgan: NOT(A AND B) = NOT(A) OR NOT(B)
```
MBkkt, line 1082: we really write de morgan comments, it's school, without jokes about HSE

server/connector/rocksdb_filter.cpp
```diff
@@ -197,152 +429,822 @@ velox::core::TypedExprPtr RewriteExpr(
...
-  if (!absl::c_all_of(pts, [](const Point& p) { return p.IsSpecific(); })) {
-    return {{}, expr};
+  if (constraints.empty()) {
+    return {ConstraintKind::None, {}, expr};
   }
 
-  // Collect the unique source sub-expressions to get rid of them.
+  // Contradictory predicate (e.g. a < 1 AND a > 1): produce a range scan
+  // with zero ranges so no rows are returned without a full scan.
+  if (absl::c_all_of(constraints,
```
MBkkt, line 1110: // Contradictory predicate (e.g. a < 1 AND a > 1): no rows can match,
// so produce zero ranges to skip reading entirely.

server/connector/rocksdb_filter.cpp
```diff
@@ -197,152 +429,822 @@ velox::core::TypedExprPtr RewriteExpr(
...
+
+  SDB_ASSERT(
+    !absl::c_all_of(ranges,
+                    [](const KeyBounds& kc) { return kc.IsSpecificPoint(); }),
+    "Specific points should prepared separately for efficiency reason");
+#endif
+
+  std::vector<ResolvedRange> result;
+  result.reserve(ranges.size());
+  for (const auto& key_contraint : ranges) {
```
MBkkt, line 1033: use auto inside loop plz

server/connector/rocksdb_filter.cpp
```diff
@@ -197,152 +429,822 @@ velox::core::TypedExprPtr RewriteExpr(
...
+    const size_t prefix_size = key_contraint.RangePrefixSize();
+    SDB_ASSERT(prefix_size > 0,
+               "ToSpecificRanges: constraint must have PrefixSize() >= 1");
```
MBkkt: comment is outdated and useless, remove it

server/connector/rocksdb_filter.cpp
```diff
@@ -197,152 +429,822 @@ velox::core::TypedExprPtr RewriteExpr(
...
-  std::vector<Point> pts;
+std::vector<ResolvedRange> ToResolvedRanges(
+  const std::vector<KeyBounds>& ranges, const velox::RowType& pk_type) {
+  if (ranges.empty()) {
+    return {ResolvedRange::Contradictory()};
+  }
+#ifdef SDB_DEV
+  for (size_t i = 0; i < ranges.size(); ++i) {
+    for (size_t j = i + 1; j < ranges.size(); ++j) {
+      SDB_ASSERT(!KeyBounds::TryIntersect(ranges[i], ranges[j]).has_value(),
```
MBkkt:
```suggestion
      SDB_ASSERT(!KeyBounds::TryIntersect(ranges[i], ranges[j]),
```
codestyle

server/connector/rocksdb_filter.cpp
```diff
@@ -197,152 +429,822 @@ velox::core::TypedExprPtr RewriteExpr(
...
+                 "Resolved Ranges must be non-overlapping");
```
MBkkt:
```suggestion
                 "Resolved ranges must be non-overlapping");
```

server/connector/rocksdb_filter.cpp
```diff
@@ -197,152 +429,822 @@ velox::core::TypedExprPtr RewriteExpr(
...
     }
     result.push_back(std::move(sp));
   }
   return result;
 }
 
-std::vector<Point> ExtractFilterExpr(const velox::core::TypedExprPtr& expr,
-                                     std::span<const std::string> pk_names) {
-  std::vector<Point> pts;
+std::vector<ResolvedRange> ToResolvedRanges(
```
MBkkt: Disjoint/Normalized?

server/connector/data_source.hpp
```diff
@@ -215,6 +215,22 @@ class RocksDBFullScanDataSource : public RocksDBBaseDataSource {
...
+  using Base = RocksDBPerColumnIteratorDataSource<Source>;
+
+ public:
+  using Base::Base;
+
+  void addSplit(
+    std::shared_ptr<velox::connector::ConnectorSplit> split) override;
+
+ private:
+  template<std::invocable<const rocksdb::ReadOptions&> CreateFn>
```
MBkkt:
```suggestion
  template<typename CreateFn>
```
this is private method, do we really need it? can became static_assert btw

server/connector/data_source.hpp
```diff
@@ -215,6 +215,22 @@ class RocksDBFullScanDataSource : public RocksDBBaseDataSource {
...
+template<typename Source>
+class RocksDBFullScanDataSource
+  : public RocksDBPerColumnIteratorDataSource<Source> {
+  using Base = RocksDBPerColumnIteratorDataSource<Source>;
+
+ public:
+  using Base::Base;
+
+  void addSplit(
+    std::shared_ptr<velox::connector::ConnectorSplit> split) override;
```
MBkkt: sad that we need to have override method, maybe we can avoid override and have only finals?

tests/sqllogic/sdb/pg/data_source/choice.test
MBkkt, line 1: does this tests covers update/delete/insert (DML) with where? I think it's main usecase, please ensure this.
Maybe in separate file btw

server/connector/data_source.cpp
```diff
@@ -787,8 +789,457 @@ std::optional<velox::RowVectorPtr> RocksDBPointLookupDataSource<Policy>::next(
   return batch;
 }
 
+// Advances the last byte of `key` to make it an exclusive upper bound that
+// covers all keys sharing `key` as a prefix. Clears `key` to empty if all
+// bytes are 0xff (meaning no upper bound; caller interprets empty as
+// unbounded).
+static void MakeExclusiveUpperBound(std::string& key) {
```
MBkkt: avoid static, anon namespace

server/connector/data_source.cpp
```diff
@@ -787,8 +789,457 @@ std::optional<velox::RowVectorPtr> RocksDBPointLookupDataSource<Policy>::next(
...
+  for (int i = static_cast<int>(key.size()) - 1; i >= 0; --i) {
+    if (static_cast<unsigned char>(key[i]) < 0xff) {
+      ++reinterpret_cast<unsigned char&>(key[i]);
+      key.resize(static_cast<size_t>(i) + 1);
+      return;
+    }
+  }
```
MBkkt: btw feels mess

server/connector/data_source.cpp
```diff
@@ -787,8 +789,457 @@ std::optional<velox::RowVectorPtr> RocksDBPointLookupDataSource<Policy>::next(
...
+      opts.auto_prefix_mode = true;
+      opts.iterate_upper_bound = &e.upper_slice;
+      e.iter = factory(opts);
+    }
+
+    // TODO: maybe make lazy and seek in next() only
+    SeekToRange();
+  }
+
+  bool Valid() const override {
```
MBkkt: final for methods when possible, codestyle at least

server/connector/data_source.cpp
MBkkt, line 1: This file needs to be splited to multiple.
Can be todo

MBkkt (commented):
I think this work needs to be merged, but some simple things can be fixed and non trivial can be improved later

server/connector/rocksdb_filter.cpp
```diff
@@ -197,152 +429,822 @@ velox::core::TypedExprPtr RewriteExpr(
...
+      MergeSourceExprs(out.source_exprs, range.source_exprs);
+    }
+    return {std::move(out)};
+  }
+
+  const std::string& dim_col = pk_names[dim_idx];
+  const auto events = CollectEventPoints(ranges, dim_col);
+  const auto atoms = BuildAtoms(events);
+
+  static const ColumnRange kUnconstrained{};
```
MBkkt: a lot of place contains this, lets move to single global constant

server/connector/rocksdb_filter.cpp
```diff
@@ -197,152 +429,822 @@ velox::core::TypedExprPtr RewriteExpr(
...
+  KeyBounds constraint{pk_names};
+  constraint._column_ranges = std::move(ranges);
+  constraint._source_exprs = std::move(source_exprs);
+  return constraint;
+}
+
+bool ColumnRange::operator==(const ColumnRange& other) const noexcept {
+  return _flags == other._flags && _left_value == other._left_value &&
+         _right_value == other._right_value;
+}
```
MBkkt: default ==?

server/connector/rocksdb_filter.cpp
```diff
@@ -197,152 +429,822 @@ velox::core::TypedExprPtr RewriteExpr(
...
+    absl::StrAppend(&result, name, ": ", it->second.toString());
+  }
+  absl::StrAppend(&result, "}");
+  return result;
+}
+
+size_t KeyBounds::RangePrefixSize() const noexcept {
+  for (size_t k = 0; k < _pk_names.size(); ++k) {
+    const ColumnRange* column_range = FindColumnRange(_pk_names[k]);
+    if (column_range == nullptr) {
```
MBkkt:
```suggestion
    if (!column_range ) {
```
codestyle

server/connector/rocksdb_filter.cpp
```diff
@@ -197,152 +429,822 @@ velox::core::TypedExprPtr RewriteExpr(
...
+// differing column so overlapping ranges become adjacent.
+// ── Sweep helpers ────────────────────────────────────────────────────────────
+
+// One atomic segment on a single dimension's axis.
+// is_point=false -> open interval (left, right); has_left/has_right mark
+// whether the endpoint is finite. is_point=true -> closed singleton {left}.
+// Atoms are represented as ColumnRanges: points use Point(), open intervals
+// use Bounded/LeftBound/RightBound with exclusive endpoints, and the fully
+// unconstrained atom is an empty ColumnRange{} (no flags set).
+using Atom = ColumnRange;
```
MBkkt, line 454: It will be right to separate this algo from code and add unit tests on this

server/connector/rocksdb_filter.cpp
```diff
@@ -197,152 +429,822 @@ velox::core::TypedExprPtr RewriteExpr(
   }
 
   if (call->name() == "or") {
```
MBkkt: velox name for this things

server/connector/serenedb_connector.hpp
mkornaukhov: Secondary index support will be in in separate PR

server/connector/rocksdb_filter.hpp
mkornaukhov: There are some ternary operators that uses static case 0 to uint8_t, so it's better to have a special constant for that

server/connector/rocksdb_filter.hpp
mkornaukhov: Yes, SpecificPoint -> ResolvedPoint

server/connector/rocksdb_filter.hpp
mkornaukhov: https://github.com/serenedb/serenedb/issues/510

server/connector/data_source.hpp
mkornaukhov: Yes, but there is warning will be emitted

server/connector/rocksdb_filter.cpp
mkornaukhov: fixed

server/connector/rocksdb_filter.cpp
mkornaukhov: added todo for now

server/connector/rocksdb_filter.cpp
mkornaukhov: get rid of this

server/connector/serenedb_connector.hpp
mkornaukhov: Left it as is

Issue #458, opened 2026-03-29
pg_catalog.pg_relation_size doesn't exist


gnusi:
When trying to expand "Indexes" entry in DBeaver:
<img width="618" height="233" alt="Image" src="https://github.com/user-attachments/assets/7eae0b46-c14d-4eb9-90e2-fd4bc3eafed3" />

PR #466, opened 2026-03-30
Add `pg_relation_size`

codeworse:
fix for DBeaver: https://github.com/serenedb/serenedb/issues/458


MBkkt (approved):
but I have similar PR, so maybe I will just get your test?

With more changes oid/etc

PR #468, opened 2026-03-31
feat : Support ENUM + user composite type

pashandor789:


MBkkt (commented):
```
  {"pg_attribute", SystemTable<PgAttribute>{}.RowType()},
  {"pg_type", SystemTable<PgType>{}.RowType()},
```

we have such code, instead I think we should use your types with something like system_types array of strings to create them?

PR #470, opened 2026-03-31
Fixes for Drop + more tests

codeworse:
- [x] fixes
- [x] more tests
- [x] add `sdb_search_tasks_status` system table:
```
   task_type   | active_tasks | pending_tasks | threads
---------------+--------------+---------------+---------
 Commit        |            0 |             0 |       5
 Consolidation |            0 |             0 |       5
(2 rows)
```

### DIFF
--- a/tests/sqllogic/pg/simple/async.test
+++ b/tests/sqllogic/pg/simple/async.test
@@ -6,8 +6,11 @@ statement ok
 INSERT INTO t VALUES (NOW());
 
 
-statement async ok
-SELECT pg_sleep(2);
+query async
+SELECT 1 FROM pg_sleep(2);
+----
+?column?
+1
 
 
 statement async ok


server/search/task.cpp
```diff
@@ -82,11 +74,12 @@ void CommitTask::operator()() {
   SDB_TRACE("xxxxx", Logger::SEARCH, "CommitTask started");
   const char run_id = 0;
   auto data = _inverted_index_shard.lock();
+  SDB_IF_FAILURE("slow_search_task") { absl::SleepFor(absl::Seconds(5)); }
   absl::Cleanup set_promise = [promise = std::move(_promise)]() mutable {
     SDB_ASSERT(promise.Valid());
     std::move(promise).Set();
   };
-  if (!data) {
+  if (!data || data.use_count() == 1) {
```
MBkkt:
```suggestion
  if (!data) {
```
`data.use_count() == 1` what is point of this? it looks mess I don't think we want this, because two reasons:
1. it's just an optimization
2. it's very rare case that actually happen only for your failure point

server/search/task.h
```diff
@@ -119,10 +126,10 @@ class ConsolidationTask : public Task {
...
   void operator()();
-  ConsolidationTask GetContinuos() const {
-    auto self = _inverted_index_shard.lock();
-    SDB_ASSERT(self);
-    return ConsolidationTask(self, _progress);
+  ConsolidationTask GetContinuos(
+    std::shared_ptr<InvertedIndexShard>&& inverted_index_shard) const {
+    SDB_ASSERT(inverted_index_shard);
+    return ConsolidationTask(inverted_index_shard, _progress);
   }
```
MBkkt, line 133: by value + move

PR #475, opened 2026-04-01
feat : Make precommit delete ugly llm symbols + fix namespace format

pashandor789:
…fail


server/pg/functions/regexp.cpp
```diff
@@ -373,7 +373,7 @@ struct PgRegexpSubstr {
   velox::functions::detail::ReCache _cache;
 };
 
-// regexp_count(string, pattern, start) — 3-arg with start position
+// regexp_count(string, pattern, start) -- 3-arg with start position
```
MBkkt:
```suggestion
```

PR #488, opened 2026-04-01
fix: Fix issue with reading postions of term from last document in se…

MBkkt:
…gment if pos iterator resetted


Dronplane (approved):
LGTM

PR #489, opened 2026-04-02
feat: OFFSETS function for inverted index

Dronplane:


libs/iresearch/include/iresearch/search/phrase_iterator.hpp
```diff
@@ -406,7 +406,7 @@ class FixedPhraseFrequency {
   }
 
   IRS_FORCE_INLINE uint32_t NextPosition() {
-    if constexpr (HasIntervals) {
+    if constexpr (HasIntervals || Offs) {
```
MBkkt, line 409: looks like todo/issue

libs/iresearch/include/iresearch/search/phrase_query.cpp
```diff
@@ -182,7 +182,9 @@ DocIterator::ptr FixedPhraseQuery::ExecuteWithOffsets(
       auto back = std::prev(std::end(phrase_state->terms));
 
       while (term_state != back) {
-        if (!add_iterator(kRequiredFeatures)) {
+        // TODO(Dronplane) Make Adapter accept itarators without offs
+        // We will not need ones from middle iterators.
+        if (!add_iterator(kRequireOffs)) {
```
MBkkt, line 187: I think the issue that right now we have for all iterators same type.
We probably can change this, and have separate type for first/last and middles

server/catalog/table_options.h
```diff
@@ -138,6 +139,27 @@ struct Column {
...
+    "sdb_inverted_index_offsets$";
+
+  // Encodes the searched column's catalog ID into the virtual column name.
+  static std::string MakeOffsetsName(Id column_id) {
+    static_assert(kOffsetsNamePrefix.ends_with(query::kReservedSymbol));
+    return absl::StrCat(kOffsetsNamePrefix, column_id);
+  }
+
+  // ARRAY(BIGINT) -- flat offsets column: interleaved start,end pairs.
+  static velox::TypePtr OffsetsType() { return velox::ARRAY(velox::BIGINT()); }
```
MBkkt: imho, type is mess, can we have todo/issue about make it array<row>?

server/catalog/table_options.h
```diff
@@ -138,6 +139,27 @@ struct Column {
...
+
+  // Request for a single OFFSETS() column: which catalog column to extract
+  // offsets for, and how many offset pairs to return per document at most.
+  struct OffsetsFieldRequest {
+    Id column_id;
+    size_t limit = pg::Objects::kDefaultOffsetsLimit;
```
MBkkt, line 165: is it possible no-limit? i think it's important

tests/sqllogic/sdb/pg/index/offsets.test
```diff
@@ -0,0 +1,274 @@
...
+WHERE PHRASE(title, 'hello') OR PHRASE(title, 'world')
+ORDER BY id;
+----
+id	to_
+1	{0,5,6,11}
+2	{0,5}
+3	{0,5}
+
+query
+SELECT id, OFFSETS(body) AS bo
```
MBkkt, line 141: can we write test queries with functions that actually materialize offsets to array of strings/highlights?

because without this it's quite complicated to image usecase. Also function/option that converts to utf8 offsets will be nice

tests/sqllogic/sdb/pg/index/offsets.test
MBkkt, line 1: should be tests with join/subqueries/unions/ctes

especially for cases when offsets function used outside of their scope


Example:
```
SELECT index.id OFFSETS(index.c) FROM (SELECT * FROM index);
```

in such case in the final execution plan we shouldn't read index.c at all

server/pg/sql_collector.cpp
```diff
@@ -246,6 +248,50 @@ void ObjectCollector::CollectFuncCall(const State& state,
...
+    size_t limit = Objects::kDefaultOffsetsLimit;
+    if (nargs == 2) {
+      const auto* limit_arg = lsecond_node(Node, expr.args);
+      if (!IsA(limit_arg, A_Const) ||
+          nodeTag(&castNode(A_Const, limit_arg)->val) != T_Integer) {
+        THROW_SQL_ERROR(
+          ERR_CODE(ERRCODE_INVALID_PARAMETER_VALUE),
+          ERR_MSG("OFFSETS() second argument must be an integer constant"));
+      }
+      const auto v = intVal(&castNode(A_Const, limit_arg)->val);
```
MBkkt: we have function for this, TryGet<int>(args, i)  use it plz

server/pg/sql_analyzer_velox.cpp
```diff
@@ -4098,6 +4113,14 @@ State SqlAnalyzer::ProcessIndex(State* parent,
         ERR_MSG("Only one inverted index scan can produce a score per query"));
     }
 
+    // TODO(Dronplane): looks like this restriction could be lifted if we make
+    // _exprs_for_offsets kind of map by index_id->offsets.
+    if (!_exprs_for_offsets.empty()) {
+      THROW_SQL_ERROR(
+        ERR_CODE(ERRCODE_FEATURE_NOT_SUPPORTED),
+        ERR_MSG("Only one inverted index scan can produce offsets per query"));
```
```suggestion
        ERR_MSG("Only one inverted index scan can produce offsets per sub-query"));
```
also change message for scorer plz, I think it's misleading, because we can produce multiple scorers/offsets in same query, it's just not that simple

server/connector/serenedb_connector.hpp
```diff
@@ -544,6 +545,18 @@ class IndexTable : public axiom::connector::Table {
         catalog::Column::GenerateScoreName(table.type()->names());
       columns.push_back(std::make_unique<SereneDBColumn>(
         score_name, velox::REAL(), catalog::Column::kInvertedIndexScoreId));
+      // Add one virtual offsets column per requested index column.
+      // It is only included in the output type when a corresponding offsets
+      // requested. We can not emit here only requested as index could be
+      // referenced multiple times and first creation of IndexTable object could
+      // be without offsets. So we must ensure we always have such columns for
+      // every indexed column.
+      auto offsets_type = catalog::Column::OffsetsType();
```
```suggestion
      auto offsets_type = catalog::Column::MakeOffsetsType();
```
otherwise looks like constructor not function call

server/connector/search_scan_data_source.cpp
```diff
@@ -133,6 +166,74 @@ std::optional<velox::RowVectorPtr> SearchDataSource<Materializer>::next(
           block_count = 0;
         }
       }
+      for (size_t fi = 0; fi < n_offsets_fields; ++fi) {
+        auto& doc_offsets = offsets_data[fi].emplace_back();
+        absl::flat_hash_set<uint64_t> seen_offsets;
```
MBkkt: containers::FlatHashSet, I actually start thinking about write pre-commit script for this :(

Also all these collect part looks quite sad, can you at least move it to separate functions, not making common code suffer because of this

server/connector/search_scan_data_source.cpp
MBkkt, line 1: offsets code is quite sad and should be isolated from other code

server/connector/search_filter_builder.hpp
```diff
@@ -33,6 +33,11 @@
 
 namespace sdb::connector {
 
+// Encodes column_id as an 8-byte big-endian binary string into field_name,
+// matching the IResearch field name produced by
+// MakeFieldName(SearchColumnInfo).
+void MakeFieldName(catalog::Column::Id column_id, std::string& field_name);
```
MBkkt, line 39: but we use tokenizer id/name too? what will be for same column indexed twice?
or is it impossible now? :(

server/catalog/table_options.h
Dronplane: Yes. One can pass 0 as limit  to request all. There are tests for that.

server/connector/search_filter_builder.hpp
Dronplane: For now we can't have column indexed twice. If someday we will allow we can have here dictionary name to resolve ambiguity. But that would be lesser problem of all others that need to be solved to allow multi-indexed column.

server/connector/search_scan_data_source.cpp
Dronplane: Yes, that "not-use-containers" appearance starts to be annoying. Maybe indeed add pre-commit scripts at least for primitive types/strings etc? WDYT?

Dronplane: Maybe a better solution https://github.com/serenedb/serenedb/pull/502 for such appearances

MBkkt: I think problem that our codestyle says absl, lets change codestyle in contributing.md instead and watch

tests/sqllogic/sdb/pg/index/offsets.test
Dronplane: added

server/connector/search_scan_data_source.cpp
Dronplane: Done

tests/sqllogic/sdb/pg/index/offsets.test
Dronplane: Done.  Added simple tests. utf8/convenience functions will go as a folowups

libs/iresearch/include/iresearch/search/phrase_iterator.hpp
Dronplane: Will create an issue

libs/iresearch/include/iresearch/search/phrase_query.cpp
Dronplane: Will create an issue

server/catalog/table_options.h
Dronplane: Will create an issue

libs/iresearch/include/iresearch/search/phrase_iterator.hpp
Dronplane: https://github.com/serenedb/serenedb/issues/514

libs/iresearch/include/iresearch/search/phrase_query.cpp
Dronplane: https://github.com/serenedb/serenedb/issues/515

server/catalog/table_options.h
Dronplane: https://github.com/serenedb/serenedb/issues/516

tests/sqllogic/sdb/pg/index/offsets.test
Dronplane: https://github.com/serenedb/serenedb/issues/518

PR #491, opened 2026-04-02
Add Union Tokenizer

### LINKED ISSUE #440: Union tokenizer
## Summary

Add a `union` tokenizer that accepts multiple sub-tokenizers and concatenates the token streams from each of them over the same input. Unlike `pipeline` (which chains analyzers sequentially — output of one feeds into the next), `union` runs all sub-tokenizers independently on the original input and merges their results.

## Motivation

This enables indexing the same field with multiple analysis strategies simultaneously — e.g. combining a stemming tokenizer with an n-gram tokenizer so that both exact-prefix and stemmed matches work against a single field without duplicating data.

## Behavior

Given sub-tokenizers `[A, B, C]` and input text:

1. `reset(data)` calls `reset(data)` on each sub-tokenizer
2. `next()` yields all tokens from `A`, then all tokens from `B`, then all tokens from `C`
3. `OffsAttr` values are preserved from each sub-tokenizer (all relative to original input)
4. `IncAttr` for the first token of each subsequent sub-tokenizer should be `0` to place it at the same logical position as the first token — this way terms from different sub-tokenizers at the same offset overlap in the position space

## Configuration

```sql
CREATE TEXT SEARCH DICTIONARY union_dict (
    TEMPLATE = 'union',

    TOKENIZER1_TEMPLATE = 'text',
    TOKENIZER1_LOCALE = 'en_US.UTF-8',
    TOKENIZER1_CASE = 'lower',
    TOKENIZER1_STEMMING = true,

    TOKENIZER2_TEMPLATE = 'ngram',
    TOKENIZER2_MINGRAM = 3,
    TOKENIZER2_MAXGRAM = 3,

    FREQUENCY = true,
    POSITION = true
);
```

## Implementation Notes

- Follows the same pattern as `PipelineTokenizer` — stores a `std::vector<Analyzer::ptr>`, uses `TOKENIZER<N>_` prefix convention for sub-tokenizer configuration
- Key difference: each sub-tokenizer sees the original input, not the output of the previous one
- Position increment handling at sub-tokenizer boundaries needs care to avoid inflating position gaps

abhishek593:
Fixes #440 
## Summary

This PR adds a new `union` tokenizer that runs multiple child tokenizers on the same input and merges their output into a single token stream.

## Behavior

  - `reset(data)` resets every child on the same input
  - `next()` interleaves child tokens by each child tokenizer's own logical position
  - Tokens from different children at the same position overlap in the merged stream
  - `OffsAttr` is not exposed

`union` cannot safely expose `OffsAttr` because merged child streams do not have a single globally monotonic offset order. With heterogeneous children, such as text and ngram, preserving original child offsets would make offsets move backwards in the merged stream, which is invalid for indexing. 

```
For example, given the input `hello world`:

  - `text` emits:
    - `hello` at offsets `[0,5)`
    - `world` at offsets `[6,11)`
  - `ngram(3)` emits:
    - `hel` at `[0,3)`
    - `ell` at `[1,4)`
    - `llo` at `[2,5)`
    - ...

The union tokenizer interleaves by child position, not by byte offset. The merged order is roughly:

  - `hello` at position `1`, offsets `[0,5)`
  - `hel` at position `1`, offsets `[0,3)`
  - `world` at position `2`, offsets `[6,11)`
  - `ell` at position `2`, offsets `[1,4)`

If child offsets were preserved, the start-offset sequence would be:

  - `0`
  - `0`
  - `6`
  - `1`

That final step goes backwards, which breaks the monotonic offset ordering expected from a single token
stream during indexing.
```

## Tests

Added tests for construction, normalization, invalid configs, nesting with `pipeline`, and merged-token behavior.

### DIFF
--- a/libs/iresearch/include/iresearch/analysis/union_tokenizer.cpp
+++ b/libs/iresearch/include/iresearch/analysis/union_tokenizer.cpp
@@ -1,0 +1,399 @@
+////////////////////////////////////////////////////////////////////////////////
+/// DISCLAIMER
+///
+/// Copyright 2025 SereneDB GmbH, Berlin, Germany
+///
+/// Licensed under the Apache License, Version 2.0 (the "License");
+/// you may not use this file except in compliance with the License.
+/// You may obtain a copy of the License at
+///
+///     http://www.apache.org/licenses/LICENSE-2.0
+///
+/// Unless required by applicable law or agreed to in writing, software
+/// distributed under the License is distributed on an "AS IS" BASIS,
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+///
+/// Copyright holder is SereneDB GmbH, Berlin, Germany
+////////////////////////////////////////////////////////////////////////////////
+
+#include "union_tokenizer.hpp"
+
+#include <vpack/builder.h>
+#include <vpack/common.h>
+#include <vpack/iterator.h>
+#include <vpack/parser.h>
+#include <vpack/slice.h>
+
+#include <string_view>
+
+#include "iresearch/utils/vpack_utils.hpp"
+
+namespace irs::analysis {
+namespace {
+
+constexpr std::string_view kUnionParamName = "union";
+constexpr std::string_view kTypeParamName = "type";
+constexpr std::string_view kPropertiesParamName = "properties";
+
+using OptionsNormalize = std::vector<std::pair<std::string, std::string>>;
+
+template<typename T>
+bool ParseVPackOptions(const vpack::Slice slice, T& options) {
+  if constexpr (std::is_same_v<T, UnionTokenizer::OptionsT>) {
+    SDB_ASSERT(options.empty());
+  }
+  if (!slice.isObject()) {
+    SDB_ERROR("xxxxx", sdb::Logger::IRESEARCH,
+              "Not a VPack object passed while constructing union_tokenizer");
+    return false;
+  }
+
+  if (auto union_slice = slice.get(kUnionParamName); !union_slice.isNone()) {
+    if (union_slice.isArray()) {
+      for (const auto member : vpack::ArrayIterator(union_slice)) {
+        if (member.isObject()) {
+          std::string_view type;
+          if (auto type_attr_slice = member.get(kTypeParamName);
+              !type_attr_slice.isNone()) {
+            if (type_attr_slice.isString()) {
+              type = type_attr_slice.stringView();
+            } else {
+              SDB_ERROR("xxxxx", sdb::Logger::IRESEARCH, "Failed to read '",
+                        kTypeParamName, "' attribute of '", kUnionParamName,
+                        "' member as string while constructing "
+                        "union_tokenizer from VPack arguments");
+              return false;
+            }
+          } else {
+            SDB_ERROR("xxxxx", sdb::Logger::IRESEARCH, "Failed to get '",
+                      kTypeParamName, "' attribute of '", kUnionParamName,
+                      "' member while constructing union_tokenizer "
+                      "from VPack arguments");
+            return false;
+          }
+          if (auto properties_attr_slice = member.get(kPropertiesParamName);
+              !properties_attr_slice.isNone()) {
+            if constexpr (std::is_same_v<T, UnionTokenizer::OptionsT>) {
+              auto analyzer =
+                analyzers::Get(type, irs::Type<text_format::VPack>::get(),
+                               {properties_attr_slice.startAs<char>(),
+                                properties_attr_slice.byteSize()});
+
+              // fallback to json format if vpack isn't available
+              if (!analyzer) {
+                analyzer =
+                  analyzers::Get(type, irs::Type<text_format::Json>::get(),
+                                 slice_to_string(properties_attr_slice));
+              }
+
+              if (analyzer) {
+                options.push_back(std::move(analyzer));
+              } else {
+                SDB_ERROR("xxxxx", sdb::Logger::IRESEARCH,
+                          "Failed to create union member of type '", type,
+                          "' with properties '",
+                          slice_to_string(properties_attr_slice),
+                          "' while constructing union_tokenizer "
+                          "from VPack arguments");
+                return false;
+              }
+            } else {
+              std::string normalized;
+              if (analyzers::Normalize(normalized, type,
+                                       irs::Type<text_format::VPack>::get(),
+                                       {properties_attr_slice.startAs<char>(),
+                                        properties_attr_slice.byteSize()})) {
+                options.emplace_back(std::piecewise_construct,
+                                     std::forward_as_tuple(type),
+                                     std::forward_as_tuple(normalized));
+
+                // fallback to json format if vpack isn't available
+              } else if (analyzers::Normalize(
+                           normalized, type,
+                           irs::Type<text_format::Json>::get(),
+                           slice_to_string(properties_attr_slice))) {
+                // in options we'll store vpack as string
+                auto vpack = vpack::Parser::fromJson(normalized.c_str(),
+                                                     normalized.size());
+                options.emplace_back(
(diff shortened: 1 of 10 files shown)


MBkkt (commented):
@abhishek593 Hi, thanks for your work, it's interesting!

Could you please update your PR with the latest main branch? 
Our CI has been updated, so we need this to run tests for your PR.

abhishek593:
@MBkkt I have updated the PR with the latest main branch.

tests/sqllogic/sdb/pg/simple/tokenizers/text_tokenizer.test
MBkkt, line 1: Is union as part of pipeline supported? If not should be proper error.
But it will be nice if it will be supported. 

If it will works important to check that offsets on pipeline also will be rejected if one of step contains union

libs/iresearch/include/iresearch/analysis/union_tokenizer.cpp
```diff
@@ -0,0 +1,408 @@
...
+        // Copy payload if this sub has one, else clear
+        if (sub.pay) {
+          _payload_buf.assign(sub.pay->value.begin(), sub.pay->value.end());
+          _payload.value = bytes_view{_payload_buf};
+        } else {
+          _payload.value = {};
+        }
+
+        // IncAttr: delta from last emitted position
+        std::get<IncAttr>(_attrs).value = _current_min_pos - _last_emitted_pos;
```
MBkkt, line 320: assert that `_current_min_pos >= _last_emitted_pos`

libs/iresearch/include/iresearch/analysis/union_tokenizer.hpp
```diff
@@ -0,0 +1,121 @@
...
+    uint32_t position{0};
+
+   private:
+    Analyzer::ptr _analyzer;
+  };
+
+  uint32_t find_min_position() const noexcept;
+
+  using subs_t = std::vector<SubAnalyzerT>;
+  using attributes = std::tuple<IncAttr, TermAttr, AttributePtr<PayAttr>>;
```
MBkkt: Please follow codestyle (.clang-tidy)

```
SubAnalyzers
Attributes
```

libs/iresearch/include/iresearch/analysis/union_tokenizer.hpp
```diff
@@ -0,0 +1,121 @@
...
+
+  subs_t _subs;
+  size_t _emit_index{0};
+  uint32_t _current_min_pos{std::numeric_limits<uint32_t>::max()};
+  uint32_t _last_emitted_pos{0};
```
MBkkt:
```suggestion
  size_t _emit_index = 0;
  uint32_t _current_min_pos = std::numeric_limits<uint32_t>::max();
  uint32_t _last_emitted_pos = 0;
```

libs/iresearch/include/iresearch/analysis/union_tokenizer.hpp
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  bool next() final;
+  bool reset(std::string_view data) final;
+
+  /// @brief calls visitor on union members in respective order.
+  /// Visiting is interrupted on first visitor returning false.
+  /// @return true if all visits returned true, false otherwise
+  template<typename Visitor>
+  bool visit_members(Visitor&& visitor) const {
```
MBkkt: Check .clang-tidy please 

```suggestion
  bool VisitMembers(Visitor&& visitor) const {
```

libs/iresearch/include/iresearch/analysis/union_tokenizer.cpp
```diff
@@ -0,0 +1,408 @@
...
+UnionTokenizer::SubAnalyzerT::SubAnalyzerT(Analyzer::ptr a)
+  : term(irs::get<TermAttr>(*a)),
+    inc(irs::get<IncAttr>(*a)),
+    pay(irs::GetMutable<PayAttr>(a.get())),
+    _analyzer(std::move(a)) {
+  SDB_ASSERT(inc);
+  SDB_ASSERT(term);
+}
+
+UnionTokenizer::SubAnalyzerT::SubAnalyzerT()
```
MBkkt:
```suggestion
UnionTokenizer::SubAnalyzer::SubAnalyzer()
```

libs/iresearch/include/iresearch/analysis/union_tokenizer.cpp
```diff
@@ -0,0 +1,408 @@
...
+                           "' member while constructing union_tokenizer "
+                           "from VPack arguments"));
+            return false;
+          }
+        } else {
+          SDB_ERROR("xxxxx", sdb::Logger::IRESEARCH,
+                    absl::StrCat("Failed to read '", kUnionParamName,
+                                 "' member as object while constructing "
+                                 "union_tokenizer from VPack arguments"));
+          return false;
```
MBkkt:
```suggestion
          SDB_ERROR("xxxxx", sdb::Logger::IRESEARCH,
                    "Failed to read '", kUnionParamName,
                                 "' member as object while constructing "
                                 "union_tokenizer from VPack arguments");
          return false;
```

MBkkt (commented):
@abhishek593 

Thanks, tests are mostly good, only few missing cases!

I think implementation is naive but it's better than no union tokenizer at all.

So my suggestion to you is add more tests, fix codestyle and other minor issues.

When we can merge this

libs/iresearch/include/iresearch/analysis/union_tokenizer.cpp
abhishek593: Done

libs/iresearch/include/iresearch/analysis/union_tokenizer.cpp
abhishek593: Done

libs/iresearch/include/iresearch/analysis/union_tokenizer.hpp
abhishek593: Done

libs/iresearch/include/iresearch/analysis/union_tokenizer.hpp
abhishek593: Done

libs/iresearch/include/iresearch/analysis/union_tokenizer.hpp
abhishek593: Done

libs/iresearch/include/iresearch/analysis/union_tokenizer.cpp
abhishek593: Added

tests/sqllogic/sdb/pg/simple/tokenizers/text_tokenizer.test
abhishek593: Union inside pipeline works - tested in `pipe_with_union`.
For offsets: added a post-construction check in `CreateTokenizer()` that probes the actual analyzer for `OffsAttr` when `offset = true`. Pipeline with a union step now correctly fails with "Unsupported index features" since the pipeline sees union has no offsets and nulls out its own `OffsAttr`. Added `pipe_union_offset_err` test for this.

abhishek593:
@MBkkt Please take a look at the updated changes. I also ran clang-tidy and fixed the naming/style issues. If you could point out what makes the current implementation naive, I'd like to improve it in a follow-up after this is merged.

MBkkt (approved):
@abhishek593 Thanks for your contribution, looks good!

As this is your first PR for SereneDB, please review our CLA at https://serenedb.com/cla and submit the form at the bottom of the page (message field required at least 8 symbols).

Regarding the naive implementation: it isn't optimized yet and will require benchmarking. However, based on the code, it looks like term/token copies can be completely avoided (same as in PipelineTokenizer).

abhishek593:
@MBkkt I have signed the CLA.

MBkkt:
@abhishek593 Thanks for your contribution!

PR #497, opened 2026-04-03
feat: BOOST function for inverted index

Dronplane:


server/functions/search.cpp
```diff
@@ -128,6 +136,10 @@ void RegisterSearchFunctions() {
   velox::registerFunction<SearchStubFunction, bool, velox::Varchar,
                           velox::Varchar, int64_t, bool, int64_t,
                           velox::Varchar>({std::string{kLevenshteinMatch}});
+
+  // BOOST(expr, boost_value)
+  velox::registerFunction<SearchStubFunction, bool, bool, double>(
```
MBkkt, line 141: why double instead real?

Dronplane: because "1.5" is double by default. If set  to "irs::score_t" aka float it would require users to write casts manually in queries. I guess same we did for NGRAM_MATCH above.

MBkkt (approved):

PR #536, opened 2026-04-09
More benchmarks on vector distances

codeworse:
- [x] numkong
- [x] ydb
- [ ] scann


MBkkt (commented):
bad, should be proper cosine

PR #542, opened 2026-04-12
[WIP] feat: Parallelizing Index Commit and Document Removal

mag7ster:


MBkkt (commented):
Lets keep this draft but start with simpler task in a separate PR

I think good first step is 
1. Add thread pool for IndexWriter (make tests pass it and common code that run database pass it too)
2. Apply removes parallel for each segment (just submit + WaitGroup)

Making flushes async is important but quite complicated task that requires more effort.

PR #544, opened 2026-04-13
feat: Ngram wildcard and Geo search

Dronplane:
Functionality:
Adds wildcard template for text dictionary
Adds Ngram-wildcard implementation for LIKE operator.
Adds geopoint and geojson templates
Adds GEO_DISTANCE, GEO_CONTAINS, GEO_ITERSECTS and GEO_IN_RANGE functions for search with inverted index

Refactoring:
Wildcard analyzer switched to Re2 matcher
Tests added for WildcardFilter
Storing values from analyzers now wired through new StoreAttr attribute not by externally detected store function
Geo analyzers now have explicit slice reset and basic reset does not expects  string_view to be actually a slice.


tests/sqllogic/sdb/pg/index/term_ops.test
```diff
@@ -99,3 +99,120 @@ a
...
 statement ok
 DROP TABLE test_term_ops;
+
+# --- Wildcard text dictionary tests ---
+
+statement ok
+CREATE TEXT SEARCH DICTIONARY test_wildcard_dict(
+    template = 'wildcard',
+    ngramsize = 3,
+    analyzer_template = 'delimiter',
```
gnusi: let's use `tokenizer_` prefix?

server/connector/search_sink_writer.cpp
```diff
@@ -198,14 +222,23 @@ void SearchSinkInsertBaseImpl::SetupColumnWriter(catalog::Column::Id column_id,
   }
 }
 
-template<typename WriteFunc>
+template<bool HasStore, typename WriteFunc>
```
gnusi, line 225: Honestly, I think having HasStore is a bit redundant. We solely rely on runtime check field.store_attr to avoid extra complexity. WDYT?

libs/iresearch/include/iresearch/analysis/token_attributes.hpp
```diff
@@ -67,6 +67,14 @@ struct PayAttr final : Attribute {
   bytes_view value;
 };
 
+// Represents a value to be stored in the index alongside the indexed field.
+// Analyzers that support storing populate this attribute in reset().
+struct StoreAttr final : Attribute {
```
gnusi, line 72: There is already `PayAttr`, we can reuse it.

Dronplane: PayAttr is for position payload. I've also seen it. But I believe better to have separate attr for separated purpose

server/connector/search_sink_writer.cpp
Dronplane: I've added it to make branchless path for analyzers that do not store anything  - that are actually most of the use cases. For HasStore = true runtime check is needed only for Nulls that have nothing to store.

tests/sqllogic/sdb/pg/index/term_ops.test
Dronplane: It is same as for minhash. BTW I guess tokenizer here is indeed better. But then we should change for minhash too to make it consistent. WDYT?

server/connector/search_filter_builder.cpp
```diff
@@ -653,12 +826,28 @@ Result FromVeloxLike(irs::BooleanFilter& filter, const VeloxFilterContext& ctx,
...
-  auto& wildcard_filter = ctx.negated ? Negate<irs::ByWildcard>(filter)
-                                      : AddFilter<irs::ByWildcard>(filter);
-  wildcard_filter.boost(ctx.boost);
-  *wildcard_filter.mutable_field() = field_name;
-  wildcard_filter.mutable_options()->term.assign(irs::ViewCast<irs::byte_type>(
-    static_cast<std::string_view>(value.value().value<velox::StringView>())));
+  if (column_info->analyzer.analyzer->type() ==
+      irs::Type<irs::analysis::WildcardAnalyzer>::id()) {
+    auto& wildcard_filter = ctx.negated
+                              ? Negate<irs::WildcardFilter>(filter)
```
gnusi: Just noticed that WildcardFilter is a bit misleading w.r.t. ByWildcard, maybe rename it to ByWildcardNgram?

Dronplane: Done

tests/sqllogic/sdb/pg/index/term_ops.test
Dronplane: Done for minhash + wildcard

MBkkt:
@Dronplane should we close this?

Dronplane:
Closed as done by another PR

PR #551, opened 2026-04-15
fix: default connection env, dependencies and ci

iljaname:
- Fixed issues with default environment connection creation
- Resolved dependency conflicts
- Removed obsolete update_ownership script


.github/workflows/serene-ui-reusable.yml
```diff
@@ -324,7 +321,7 @@ jobs:
     build-macos-app:
         name: Build macOS App
         if: inputs.RUN_TESTS || inputs.BUILD_MACOS_APP
-        runs-on: macos
+        runs-on: macOS
```
MBkkt, line 324: github runners are case-insensitive?

MBkkt (approved):

.github/workflows/serene-ui-reusable.yml
iljaname: According to github docs: Labels are case-insensitive.

But only after this fix gh runner on mac started to receive tasks

Issue #554, opened 2026-04-22
Improve explain output


mkornaukhov:
- [ ] Explain should print relation/column names, from last name to full name (add more layers until it becomes unambiguous)
- [ ] `explain (explain_output=all/physical_only/optimized_only)` instead of pragma https://duckdb.org/docs/current/guides/meta/explain
- [ ] Add new explain output option - before_optimization (already exists in output of "all") 
- [x] Better search/rocksdb filter printing: each on new line, not `...` in the end.
- [ ] New format kind that prints plan as axiom/velox do
- [ ] New format kind FULL: should print without any `...` fro to many rows or large width
- [ ] Add option that will always use full scan functions (or maybe disable our rule). Maybe option to disable any rule/rules?

PR #555, opened 2026-04-22
refactor: improve non-search scans

mkornaukhov:
- Print materializer for secondary index scans, fancier filters
- Straightforward algorithm for choosing the most appropriate scan with tests on expected logics

Closes some bullets in #554


server/connector/optimizer/rocksdb_plan.cpp
```diff
@@ -199,39 +169,68 @@ std::vector<Candidate> BuildCandidates(
...
-  if (a.kind != b.kind) {
-    return a.kind == connector::ConstraintKind::Points;
+// Scan-type order: Points > Ranges > Full.
+// For equal scan types:
+//   Points: lower num_points * (num_columns + sk_extra) is better.
+//   Ranges/Full: fewer effective columns (num_columns + sk_extra) is better.
+// Final tiebreaker: PK beats SK (no extra materialization needed).
+[[nodiscard]] bool StrictlyBetter(const PhysicalScanCandidate& lhs,
+                                  const PhysicalScanCandidate& rhs) {
+  static constexpr auto kScanRank = [](connector::ConstraintKind kind) {
```
gnusi: Maybe just assign these values to enum?

server/connector/optimizer/rocksdb_plan.cpp
```diff
@@ -199,39 +169,68 @@ std::vector<Candidate> BuildCandidates(
     }
     const auto& sk = basics::downCast<const catalog::SecondaryIndex>(*index);
     auto cols = sk.GetColumnIds();
-    Candidate c;
-    c.kind = Candidate::Kind::Sk;
-    c.column_ids.assign(cols.begin(), cols.end());
-    c.sk_shard_id = shard_id;
-    c.sk_unique = sk.IsUnique();
-    out.push_back(std::move(c));
+    out.push_back({IndexCandidate::Kind::Sk,
```
gnusi: emplace_back

server/connector/optimizer/rocksdb_plan.cpp
```diff
@@ -180,15 +156,9 @@ std::vector<Candidate> BuildCandidates(
   // FROM table_name: PK + all rocksdb-backed secondary indexes.
   auto pk_cols = bind_data.table->PKColumns();
   if (!pk_cols.empty()) {
-    Candidate pk;
-    pk.kind = Candidate::Kind::Pk;
-    pk.column_ids.assign(pk_cols.begin(), pk_cols.end());
-    out.push_back(std::move(pk));
+    out.push_back({IndexCandidate::Kind::Pk, {pk_cols.begin(), pk_cols.end()}});
```
gnusi: emplace_back

server/connector/optimizer/rocksdb_plan.cpp
```diff
@@ -166,12 +144,10 @@ std::vector<Candidate> BuildCandidates(
       }
       const auto& sk = basics::downCast<const catalog::SecondaryIndex>(*index);
       auto cols = sk.GetColumnIds();
-      Candidate c;
-      c.kind = Candidate::Kind::Sk;
-      c.column_ids.assign(cols.begin(), cols.end());
-      c.sk_shard_id = shard_id;
-      c.sk_unique = sk.IsUnique();
-      out.push_back(std::move(c));
+      out.push_back({IndexCandidate::Kind::Sk,
```
gnusi: emplace_back

server/connector/rocksdb_filter.cpp
```diff
@@ -1487,24 +1518,26 @@ ExtractAndRewriteResult ExtractAndRewriteFilterExpr(
...
-    for (size_t i = 0; i < prefix; ++i) {
-      const ColumnRange* r = c.FindColumnRange(key_ids[i]);
-      absl::StrAppend(&key, r ? r->toString() : std::string{"-"}, "|");
-    }
-    return key;
-  };
+  // so O(n) hash-set dedup on a prefix key beats std::unique.
   {
-    containers::FlatHashSet<std::string> seen;
+    absl::flat_hash_set<ScanKeyRef> seen;
```
gnusi: containers::FlatHashSet?

server/connector/duckdb_table_function.h
```diff
@@ -86,6 +86,12 @@ struct ScanSource {
            _kind == ScanSourceKind::Ann || _kind == ScanSourceKind::RangeSearch;
   }
 
+  // Covers SecondaryIndexScan / SkPointScan / SkRangeScan.
+  bool IsSkLike() const {
```
Dronplane: NIT noexcept?

server/connector/rocksdb_filter.cpp
```diff
@@ -1436,6 +1430,43 @@ std::vector<ResolvedRange> ToDisjointRanges(
...
+// Thin reference wrapper for scan-equivalence dedup. Holds only a pointer to a
+// KeyBounds and the shared key_ids span -- no per-entry heap allocation.
+// Two ScanKeyRefs are equal iff their prefix_size matches and every prefix
+// column range compares equal (via ColumnRange::operator==).
+struct ScanKeyRef {
+  const KeyBounds* key_bounds;
+  std::span<const catalog::Column::Id> key_ids;
+
+  bool operator==(const ScanKeyRef& other) const noexcept {
+    const size_t prefix_size = key_bounds->RangePrefixSize();
```
Dronplane: Could we ASSERT that both key_ids has at least prefix_size members?
I guess it is naturally true  for now as key_ids is same for all and represents full PK. But just to have clear intent for the reader.

server/connector/optimizer/rocksdb_plan.cpp
```diff
@@ -199,39 +169,68 @@ std::vector<Candidate> BuildCandidates(
...
+  auto cols_l = effective_cols(lhs);
+  auto cols_r = effective_cols(rhs);
+
+  if (lhs.result.kind == connector::ConstraintKind::Points) {
+    auto cost_l = lhs.result.constraints.size() * cols_l;
+    auto cost_r = rhs.result.constraints.size() * cols_r;
+    if (cost_l != cost_r) {
+      return cost_l < cost_r;
+    }
+  } else if (lhs.result.kind == connector::ConstraintKind::Ranges &&
```
Dronplane, line 212: Comment says " Two SK range scans with equal effective cols:" but condition itself does not checks cols_l and cols_r. Is it correct?

server/connector/optimizer/rocksdb_plan.cpp
```diff
@@ -199,39 +169,68 @@ std::vector<Candidate> BuildCandidates(
...
+             lhs.index->kind == IndexCandidate::Kind::Sk &&
+             rhs.index->kind == IndexCandidate::Kind::Sk) {
+    // Two SK range scans with equal effective cols: prefer the wider index.
+    // More index key columns means more filter predicates are pushed into the
```
Dronplane: > More index key columns means more filter predicates are pushed into

Is this true? For example lhs can have 2 columns indexed and they cover 3 of 4 filter predicates. Index rhs has 3 columns but the cover only 1 filter predicate. Or I'm missing something?

server/connector/rocksdb_filter.cpp
```diff
@@ -1487,24 +1518,26 @@ ExtractAndRewriteResult ExtractAndRewriteFilterExpr(
...
-    containers::FlatHashSet<std::string> seen;
+    absl::flat_hash_set<ScanKeyRef> seen;
     seen.reserve(constraints.size());
-    auto kept = std::remove_if(
-      constraints.begin(), constraints.end(),
-      [&](KeyBounds& c) { return !seen.insert(scan_key(c)).second; });
-    constraints.erase(kept, constraints.end());
+    std::vector<size_t> kept;
+    kept.reserve(constraints.size());
+    for (size_t i = 0; i < constraints.size(); ++i) {
```
Dronplane: Maybe single loop approach?
```
size_t i = 0;
while (i < constraints.size()) {
  if (seen.insert(ScanKeyRef{&constraints[i], key_ids}).second) {
    ++i;
    continue;
  }
  // replace duplicate at i with current last. pop back last to do cheap erase 
  if (i != constraints.size() - 1) {
    constraints[i] = std::move(constraints.back());
  }
  constraints.pop_back();
}
```
That affects order of constraints. But looks like that is not important in general.

server/connector/optimizer/rocksdb_plan.cpp
mkornaukhov: fixed

server/connector/optimizer/rocksdb_plan.cpp
mkornaukhov: fixed

server/connector/optimizer/rocksdb_plan.cpp
mkornaukhov: fixed

server/connector/optimizer/rocksdb_plan.cpp
mkornaukhov: fixed

server/connector/rocksdb_filter.cpp
mkornaukhov: ScanKeyRef contains a reference into constraints, they are going to be invalidated in this case

server/connector/optimizer/rocksdb_plan.cpp
mkornaukhov: They are really equal for secondary indexes. It's just all requested columns (from primary key) + one more column for getting primary keys bytes.

server/connector/optimizer/rocksdb_plan.cpp
mkornaukhov: Yeah, you are right. In general we need to somehow analyze remaining filter related to scan. Also some data from statics will be useful.

I assumed that number of filters is linearly proportional to the number of columns, but this has no good reason, yeah. I think it's ok for now. Or you have better idea for that?

server/connector/rocksdb_filter.cpp
mkornaukhov: fixed, neuroslopped a little bit

server/connector/duckdb_table_function.h
mkornaukhov: fixed

server/connector/rocksdb_filter.cpp
mkornaukhov: Moreover, key ids must be identical. Fixed

server/connector/optimizer/rocksdb_plan.cpp
Dronplane: then some assert? And do we really need to calculate `cols_l` and `cols_r` in this case?

server/connector/optimizer/rocksdb_plan.cpp
Dronplane: Can we check here number of columns "used" in filter per index? That is also not ideal but a bit better metric.
If not - let's at least make comment clear that this is our best-effort estimate and maybe a TODO for improvement.

server/connector/rocksdb_filter.cpp
Dronplane: Frankly speaking I don't see how. We never touch what is landed in the set and vector only shrinks so no reallocation.

server/connector/rocksdb_filter.hpp
```diff
@@ -302,14 +315,20 @@ struct ResolvedRange {
   // have different prefix depths but share the common prefix, the shorter
   // range's range_col is compared against the exact prefix value of the longer
   // range at that position.
-  bool operator<(const ResolvedRange& other) const {
+  std::weak_ordering operator<=>(const ResolvedRange& other) const {
```
MBkkt: strong_ordering

server/connector/rocksdb_filter.hpp
```diff
@@ -155,6 +156,18 @@ struct ColumnRange {
   // NOLINTNEXTLINE(readability-identifier-naming)
   [[nodiscard]] std::string toString() const;
 
+  template<typename H>
+  friend H AbslHashValue(H h, const ColumnRange& cr) {
+    h = H::combine(std::move(h), cr._flags);
+    if (cr.HasLeft()) {
+      h = H::combine(std::move(h), cr._left_value.Hash());
```
MBkkt: why Hash? cant we implement for _left_value AbslHashValue?

MBkkt (approved):

server/connector/optimizer/rocksdb_plan.cpp
mkornaukhov: Added assert, for this case is not needed

server/connector/optimizer/rocksdb_plan.cpp
mkornaukhov: Added clarification and TODO

server/connector/rocksdb_filter.cpp
mkornaukhov: Really we need sorted order later, to optimize some cases where ranges will read full table. But we actually need sort (that was a bug previously, I'll add unit tests for that later #570), so I've applied the suggest change.

server/connector/rocksdb_filter.hpp
mkornaukhov: fixed

server/connector/rocksdb_filter.hpp
mkornaukhov: fixed

PR #556, opened 2026-04-22
feat: Add README to serene-ui 

iljaname:
Add a README for SereneUI to give contributors and users a clearer starting point when they first open the project.
This helps explain the purpose of the app, reduces setup friction, and makes the repository easier to navigate without relying on external context.


MBkkt (approved):

gnusi:
Maybe add some gifs?

PR #558, opened 2026-04-22
feat: Add current_user, current_role and pg_backend_pid system functions

Dronplane:
current_user, current_role now returns actual role name. Test added that checks role returned is indeed in pg_roles.
pg_backend_pid is a stub that just return 0 fo now as we do not spawn backends per connection

### DIFF
--- a/server/connector/functions/system.cpp
+++ b/server/connector/functions/system.cpp
@@ -79,11 +79,6 @@ void CurrentUserFunction(duckdb::DataChunk&, duckdb::ExpressionState& state,
   result.Reference(conn_ctx.user());
 }
 
-void PgBackendPidFunction(duckdb::DataChunk&, duckdb::ExpressionState&,
-                          duckdb::Vector& result) {
-  result.Reference(duckdb::Value::INTEGER(0));
-}
-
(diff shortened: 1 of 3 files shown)


server/connector/functions/system.cpp
```diff
@@ -549,6 +561,16 @@ void RegisterPgSystemFunctions(duckdb::DatabaseInstance& db) {
...
                                                  PgSchemaSizeOidFunction});
+
+  loader.RegisterFunction(duckdb::ScalarFunction{
+    "current_user", {}, duckdb::LogicalType::VARCHAR, CurrentUserFunction});
+
+  // current_role is same as current_user in postgres
+  loader.RegisterFunction(duckdb::ScalarFunction{
+    "current_role", {}, duckdb::LogicalType::VARCHAR, CurrentUserFunction});
+
+  loader.RegisterFunction(duckdb::ScalarFunction{
```
MBkkt: let's make stub macro (system_functions)

Dronplane: good point

mkornaukhov (approved):
LGTM

PR #563, opened 2026-04-23
ANN search improvements: Index options + Filter

codeworse:
1. Add options for index creation:
```sql
CREATE INDEX i on t USING inverted(embedding hnsw (m = 64, ef_construction = 400);
```
2. Add filtering in k-NN/range queries:
```sql
SELECT id FROM i WHERE id % 2 = 0 ORDER BY embedding <-> [0, 0, 0]::FLOAT[3] LIMIT 10; -- k-NN
SELECT id FROM i WHERE id % 2 = 0 AND embedding <-> [0, 0, 0]::FLOAT[3] < radius; -- range scan
```
3. Some bug fixes


gnusi (approved):

server/catalog/index.cpp
```diff
@@ -33,6 +36,142 @@
...
+                                 ERROR_BAD_PARAMETER,
+                                 "Column '",
+                                 column_name,
+                                 "': hnsw option '",
+                                 key,
+                                 "' must be an integer, got '",
+                                 v.ToString(),
+                                 "'"};
+}
+
```
MBkkt, line 64: hnsw option in error message, but function name is GetIntOption, seems wrong to me

server/catalog/index.cpp
```diff
@@ -33,6 +36,142 @@
...
+    return str_value.GetValue<std::string>();
+  }
+  return std::unexpected<Result>{std::in_place,
+                                 ERROR_BAD_PARAMETER,
+                                 "Column '",
+                                 column_name,
+                                 "': hnsw option '",
+                                 key,
+                                 "' must be a string, got '",
+                                 v.ToString(),
```
MBkkt, line 79: same here

tests/sqllogic/sdb/pg/settings/show_all.test
```diff
@@ -186,6 +186,7 @@ scalar_subquery_error_on_multiple_rows	on	When a scalar subquery returns multipl
 scheduler_process_partial	off	Partially process tasks before rescheduling - allows for more scheduler fairness between separate queries
 schema	public	Sets the default search schema. Equivalent to setting search_path to a single value.
 scram_iterations	4096	Sets the iteration count for SCRAM secret generation.
+sdb_ef_search	0	Per-session override for the HNSW search-time neighbourhood size (efSearch). 0 (default) uses Top-K value instead.
```
MBkkt, line 189: what if top-k quite large? does it handles ok?

Btw duckdb starts with some limit, around 10k or 100k, starts to avoid topn and use sort + limit, are we ok with it? Needs to check common queries, rule/etc
Create issue probably?

tests/sqllogic/sdb/pg/index/basic.test
```diff
@@ -234,7 +234,7 @@ db error: ERROR: Failed to create index: Text search dictionary 'test_english' d
 statement error
 CREATE INDEX index_ts on tstest USING inverted(c test_english (foo=bar));
 ----
-db error: ERROR: syntax error at or near "("
+db error: ERROR: Opclass option must be a constant value
```
MBkkt, line 237: quite unclear error, lets do something better to make clear that we support two options

1. `<some_name>`
2. `<some_type> (params ...)` (constant name, probably can output list of possible values?)

Also probably needs to add test with `opclass_type ()` -- empty parenthesis case

server/connector/search_pk_lookup.h
```diff
@@ -0,0 +1,91 @@
...
+void LookupSegmentsValues(const auto& segments, const auto& doc_ids,
+                          const irs::IndexReader& reader, auto& result) {
+  SDB_ASSERT(result.size() == segments.size());
+  SDB_ASSERT(result.size() == doc_ids.size());
+  const size_t n = result.size();
+  std::vector<size_t> idx(n);
+  std::iota(idx.begin(), idx.end(), 0);
+  std::ranges::sort(
+    idx, {}, [&](size_t i) { return std::pair{segments[i], doc_ids[i]}; });
+
```
MBkkt, line 64: looks quite naive and slow

server/connector/optimizer/iresearch_plan.cpp
```diff
@@ -1541,8 +1657,8 @@ ParsedOffsetsCall ParseOffsetsCall(duckdb::BoundFunctionExpression& func,
       "OFFSETS(): column '%s' not found in table", col_name);
   }
   const auto& idx_col_ids = index_entry->GetInvertedIndex()->GetColumnIds();
-  const bool in_index = std::find(idx_col_ids.begin(), idx_col_ids.end(),
-                                  target_col_id) != idx_col_ids.end();
+  const bool in_index =
+    absl::c_find(idx_col_ids, target_col_id) != idx_col_ids.end();
```
MBkkt, line 1661: absl::c_linear_search

server/connector/optimizer/iresearch_plan.cpp
```diff
@@ -2018,8 +2156,20 @@ class IresearchPlanOptimizer : public duckdb::OptimizerExtension {
...
-    bool changed = Walk(plan, plan, /*in_mutation=*/false, /*pass=*/1);
-    changed |= Walk(plan, plan, /*in_mutation=*/false, /*pass=*/2);
+    int ef_search_override = 0;
+    {
+      duckdb::Value v;
+      if (input.context.TryGetCurrentSetting("sdb_ef_search", v) &&
+          !v.IsNull()) {
+        ef_search_override = v.GetValue<int32_t>();
+      }
+    }
```
MBkkt, line 2166: Why do we need to pass this in optimizer rule? Why can't we get it in bind function instead?

server/connector/duckdb_physical_create_index.cpp
```diff
@@ -199,12 +199,17 @@ SereneDBPhysicalCreateIndex::GetGlobalSinkState(
         throw duckdb::CatalogException("column \"%s\" not found in table",
                                        col_name);
       }
+      duckdb::case_insensitive_map_t<duckdb::Value> opclass_options;
+      if (i < _info->column_opclass_options.size()) {
```
MBkkt, line 203: empty check instead + assert? otherwise I think our code is broken for mixed cases

server/connector/duckdb_search_ann_scan.cpp
```diff
@@ -41,88 +46,109 @@
...
   }
 
-  if (!state->ann_pk_bytes.empty()) {
-    state->materializer = MakeRowMaterializer(
-      context, bind_data, state->snapshot, state->ann_pk_bytes,
-      state->projected_columns, state->projected_types, bind_data.column_ids,
-      nullptr);
+  SDB_ASSERT(state.scan->top_k > 0);
+  const size_t top_k = state.scan->top_k;
+  std::vector<float> dis(top_k, std::numeric_limits<float>::max());
```
MBkkt, line 63: I think this one is kind of bad, we probably want to use duckdb flatvector to be able return distances without copy?

server/connector/duckdb_search_ann_scan.cpp
```diff
@@ -41,88 +46,109 @@
...
+  for (const auto& e : _exprs) {
+    _executor.AddExpression(*e);
   }
+  duckdb::vector<duckdb::LogicalType> scratch_types(filter_types.begin(),
+                                                    filter_types.end());
+  _scratch.Initialize(duckdb::Allocator::DefaultAllocator(), scratch_types);
+
+  duckdb::vector<duckdb::LogicalType> bool_types(_exprs.size(),
+                                                 duckdb::LogicalType::BOOLEAN);
+  _bool_out.Initialize(duckdb::Allocator::DefaultAllocator(), bool_types);
```
MBkkt, line 113: we should use query allocator, not default one

server/connector/duckdb_ann_filter.h
```diff
@@ -0,0 +1,63 @@
...
+namespace sdb::connector {
+
+class ANNFilter final : public faiss::IDSelector {
+ public:
+  ANNFilter(duckdb::ClientContext& context, const irs::IndexReader& reader,
+            std::unique_ptr<RowMaterializer> materializer,
+            std::vector<duckdb::unique_ptr<duckdb::Expression>> exprs,
+            std::vector<duckdb::LogicalType> filter_types);
+
+  bool is_member(faiss::idx_t id) const override;
```
MBkkt, line 41:
```suggestion
  bool is_member(faiss::idx_t id) const final;
```
codestyle

server/connector/duckdb_ann_filter.cpp
```diff
@@ -0,0 +1,111 @@
...
+  std::string_view pk = irs::ViewCast<char>(val);
+
+  _scratch.Reset();
+  std::array<std::string_view, 1> pks{pk};
+  _materializer->Materialize(pks, _scratch);
+  _scratch.SetCardinality(1);
+
+  _bool_out.Reset();
+  _bool_out.SetCardinality(1);
+  _executor.Execute(_scratch, _bool_out);
```
MBkkt, line 96: this looks quite slow, do we have issue/plans how to improve this?

server/connector/duckdb_search_range_scan.cpp
```diff
@@ -40,83 +43,79 @@
...
+  auto& snapshot =
+    GetSereneDBContext(context).EnsureSearchSnapshot(state.scan->index_id);
+  auto& reader = snapshot.reader;
+
+  if (reader.size() == 0) {
+    state.finished = true;
+    return;
+  }
+  std::vector<float> dis;
+  std::vector<int64_t> ids;
```
MBkkt, line 59: same

server/connector/duckdb_search_range_scan.cpp
```diff
@@ -40,83 +43,79 @@
...
+  reader.RangeSearch(state.scan->field_name, info, dis, ids);
+
+  auto segments =
+    ids | std::views::transform([](int64_t id) {
+      return irs::UnpackSegmentWithDoc(static_cast<uint64_t>(id)).first;
+    });
+  auto doc_ids =
+    ids | std::views::transform([](int64_t id) {
+      return irs::UnpackSegmentWithDoc(static_cast<uint64_t>(id)).second;
+    });
```
MBkkt, line 75: Are we sure that we want to use such code with ranges in hot-path?

server/connector/duckdb_table_function.cpp
```diff
@@ -300,13 +300,33 @@ void ANNScan::AppendSummary(
...
+  if (!filter_expressions.empty()) {
+    std::string summary;
+    for (const auto& expr : filter_expressions) {
+      if (!summary.empty()) {
+        summary += " AND ";
+      }
+      summary += expr->ToString();
+    }
+    out.insert("Filter", summary);
+  }
```
MBkkt, line 312: reuse this code

server/connector/duckdb_table_function.h
```diff
@@ -277,18 +277,25 @@ struct SecondaryIndexScan : ScanSource {
...
-  ANNScan() : ScanSource(ScanSourceKind::Ann) {}
+struct ANNScan : VectorSearchScan {
+  ANNScan() : VectorSearchScan{ScanSourceKind::Ann} {}
 
-  ObjectId index_id;
-  // Field name: big-endian catalog::Column::Id bytes, no MangleString
-  std::string field_name;
-  std::vector<float> query_vector;
   size_t top_k = 0;
+  int ef_search = 0;
```
MBkkt, line 298: ef_search probably make sense for range scan too?

server/connector/optimizer/iresearch_plan.cpp
```diff
@@ -178,6 +180,9 @@ std::optional<irs::HNSWMetric> DistanceMetricForFunction(
       name == connector::kCosineDistanceOp) {
     return irs::HNSWMetric::Cosine;
   }
+  if (name == connector::kInnerProduct || name == connector::kIPDistanceOp) {
+    return irs::HNSWMetric::InnerProduct;
```
MBkkt, line 184: `name == connector::kInnerProduct`
innerproduct doesn't look correct here, because it's similarity, so for similarity we need DESC not ASC order, opposite to distance

server/connector/duckdb_ann_filter.cpp
```diff
@@ -0,0 +1,111 @@
...
+  const std::vector<duckdb::unique_ptr<duckdb::Expression>>& filter_expressions,
+  const std::vector<catalog::Column::Id>& filter_column_ids, ObjectId index_id,
+  const rocksdb::Snapshot* rocks_snapshot,
+  const SereneDBScanBindData& bind_data) {
+  std::vector<duckdb::idx_t> filter_projection(filter_column_ids.size());
+  std::vector<duckdb::LogicalType> filter_types(filter_column_ids.size());
+  for (size_t i = 0; i < filter_column_ids.size(); ++i) {
+    filter_projection[i] = i;
+    const auto cat_id = filter_column_ids[i];
+    const auto it = absl::c_find(bind_data.column_ids, cat_id);
```
MBkkt, line 42: this and other find doesn't look good, we probably should have kind of hashmap, like we had for velox, from column id to column idx and reverse

server/connector/duckdb_ann_filter.cpp
```diff
@@ -0,0 +1,111 @@
...
+  _scratch.Reset();
+  std::array<std::string_view, 1> pks{pk};
+  _materializer->Materialize(pks, _scratch);
+  _scratch.SetCardinality(1);
+
+  _bool_out.Reset();
+  _bool_out.SetCardinality(1);
+  _executor.Execute(_scratch, _bool_out);
+
+  for (duckdb::idx_t i = 0; i < _bool_out.ColumnCount(); ++i) {
```
MBkkt, line 98: why > 1? it seems like all filters can be combined into single AND

server/connector/duckdb_search_ann_scan.cpp
```diff
@@ -41,88 +46,109 @@
...
+      return irs::UnpackSegmentWithDoc(static_cast<uint64_t>(id)).first;
+    });
+  auto doc_ids =
+    ids | std::views::transform([](int64_t id) {
+      return irs::UnpackSegmentWithDoc(static_cast<uint64_t>(id)).second;
+    });
+
+  state.pk_bytes.assign(n, std::string{});
+  LookupSegmentsValues(segments, doc_ids, reader, state.pk_bytes);
+  std::erase_if(state.pk_bytes, [](const auto& pk) { return pk.empty(); });
```
MBkkt, line 90: looks slow

MBkkt (commented):
Also should be way more tests:
1. different opclass cases, with tokenizer/etc -- seems like it's broken now, but I'm not sure
2. DML should check count `statement count <N>` instead of `statement ok`
3. needs to add tests that ensures we don't project columns that used only for ann filters
4. just more tests plz, seems completely untested, e.g. more than single vector docs in limit, more than single batch docs in faiss, check recall, etc

PR #568, opened 2026-04-23
fix: Fix bind variables

MBkkt:


server/pg/pg_comm_task.cpp
```diff
@@ -536,13 +546,63 @@ void DescribeParameters(const DuckDBStatement& stmt, message::Buffer& buffer) {
...
+    return;
+  }
+
+  duckdb::vector<duckdb::Value> dummy;
+  dummy.reserve(nparams);
+  for (size_t i = 0; i < nparams; ++i) {
+    const auto oid =
+      i < stmt.param_oids.size() ? stmt.param_oids[i] : int32_t{0};
+    auto type = oid != 0 ? Oid2Type(oid) : duckdb::LogicalType::VARCHAR;
+    duckdb::Value v{"1"};
```
MBkkt, line 594: 1 instead of 0 to avoid cases like `1/$1`

pashandor789 (approved):

PR #503, opened 2026-04-05
refactor: allow only single wand per posting list

afigor2701:

PR #573, opened 2026-04-26
feat: Enable wand for inverted index

gnusi:
BM25 top-K queries currently iterate every matching posting. iresearch already has a WAND implementation that lets the iterator skip postings whose max-impact is below the running heap minimum; this PR exposes it as an opt-in catalog flag and wires it into the server's top-K scan path.

```sql
CREATE INDEX i ON t USING inverted(col) WITH (optimize_top_k = 'bm25(1.2, 0.75)');
```

The string accepts any registered scorer (`bm25`, `tfidf`, `raw_tf`, `lm_jm`, `lm_dirichlet`, `indri_dirichlet`, `dfi`) with the same overload resolution as `ORDER BY BM25(...)`. WAND engages only when the persisted scorer matches the `ORDER BY` scorer; EXPLAIN renders `, optimized` on the `IRESEARCH_SCAN` row when it does.

Session toggle `SET sdb_disable_top_k_optimization = true` skips the pull-up. Default false.

Also fixes a regression introduced by #503 that hardcoded the WAND dispatch gate to `false`, leaving `WandIterator` / `SingleWandIterator` as dead code, and cleans up the surrounding `IteratorOptions` / `IteratorFieldOptions` / `WandContext` confusion left after that refactor.

## Benchmarks

500k-row corpus, varied document lengths, `SELECT 1` to strip materialization, median of 7 runs (Release build):

| Query | LIMIT | WAND | Plain | Speedup |
|---|---|---|---|---|
| Single 100%-frequency term | 10 | 0.36 ms | 1.35 ms | **3.8x** |
| Single 80%-frequency term | 100 | 0.43 ms | 0.88 ms | **2.0x** |
| ts_any, skewed (common+rare) | 100 | 0.60 ms | 1.14 ms | **1.9x** |
| ts_any, uniform 3-term | 100 | 1.03 ms | 1.35 ms | 1.3x |
| ts_any, common+rare+fox | 100 | 0.99 ms | 2.10 ms | **2.1x** |


MBkkt (commented):
if tests has sdb_faults they should be recovery tests, because recovery tests don't run different tests in parallel on single instance, fix plz

MBkkt (approved):

PR #575, opened 2026-04-27
feat: GEO S2 search support

Dronplane:
Introduces GEO functions ST_Contains, ST_Intersects, ST_Distance_Between, ST_Distance_Centroid backed by inverted index. ST_Distance_Centroid has operator alias "<->"

All functions accepts geo data in two forms - GeoJSON (as JSON type) and WKT (as GEOMETRY type). As we are backed by S2 library we accept only OGC:CRS84 2 dimensional GEOMETRY.

For indexing geo data geojson and geopoint tokenizers were added.

Limitations (enforced at index creation time):
geopoint could be used only with JSON column (as there is no json path in GEOMETRY)
lat/lon encodings for shape now supported only for JSON columns (adding for GEOMETRY is a followup)
geo column does not contributes to score but still could be part of the filter.
 
Follow ups:
1. Implement ST_Distance_Between, ST_Distance_Centroid as free functions
2. Enable Spatial extension to get other functions.
3. Move geometry tokenizers from text dictionary to opclass?
4. Enable lat/lon encoding for geometry?
5. S2 shape type (BLOB with alias)?


libs/iresearch/include/iresearch/utils/utf8_utils.hpp
```diff
@@ -195,4 +198,66 @@ inline bool ToUTF32(bytes_view data, OutputIterator&& out) {
   return true;
 }
 
+// Verify that |str| is well-formed UTF-8: every byte is part of a sequence
+// whose lead byte declares a valid 1..4 byte length, the bytes don't run off
+// the end, and continuation bytes have the form 10xxxxxx.
+inline bool IsValid(std::string_view str) noexcept {
```
MBkkt: this is slow and needs to be removed, we can use code from simdjson (or if really needed add simdutf)

libs/iresearch/include/iresearch/utils/utf8_utils.hpp
```diff
@@ -195,4 +198,66 @@ inline bool ToUTF32(bytes_view data, OutputIterator&& out) {
...
+// Bytes that aren't part of a valid sequence (lone 0x80..0xFF, truncated
+// multi-byte sequences, etc.) are mapped to their own Unicode codepoint via
+// Latin-1: bytes 0x80..0xFF expand to the two-byte UTF-8 sequence
+// (0xC0 | (b >> 6), 0x80 | (b & 0x3F)).
+//
+// Useful for embedding tokenizer output into UTF-8-only string types -- e.g.
+// the wildcard / ngram analyzers' word-boundary sentinel byte (0xFF) is
+// invalid as a standalone UTF-8 byte; this routine converts it (and any
+// other invalid bytes) into a stable, reversible UTF-8 form.
+inline std::string ToUtf8Safe(std::string_view bytes) {
```
MBkkt: This is slow and needs to be completely removed

server/catalog/index.cpp
```diff
@@ -46,6 +51,70 @@ Result ValidateInvertedIndexColumns(
...
 }
 
+// Validate that a geo-family analyzer (GeoJsonAnalyzer / GeoPointAnalyzer) is
+// compatible with the column it's bound to. Runs once per column at CREATE
+// INDEX time.
+//
+// Rules:
+//   - Column must be VARCHAR (GeoJSON text -- validated by the JSON parser at
+//     insert time) or GEOMETRY (strongly typed, CRS declared at the column
+//     level). BLOB is rejected: we have no way to confirm its bytes are WKB.
```
MBkkt: I think we want to check it json? Why do we want to allow any varchar?

Dronplane: We will add check for JSON alias.

libs/iresearch/include/iresearch/utils/utf8_utils.hpp
Dronplane: Another approach was taken - this part removed as not needed

server/connector/search_filter_printer.cpp
```diff
@@ -259,24 +260,6 @@ void StringifyWildcard(std::string* out, const ByWildcard& filter, FT&& ft) {
...
-void StringifyPhrase(std::string* out, const ByPhrase& filter, FT&& ft) {
-  std::string parts_str;
-  for (const auto& part : filter.options()) {
-    std::string part_str;
-    part.part.visit(PhrasePartVisitor{.out = &part_str});
-    absl::StrAppend(&parts_str, part_str, "(", part.offs_max, ", ",
-                    part.offs_min, ")", "; ");
-  }
-  absl::StrAppend(out, "PHRASE[", ft(filter.field()), " = <", parts_str, ">]");
-}
```
MBkkt, line 278: why we moved this?

server/connector/optimizer/iresearch_plan.cpp
```diff
@@ -146,43 +146,55 @@ std::string MakeHnswFieldName(catalog::Column::Id col_id) {
...
+    return std::nullopt;
   }
-  if (name == connector::kL2Norm) {
-    return ExpectedHNSW{irs::HNSWMetric::L2Sqr, duckdb::OrderType::ASCENDING};
+  if (result.is_norm ? func.children.size() != 1 : func.children.size() < 2) {
+    return std::nullopt;
   }
-  if (name == connector::kL1Norm) {
-    return ExpectedHNSW{irs::HNSWMetric::L1, duckdb::OrderType::ASCENDING};
+  if (func.children[0]->return_type.id() != duckdb::LogicalTypeId::ARRAY) {
```
MBkkt: it's doesn't work linestring (array<struct<double, double>>)

MBkkt: list also should be supported by hnsw

server/catalog/index.cpp
```diff
@@ -251,6 +256,81 @@ Result ValidateInvertedIndexColumns(
...
+//     to confirm its bytes are WKB.
+//   - For GEOMETRY columns, the declared CRS must be CRS84 (EPSG:4326 /
+//     OGC:CRS84 / SRID 4326). The sink-writer path does no per-row SRID
+//     check, so the column declaration is the contract.
+//   - For GEOMETRY + GeoJsonAnalyzer: coding must be S2Point. LatLng codings
+//     would require a shape -> LatLng-bytes encoder that ShapeContainer
+//     doesn't implement yet -- reject to avoid silent data loss at read
+//     time. (VPack coding is rejected at dictionary creation time, so it
+//     can't reach this check via SQL.)
+Result ValidateGeoAnalyzerColumn(std::string_view column_name,
```
MBkkt: tokenizer

MBkkt (approved):

server/catalog/index.cpp
Dronplane: Done

server/connector/search_filter_printer.cpp
Dronplane: Fixed

server/connector/optimizer/iresearch_plan.cpp
Dronplane: Added check for array/list elements type

PR #580, opened 2026-04-27
feat: FTS sugar

gnusi:
# FTS surface

## Match operator

| Form | Type | Behaviour |
|---|---|---|
| `col @@ tsquery` | `(ANY, TSQUERY) → BOOLEAN` | Filter builder claims at bind time. Commutative — `'q' @@ col` also works. |
| `col_a @@ col_b` | rejected | Bind error: ambiguous column position. |

## TSQUERY constructors

| Constructor | Signature | Semantics | iresearch filter |
|---|---|---|---|
| `'literal'` (bare string) | `VARCHAR → TSQUERY` (cast) | Tokenises via column analyzer; multi-token → ANY_OF. | `ByTerm` / `ByTerms{min_match=1}` / `Empty` |
| `ts_phrase(text)` | `VARCHAR → TSQUERY` | Tokenises text, builds adjacent phrase parts. Requires Pos+Freq features. | `ByPhrase` |
| `ts_like(pattern)` | `VARCHAR → TSQUERY` | Wildcard match (`%`/`_`); raw, not tokenised. Wildcard-analyzer columns route to ngram-aware filter. | `ByWildcard` / `ByWildcardNgram` |
| `ts_starts_with(text)` | `VARCHAR → TSQUERY` | Prefix match (raw, not tokenised). | `ByPrefix` |
| `ts_ngram(text, threshold)` | `(VARCHAR, DOUBLE) → TSQUERY` | N-gram similarity against tokenised input. Requires Pos+Freq. | `ByNGramSimilarity` |
| `ts_levenshtein(term, dist [, transpositions [, prefix]])` | `(VARCHAR, INT [, BOOL [, VARCHAR]]) → TSQUERY` | Edit-distance match. Optional `prefix` is a literal leading substring that must match exactly; only the suffix beyond it participates in edit-distance computation. | `ByEditDistance` |
| `ts_regexp(pattern [, syntax])` | `(VARCHAR [, VARCHAR]) → TSQUERY` | Regex against indexed terms. `syntax` ∈ `{'perl','posix'}`. | `ByRegexp` |
| `ts_between(min, max, min_incl, max_incl)` | `(ANY, ANY, BOOL, BOOL) → TSQUERY` | Range filter. NULL bound = unbounded that side. VARCHAR/BOOLEAN→`ByRange`, numeric→`ByGranularRange`. | `ByRange` / `ByGranularRange` |
| `ts_lt / ts_le / ts_gt / ts_ge(bound)` | `ANY → TSQUERY` | Single-bound shortcut for `ts_between(...)`. | `ByRange` / `ByGranularRange` |
| `ts_any(q1, q2, ...)` (variadic) | `TSQUERY... → TSQUERY` | OR of tsqueries. All-terms path → `ByTerms` with `min_match=1`. | `Or` / `ByTerms` |
| `ts_any([list])` | `LIST(TSQUERY) → TSQUERY` | Same as variadic but takes a list. | `Or` / `ByTerms` |
| `ts_any([list], min_match)` | `(LIST(TSQUERY), INT) → TSQUERY` | At least N of the list must match. | `Or{min_match=N}` / `ByTerms{min_match=N}` |
| `ts_all(q1, q2, ...)` / `ts_all([list])` | variadic / `LIST(TSQUERY) → TSQUERY` | AND of tsqueries. | `And` / `ByTerms{min_match=count}` |
| `ts_tokenize(text)` | `VARCHAR → TSQUERY` | Same as bare string — tokenises via column analyzer. | `ByTerm` / `ByTerms` / `Empty` |
| `ts_tokenize(text, name)` | `(VARCHAR, VARCHAR) → TSQUERY` | `'identity'`: raw `ByTerm`. Other names: catalog lookup, override per leaf. | `ByTerm` / `ByTerms` |
| `ts_tokenize(text_array)` | `LIST(VARCHAR) → LIST(TSQUERY)` | Array form, ambient analyzer. Use inside `ts_any` / `ts_all`; tokens flatten. | aggregated `ByTerms` |
| `ts_tokenize(text_array, name)` | `(LIST(VARCHAR), VARCHAR) → LIST(TSQUERY)` | Array form with override; `'identity'` for raw, otherwise catalog lookup. | aggregated `ByTerms` |
| `to_tsquery(string)` | `VARCHAR → TSQUERY` | Lucene-style query parser: `AND`/`OR`/`NOT`, `term~N` (fuzzy / slop), `term^N` (boost), `*` / `?` wildcards, `[a TO b]` ranges, phrases. Each parsed term tokenises through the column's ambient analyzer; constant input only. | composed `MixedBooleanFilter` |
| `plainto_tsquery(text)` | `VARCHAR → TSQUERY` | Tokenise + AND. | `ByTerms{min_match=count}` |
| `phraseto_tsquery(text)` | `VARCHAR → TSQUERY` | Same body as `ts_phrase(text)`. | `ByPhrase` |
| `websearch_to_tsquery(text)` | `VARCHAR → TSQUERY` | Web-search syntax: `"phrase"`, `OR`, `-token`. | composed tree |
| `tsquery_phrase(q1, q2 [, gap])` | `(TSQUERY, TSQUERY [, INT]) → TSQUERY` | Function form of `q1 ## [gap] ## q2`. | `ByPhrase` |

## TSQUERY operators

| Operator | Signature | Semantics |
|---|---|---|
| `\|\|` | `(TSQUERY, TSQUERY) → TSQUERY` | OR. Bare-string operands resolve to VARCHAR concat (lower cast cost) — wrap one side in a TSQUERY constructor or per-leaf `::tokenize(...)` to force the TSQUERY overload. |
| `&&` | `(TSQUERY, TSQUERY) → TSQUERY` | AND. Same overload-resolution caveat as `\|\|`. |
| `!!` | `TSQUERY → TSQUERY` | NOT (prefix). |
| `^` | `(TSQUERY, DOUBLE) → TSQUERY` | Boost factor. Tighter precedence than `Op` tier — no parens needed inside `@@`. |
| `##` | `(TSQUERY, TSQUERY) → TSQUERY`, `(TSQUERY, INTEGER, ...) → TSQUERY`, `(TSQUERY, INTEGER[], ...) → TSQUERY` | Phrase sequence. Bare-string operands accepted (each tokenised). Gap = `INT` exact, `INT[]` for `[lo, hi]` interval. |

## Composable casts (`::tokenize(...)`, `::boost(K)`)

Both casts are parametric-type modifiers on the TSQUERY alias family
(`TSQUERY` / `TOKENIZED_TSQUERY` / `BOOSTED_TSQUERY`). DuckDB carries
them as `BoundCastExpression`s with a `LogicalType` modifier; the
filter builder reads the modifier off the cast at dispatch time.

| Form | Effect |
|---|---|
| `'text'::tokenize('analyzer')` | Override column analyzer with named catalog tokenizer for this VARCHAR operand. |
| `'text'::tokenize('identity')` | Skip tokenisation; emit raw `ByTerm` with the literal bytes. |
| `'text'::tokenize(NULL)` | Sugar for `::tokenize('identity')`. |
| `ts_tokenize(text [, name])` / `ts_tokenize(text_array [, name])` | Function spellings. Same resolution as the cast — `'identity'` for raw, otherwise catalog lookup. |
| `expr::boost(K)` | Multiplies the inherited boost by `K` (positive `DOUBLE`). Composable analog of the `^` operator — works on any TSQUERY-typed operand (literals, constructors, sub-trees) **and on BOOLEAN-typed SQL predicates outside `@@` that the inverted-index claimer accepts**. Stacks multiplicatively when nested. |
| `(tsquery_expr)::tokenize(...)` | **No-op** — DuckDB elides TSQUERY → TSQUERY-with-modifier casts. Apply per-leaf instead. |
| Chained `'x'::tokenize('a')::tokenize('b')` | Outermost wins (DuckDB folds the inner cast). Don't chain. |
| Unknown analyzer name | `THROW_SQL_ERROR(ERRCODE_UNDEFINED_OBJECT, "tokenizer not found in catalog")` with a `HINT` suggesting `CREATE TEXT SEARCH DICTIONARY` or `'identity'`. |
| Negative `::boost(K)` factor | Rejected at filter-build time (`boost factor must be >= 0`). |

## Bare-string semantics inside `@@`

| Bare-string form | Tokenisation |
|---|---|
| `col @@ 'word'` | Column analyzer (default ambient) → ANY_OF tokens. |
| `col @@ ('a' \|\| 'b')` | VARCHAR concat first → single string → tokenised once. To get TSQUERY OR, wrap each side in `ts_tokenize(...)` or `::tokenize(...)`. |

## SQL-native predicates against indexed columns

| Predicate | Identity-analyzer column | Non-identity analyzer |
|---|---|---|
| `col = 'foo'` | `ByTerm` | error → use `col @@ 'foo'` |
| `col < / <= / > / >= 'foo'` | `ByRange` | error → use `col @@ ts_lt('foo')` / `ts_le` / `ts_gt` / `ts_ge` / `ts_between(...)` |
| `col LIKE 'foo%'` | `ByWildcard` | error (or `ByWildcardNgram` for wildcard analyzer) → use `col @@ ts_like('foo%')` |
| `col IN (...)` | `ByTerms` | error → use `col @@ ts_any(...)` |
| `col ~ / !~ / ~* / !~* 'pat'` | regex via `ByRegexp` claim path | same |

## Scoring functions (claimed at bind time, projected as columns)

| Function | Returns | Notes |
|---|---|---|
| `bm25(tableoid [, k1, b])` | `FLOAT` | k1, b are constants. |
| `tfidf(tableoid [, with_norms])` | `FLOAT` | |
| `raw_tf(...)`, `lm_jm`, `lm_dirichlet`, `indri_dirichlet`, `dfi` | `FLOAT` | |
| `sdb_offsets(col)` | `BIGINT[]` | Token offsets. |

## Errors / hints

Every TSQUERY-surface failure inside `@@` (malformed constructor args,
type mismatches, NULL where a non-NULL constant is required, unsupported
expression class, etc.) throws `THROW_SQL_ERROR(ERRCODE_INVALID_PARAMETER_VALUE)`
with a `HINT` showing the constructor's syntax and a small example.

The hint is constructor-specific — `ts_phrase`, `ts_like`, `ts_levenshtein`,
`ts_between`, `ts_compound`, `ts_any` / `ts_all`, `to_tsquery` family,
`&&` / `||` / `!!` / `^` / `##` operators, `::tokenize(...)` cast,
`::boost(K)` cast — so users get an actionable redirect rather than the
generic "expression evaluated outside @@ context" stub.

```text
ERROR:  LEVENSHTEIN expects 1 to 4 arguments (text, distance?, transpositions?, prefix?), got 5
HINT:   Example: ts_levenshtein('test', 1, true, 'pre'). Distance must be 0-4
        (0-3 with transpositions). Optional `prefix` matches exactly.
```

## Session knobs

| Setting | Default | Effect |
|---|---|---|
| `sdb_scored_terms_limit` | 1024 | Caps terms collected for scoring in `ts_starts_with` / `ts_like` / `ts_between` / `ts_regexp` / `ts_levenshtein`. 0 disables scored-term collection. |
| `sdb_ef_search` | 0 | HNSW efSearch override. 0 = use Top-K value. |

### Bare-string match (default: column analyzer)

```sql
-- column `b` indexed with text+lower analyzer
WHERE b @@ 'Quick Brown Fox'
  -- analyzer lowercases + segments -> tokens: [quick, brown, fox]
  -- emitted: ByTerms{[quick,brown,fox], min_match=1}  (any of the tokens)

-- The query above is exactly equivalent to:
WHERE b @@ ts_any('Quick', 'Brown', 'Fox')

-- ALL_OF semantics need plainto_tsquery:
WHERE b @@ plainto_tsquery('Quick Brown')
  -- both 'quick' AND 'brown' must match
```

### Phrase / proximity / boost

```sql
-- exact phrase
WHERE b @@ ts_phrase('quick brown fox')

-- proximity gap: up to 2 intervening tokens
WHERE b @@ 'quick' ## 2 ## 'fox'

-- gap range [1, 3]
WHERE b @@ 'quick' ## [1, 3] ## 'fox'

-- function form
WHERE b @@ tsquery_phrase(ts_phrase('hello'), ts_phrase('world'), 3)

-- boost (tighter precedence than @@ / || / &&)
WHERE b @@ ts_phrase('quick fox') ^ 2.5
ORDER BY bm25(tableoid)
```

### Term sets / boolean composition

```sql
-- variadic and list forms are equivalent
WHERE b @@ ts_any('foo', 'bar', 'baz')
WHERE b @@ ts_any(['foo', 'bar', 'baz'])

-- 2-of-N
WHERE b @@ ts_any(['foo', 'bar', 'baz'], 2)

-- AND of multiple constructors
WHERE b @@ (ts_phrase('quick brown') && ts_like('fox%'))

-- mixed boolean tree
WHERE b @@ (ts_phrase('quick') || (ts_levenshtein('brwon', 1) && !!ts_phrase('lazy')))
```

### Levenshtein (fuzzy match) with prefix

```sql
-- vanilla fuzzy match (distance 1, transpositions on by default)
WHERE b @@ ts_levenshtein('brwn', 1)

-- without transpositions: 'cat' vs 'act' is now distance 2, not 1
WHERE b @@ ts_levenshtein('cat', 1, false)

-- prefix-anchored fuzzy: literal 'app' must match exactly; only the
-- suffix is fuzzy-compared against 'le' within distance 1.
-- Matches 'apple' (dist 0), 'apply' (dist 1).
-- Excludes 'apricot' (no full 'app' prefix), 'application' (suffix too far).
WHERE b @@ ts_levenshtein('le', 1, true, 'app')

-- empty prefix is equivalent to the 3-arg form
WHERE b @@ ts_levenshtein('apple', 1, true, '')
```

### Range / comparisons

```sql
-- analyzer-aware single bound
WHERE b @@ ts_le('quick')

-- numeric column (granular range under the hood)
WHERE i @@ ts_between(10, 100, true, false)

-- unbounded one side
WHERE b @@ ts_between(NULL, 'mango', false, true)
```

### `::boost(K)` cast (composable per-leaf boost)

`::boost(K)` is a parametric-type cast that multiplies the inherited
boost by `K`. It composes with the `^` operator and stacks
multiplicatively when nested. Unlike `^` (which lives at operator
precedence), the cast attaches directly to any TSQUERY-typed operand.

```sql
-- equivalent to ts_phrase('quick fox') ^ 2.0
WHERE b @@ ts_phrase('quick fox')::boost(2.0)

-- per-leaf boost inside a boolean tree
WHERE b @@ (ts_phrase('quick fox')::boost(3.0)
         || ts_phrase('lazy dog')::boost(0.5))
ORDER BY bm25(tableoid) DESC

-- stacks multiplicatively (effective factor = 2 * 1.5 = 3.0)
WHERE b @@ (ts_phrase('foo')::boost(2.0))::boost(1.5)

-- works on bare-string literals too
WHERE b @@ ('priority'::tokenize('lower_text'))::boost(5.0)

-- negative factor rejected
WHERE b @@ ts_phrase('foo')::boost(-1.0)
-- ERROR: boost factor must be >= 0, got -1
-- HINT:  Example: ts_phrase('text') ^ 2.0. The factor must be a
--        non-negative numeric constant. For composable boosts use ::boost(K).
```

`::boost(K)` also works **outside `@@`** on plain SQL predicates that
the inverted-index optimizer can claim. The cast wraps a BOOLEAN
predicate, the boost peel runs at the SQL-surface entry point, and
the factor flows into the resulting iresearch filter:

```sql
-- boost a numeric range
WHERE (b > 50)::boost(2.0)

-- per-leg boost inside an OR
WHERE (a = 'priority')::boost(3.0)
   OR (a = 'normal')

-- boost a LIKE prefix (DuckDB rewrites to prefix(); claimed via ByPrefix)
WHERE (sku LIKE 'alpha-%')::boost(1000.0)

-- nested casts multiply (effective 1.0)
WHERE ((b > 50)::boost(2.0))::boost(0.5)

-- negation preserves boost
WHERE NOT (b > 50)::boost(2.0)

-- bind-time error: predicate the inverted index can't claim
WHERE (id > id)::boost(2.0)
-- ERROR: ::boost(K) used on a predicate the inverted index could not
--        claim
-- HINT:  boost is only meaningful inside an inverted-index match.
--        Move the boost into an `@@` match or remove it.
```

DuckDB's optimizer rewrites `col LIKE 'pre%'` to `prefix(col, 'pre')`
at bind time; the filter builder claims `prefix()` directly and emits
`irs::ByPrefix` on identity-analyzed columns (matches `ts_starts_with`
semantics). Other LIKE rewrites (`contains()` for `LIKE '%foo%'`) are
intentionally not claimed — use `ts_like('%foo%')` explicitly for
contains semantics.

### Tokenizer override (per-leaf only)

```sql
-- column 'b' uses lowercase analyzer; this leaf forces uppercase
WHERE b @@ 'Quick'::tokenize('upper_text')
  -- 'QUICK' against indexed lowercase tokens -> 0 rows

-- raw bytes (skip tokenisation entirely)
WHERE b @@ 'exact-sku-42'::tokenize('identity')
WHERE b @@ 'exact-sku-42'::tokenize(NULL)        -- NULL is sugar

-- function spelling -- equivalent to the cast (catalog-resolved)
WHERE b @@ ts_tokenize('Quick', 'upper_text')

-- per-leg override (compound expressions cast at the LEAVES, not the root)
WHERE b @@ ('quick'::tokenize('lower_text') || 'BROWN'::tokenize('upper_text'))

-- unknown analyzer name -> bind error
WHERE b @@ ts_tokenize('foo', 'nonexistent')
-- ERROR: ts_tokenize(text, 'nonexistent'): tokenizer not found in catalog
-- HINT:  Create it via CREATE TEXT SEARCH DICTIONARY or use 'identity' for raw bytes.
```

### Array form: tokenise N strings, flatten into one filter

```sql
-- column-ambient (1-arg)
WHERE b @@ ts_any(ts_tokenize(['Quick brown', 'lazy dog']))
  -- tokens flatten: [quick, brown, lazy, dog] under ANY_OF semantics

-- ALL_OF -> every produced token must match
WHERE b @@ ts_all(ts_tokenize(['Quick brown', 'lazy dog']))
  -- requires all of: quick, brown, lazy, dog

-- min_match
WHERE b @@ ts_any(ts_tokenize(['quick', 'brown', 'slow'], 'identity'), 2)
  -- at least 2 of [quick, brown, slow] must match

-- named analyzer (catalog-resolved)
WHERE b @@ ts_any(ts_tokenize(['Quick', 'BROWN'], 'lower_text'))

-- raw bytes per element
WHERE b @@ ts_any(ts_tokenize(['exact-sku-42', 'literal-foo'], 'identity'))

-- NULL elements skipped
WHERE b @@ ts_any(ts_tokenize(['quick', NULL, 'red'], 'identity'))
```

### SQL-native predicates (identity-analyzer columns only)

```sql
-- works on identity-analyzed columns
WHERE b = 'foo'              -- ByTerm
WHERE b LIKE 'foo%'          -- ByWildcard
WHERE b BETWEEN 'a' AND 'm'  -- ByRange
WHERE b IN ('foo', 'bar')    -- ByTerms
WHERE name ~ 'foo.*'         -- ByRegexp via DuckDB's regex_range_filter

-- on a non-identity (e.g. text/lower) column, these throw with a
-- redirect hint -- use the TSQUERY surface instead:
--    `col @@ 'foo'`
--    `col @@ ts_like('foo%')`
--    `col @@ ts_between('a', 'm', true, true)`
--    `col @@ ts_any('foo', 'bar')`
```

### Scoring + session knobs

```sql
-- top-K by BM25
SELECT a, bm25(tableoid) AS score
FROM tsq_idx
WHERE b @@ 'quick brown fox'
ORDER BY score DESC
LIMIT 10;

-- cap the per-filter scored-term collection (default 1024)
SET sdb_scored_terms_limit = 256;
```

### PG-compat tsquery family

```sql
WHERE b @@ to_tsquery('quick AND brown OR "lazy dog"')  -- Lucene parser
WHERE b @@ to_tsquery('term^2 AND fuzzy~1')             -- boost + Levenshtein
WHERE b @@ to_tsquery('[a TO m] AND prefix*')           -- range + prefix wildcard
WHERE b @@ plainto_tsquery('quick brown')               -- AND of tokens
WHERE b @@ phraseto_tsquery('quick brown')              -- phrase of tokens
WHERE b @@ websearch_to_tsquery('"quick fox" -slow')    -- web-search
```

#### Mixed-part phrase (`##` with `ts_like` / `ts_levenshtein` / `ts_any` / `ts_between`)

`##` accepts heterogeneous phrase parts. Each part can be a bare term, `ts_like`/`ts_starts_with` wildcard, `ts_levenshtein` fuzzy, nested `ts_phrase`, `ts_any` (treated as a `ByTerms` slot), or a single-bound `ts_between`. Gaps may sit between any two parts. Useful for "approximate phrase" patterns where you want some flexibility per slot:

```sql
-- "<wildcard> <fuzzy> <any-of>"  with at most one intervening word between
-- the wildcard and the fuzzy slot.
WHERE b @@ (ts_like('quick%') ## 1 ## ts_levenshtein('brwn', 1) ## ts_any('fox', 'dog'))

-- nested PHRASE inside `##`: "[hello world] ... [over the lazy dog]"
-- with a 2-3 word gap between the two sub-phrases.
WHERE b @@ (ts_phrase('hello world') ## [2, 3] ## ts_phrase('over the lazy dog'))

-- "<term>  <term>  <single-bound range>"  -- the last slot matches any
-- token whose byte-encoded form is <= 'h'. Useful with identity-analyzed
-- columns where you want lexicographic suffix bounds in a phrase.
WHERE b @@ ('quick' ## 'brown' ## ts_le('h'))

-- function form for a 2-part phrase with a fixed gap
WHERE b @@ tsquery_phrase(ts_like('quick%'), ts_levenshtein('fox', 1), 2)
```

#### `ts_regexp`

`ts_regexp` matches *indexed terms* against the pattern (per-token, not whole-document). `syntax` is `'perl'` (default RE2) or `'posix'` (POSIX ERE):

```sql
-- match any indexed term ending in -tion
WHERE b @@ ts_regexp('.*tion')

-- POSIX ERE dialect
WHERE b @@ ts_regexp('foo|bar|ba[zq]', 'posix')

-- combine with a term and a phrase under @@ semantics
WHERE b @@ ((ts_phrase('quick fox') && ts_regexp('jumps?')) || ts_levenshtein('lazy', 1))

-- SQL-native regex on identity-analyzer columns is also claimed
-- (DuckDB's regex_range_filter rewrites it into a coarse range
-- pre-filter + a runtime regex check).
WHERE name ~ '^foo[0-9]+$'
WHERE name !~* '.*deprecated.*'
```

#### `ts_like` / wildcard

`ts_like` accepts SQL `%` / `_` wildcards; `ts_starts_with` is the common case (`prefix%`) without per-character cost. On `WildcardAnalyzer` columns the filter routes to `ByWildcardNgram` (ngram-aware, faster on large term dictionaries):

```sql
WHERE b @@ ts_like('quic%fox')             -- '%' matches any sequence
WHERE b @@ ts_like('te_t')                 -- '_' matches one char ("test","text",...)
WHERE b @@ ts_like('100\%off' ESCAPE '\')  -- (cast equivalent: see ESCAPE on SQL LIKE)
WHERE b @@ ts_starts_with('quick')              -- equivalent to ts_like('quick%') but cheaper

-- combine: LIKE plus a fuzzy term
WHERE b @@ (ts_like('quic%') && ts_levenshtein('brwn', 1))

-- SQL-native LIKE on identity-analyzed columns
WHERE b LIKE 'foo%'
```

#### NGRAM similarity

`ts_ngram(text, threshold)` runs the indexed n-gram tokens against `text`'s n-grams and matches when overlap ≥ threshold. Threshold is a `DOUBLE` in `(0.0, 1.0]`. The column needs Pos+Freq features (n-gram analyzer typically registers them):

```sql
-- 70% n-gram overlap with the query string (typo-tolerant)
WHERE b @@ ts_ngram('quick fox jumps', 0.7)

-- 100% overlap (every query n-gram must be present in the indexed value)
WHERE b @@ ts_ngram('quick brown', 1.0)

-- combine with boolean operators
WHERE b @@ (ts_ngram('quick fox', 0.6) || ts_phrase('lazy dog'))

-- with boost
WHERE b @@ ts_ngram('quick brown fox', 0.8) ^ 1.5
ORDER BY bm25(tableoid)
```

#### Complex boolean trees (`||`, `&&`, `!!`)

Operators compose to arbitrary depth. `!!` is prefix NOT and applies to one operand; `||` and `&&` are left-associative at the same precedence tier. Wrap in parens to disambiguate or to override DuckDB's overload-resolution preference for VARCHAR concat on bare-string operands:

```sql
-- (A AND B) OR (C AND NOT D)
WHERE b @@ ((ts_phrase('quick brown') && ts_like('fox%'))
         || (ts_levenshtein('lazy', 1) && !!ts_phrase('spam')))

-- exclude rows matching either of two phrases
WHERE b @@ !!(ts_phrase('out of stock') || ts_phrase('discontinued'))

-- "must contain X, must not contain Y, may contain Z (boost)"
WHERE b @@ (ts_phrase('quick fox')
         && !!ts_any('spam', 'phishing', 'malware')
         && (ts_starts_with('premium') ^ 2.0 || ts_any('vip','enterprise')))
ORDER BY bm25(tableoid) DESC
LIMIT 20;

-- per-leg analyzer override inside a complex tree
WHERE b @@ (
       'Quick'::tokenize('lower_text')              -- lowercase
    || 'EXACT-SKU-42'::tokenize('identity')          -- raw bytes
    || (ts_like('foo%') && !!ts_levenshtein('bar', 1))     -- mixed
)

-- ##-phrase nested inside a boolean tree
WHERE b @@ (
       (ts_phrase('breaking news') ## 5 ## ts_phrase('press conference')) ^ 3.0
    || ts_phrase('exclusive interview')
    && !!ts_regexp('.*sponsored.*')
)

-- ANY_OF/ALL_OF inside boolean composition
WHERE b @@ (
      ts_all(['quick', 'brown', 'fox'])              -- every token present
   && ts_any(['lazy', 'sleepy', 'tired'], 1)         -- at least one
   && !!ts_any(['spam', 'phishing'])                 -- none of these
)
```


server/connector/search_filter_builder.cpp
```diff
@@ -803,98 +1095,57 @@ Result FromPhrase(irs::BooleanFilter& filter, const FilterContext& ctx,
...
-            ERR_MSG("PHRASE has consecutive gaps at argument ", i));
-        }
-        gap_min = gap_max = gap.value();
-        has_gap = true;
-      } break;
+        pending_gap.reset();
+      }
+      continue;
+    }
+    if (!IsPhraseGapValue(*const_val)) {
```
Dronplane: This check duplicates same checks that ParsePhraseGap does. Could we just parse gap here?If fail - report it (adjust ParsePhraseGap error message in case of invalid type found so it names expected types)
Yes having valid gap type + invalid argument sequence could save us time actually parsing the gap value, but that discourages good path (type check twice). And for fail case we still report either invalid sequence or invalid value. So anyway user sees just first error. That makes just type check first not very valuable.

server/connector/search_filter_builder.cpp
```diff
@@ -907,371 +1158,2131 @@ Result FromPhrase(irs::BooleanFilter& filter, const FilterContext& ctx,
...
+                                   "## gap must be a constant"};
+  }
+  return ParsePhraseGap(*val, "##");
+}
+
+// True iff `expr`'s return_type indicates a gap operand (bare INTEGER
+// or INTEGER[]) rather than a TSQUERY part. Used to distinguish the
+// RHS of a `##` binary: if INTEGER/INTEGER[], treat as gap; else as
+// next phrase part.
+bool IsPhraseSeqGapType(const duckdb::Expression& expr) {
```
Dronplane: A bit strange that ARRAY is missing here. IsPhraseGapValue has it. Is there any specific reason why array is not possible to write in tsquery? If no we can re-use here IsPhraseGapValue. More to say parsing gap internally calls ParsePhraseGap that handles ARRAY

server/pg/commands/create_tsdictionary.cpp
```diff
@@ -466,6 +471,43 @@ class CreateTSDictionaryOptions : public OptionsParser {
     _builder.add(GetVPackName(tokenizer_options::kNumHashes.name), hashes);
   }
 
+  void ParseWildcard(std::string_view prefix) {
+    // The wildcard template's nested base analyzer is configured with
+    // a `tokenizer_` user-facing prefix (e.g. `tokenizer_template`,
+    // `tokenizer_delimiter`) so the parameter spelling matches the
+    // catalog noun -- this differs from MinHash's `analyzer_` legacy
```
Dronplane: Do we have a reason to keep legacy 'analyzer' noun in MinHash?  Maybe change it too while we have a luxury to not care about backward compatibility?

server/connector/functions/search.cpp
```diff
@@ -21,82 +21,119 @@
...
+  auto snapshot = conn_ctx.EnsureCatalogSnapshot();
+  if (!snapshot) {
+    return {};
+  }
+  auto entry =
+    snapshot->GetTokenizer(db_id, qualified.schema, qualified.relation);
+  if (!entry) {
+    return {};
+  }
+  auto wrapper_or = entry->GetTokenizer();
```
Dronplane: NIT but why "or" in name?

server/pg/commands/create_tsdictionary.cpp
gnusi: I'll drop it

server/connector/search_filter_builder.cpp
gnusi: Fixed

server/catalog/inverted_index.h
```diff
@@ -46,7 +46,7 @@ struct InvertedIndexColumnInfo {
 };
 
 struct ColumnAnalyzer {
```
MBkkt:
```suggestion
struct ColumnTokenizer {
```

server/connector/ts_anyall.cpp
```diff
@@ -0,0 +1,369 @@
...
+    opts.terms.emplace(std::move(t));
+  }
+}
+
+}  // namespace
+
+// Parses an `any_of` / `all_of` call into (args, synthesised, min_match).
+// `synthesised` owns BoundConstantExpression wrappers when the list was
+// constant-folded; the caller must keep it alive for the duration of `args`
+// use. `min_match` is empty for `all_of` and for `any_of` when not provided.
```
MBkkt: AnyOf/AllOf is bad names

server/connector/tsq_anyallof.cpp
```diff
@@ -0,0 +1,369 @@
...
+// constant-folded; the caller must keep it alive for the duration of `args`
+// use. `min_match` is empty for `all_of` and for `any_of` when not provided.
+void ExtractAnyAllOfArgs(
+  const duckdb::BoundFunctionExpression& func, bool is_any,
+  std::vector<const duckdb::Expression*>& args,
+  std::vector<duckdb::unique_ptr<duckdb::Expression>>& synthesised,
+  std::optional<size_t>& min_match) {
+  constexpr auto kSyntaxHint =
+    "Example: any_of([TERM('a'), TERM('b')]) (OR), all_of([TERM('a'), "
+    "TERM('b')]) (AND), any_of([TERM('a'), TERM('b'), TERM('c')], 2).";
```
MBkkt: 1. should be static constexpr std::string_view
2. is someone actually saw this? looks quite complicated for hint

server/connector/tsq_common.hpp
```diff
@@ -0,0 +1,222 @@
...
+
+  FilterContext WithBoost(irs::score_t factor) const {
+    return {.negated = negated,
+            .boost = boost * factor,
+            .column_getter = column_getter,
+            .column_cache = column_cache,
+            .identity = identity,
+            .tokenizer = tokenizer,
+            .client_context = client_context,
+            .scored_terms_limit = scored_terms_limit};
```
MBkkt:
```suggestion
            .scored_terms_limit = scored_terms_limit,};
```
codestyle, in a differetn places

server/connector/ts_common.hpp
```diff
@@ -0,0 +1,222 @@
...
+  Term,
+  Like,
+  Prefix,
+  Ngram,
+  Fuzzy,
+  AnyOf,
+  AllOf,
+  Range,
+  Regexp,
+  Less,
```
MBkkt, line 219: Lt, Le, Ge, Gt, Eq, Ne -- please

server/connector/tsq_common.hpp
```diff
@@ -0,0 +1,222 @@
...
+  Unknown,
+  Phrase,
+  Term,
+  Like,
+  Prefix,
+  Ngram,
+  Fuzzy,
+  AnyOf,
+  AllOf,
+  Range,
```
MBkkt: Between?

server/connector/ts_common.hpp
```diff
@@ -0,0 +1,222 @@
...
+  LessEq,
+  Greater,
+  GreaterEq,
+  Tokenize,
```
MBkkt, line 223: is it lexize?

server/connector/ts_phrase.cpp
```diff
@@ -0,0 +1,619 @@
...
+
+// True iff `id` is one of the integral types we accept as a phrase-gap
+// scalar (any signed integer narrower than or equal to BIGINT). Values
+// in this set need a sign check at parse time.
+bool IsPhraseGapIntegralTypeId(duckdb::LogicalTypeId id) {
+  return id == duckdb::LogicalTypeId::TINYINT ||
+         id == duckdb::LogicalTypeId::SMALLINT ||
+         id == duckdb::LogicalTypeId::INTEGER ||
+         id == duckdb::LogicalTypeId::BIGINT;
+}
```
MBkkt: I think we can support more cases, any numeric is ok, we just need to get constant number and check is it in right range or not

tests/server/connector/duckdb_search_filter_builder_test.cpp
```diff
@@ -649,21 +681,13 @@ TEST_F(SearchFilterBuilderTest, test_TypesResolving) {
...
                  SegmentationAnalyzerProvider);
   }
-  {
-    // TERM_EQ with segmentation analyzer is accepted.
-    std::vector<ColumnSpec> columns{
-      {.id = 1, .type = duckdb::LogicalType::VARCHAR, .name = "b"}};
-    irs::And expected;
-    AddTermFilter<std::string_view>(expected, 1, std::string_view{"foo"});
-    AssertFilter(expected, "SELECT * FROM foo WHERE TERM_EQ(b, 'foo')", columns,
-                 true, SegmentationAnalyzerProvider);
```
MBkkt, line 665: keep test

tests/server/connector/duckdb_search_filter_builder_test.cpp
```diff
@@ -1427,16 +1451,6 @@ TEST_F(SearchFilterBuilderTest, test_LikeCustomEscape) {
     columns, true);
 }
 
-TEST_F(SearchFilterBuilderTest, test_TermLikeEscape) {
```
MBkkt, line 1430: keep test

tests/server/connector/duckdb_search_filter_builder_test.cpp
```diff
@@ -1617,115 +1633,199 @@ TEST_F(SearchFilterBuilderTest, test_PhraseGapBetweenMultiTokenPatterns) {
...
+  AssertFilter(expected, "SELECT * FROM foo WHERE b @@ LESS(100)", columns,
+               true);
+}
+
+TEST_F(SearchFilterBuilderTest, test_TermGreaterEq_IntegerColumn) {
+  std::vector<ColumnSpec> columns{
+    {.id = 1, .type = duckdb::LogicalType::INTEGER, .name = "b"}};
+  irs::And expected;
+  AddRangeFilter<int32_t>(expected, 1, 50, true, std::nullopt, false);
+  AssertFilter(expected, "SELECT * FROM foo WHERE b @@ GREATER_EQUAL(50)",
```
MBkkt: does b >= 50 works? Also quite difficult to understand such expressions, because b outside of comparsion. Can't we do something better?

tests/server/connector/duckdb_search_filter_builder_test.cpp
```diff
@@ -1794,175 +1963,158 @@ TEST_F(SearchFilterBuilderTest, test_TermLike_WithNot) {
   irs::And expected;
   auto& not_filter = expected.add<irs::Not>();
   AddLikeFilter(not_filter, 1, "%foo_");
-  AssertFilter(expected, "SELECT * FROM foo WHERE NOT(TERM_LIKE(a, '%foo_'))",
+  AssertFilter(expected, "SELECT * FROM foo WHERE NOT(a @@ (LIKE('%foo_')))",
                columns, true, SegmentationAnalyzerProvider);
 }
 
-TEST_F(SearchFilterBuilderTest, test_TermGe_AndTermLe_Range) {
```
MBkkt, line 1801: test removed

MBkkt (approved):

PR #588, opened 2026-04-30
chore: Fix connection sequence and server replies for client tools compatibility

Dronplane:
1. Added "NoOverwrite" mode for writable settings that we do not allow to change yet. So client tools can issue SET requests that sets same value we already have.
2. For client_encoding added canonicalized comparison so client tools can mix casing/dashes in encoding names. (UTF8, utf-8, utf8 are treated as equal)
3. Moved sending backend key packet after parameter status report to match what postgresql does (not strictly defined in protocol but just to stay closer to native implementation)
4. Changed true/false to on/off for boolean settings as some clients (JDBC for example) are upset if it is not match postgresql on/off
5. Added applying with same validation as for SET  settings received as part of start-up packet.
6. Fixed parameter inferrinng during Parse - patch for DuckDB so if caller supplies OIDs we need not to guess from context.
7. Add OID, CID, TID, XID, XID8 types to pg_catalog 

Linked DuckDB patch https://github.com/serenedb/duckdb/pull/7


server/pg/pg_comm_task.cpp
```diff
@@ -370,6 +363,20 @@ void PgSQLCommTaskBase::HandleClientHello(std::string_view packet) {
...
+            _connection_ctx->SetSettingChecked(user_setting.first,
+                                               user_setting.second, false);
+          }
+        } catch (const duckdb::Exception& e) {
+          // As per protocol description  - encountered during client hello
+          // error aborts connection.
+          return SendError(duckdb::ErrorData(e).RawMessage(),
+                           ERRCODE_INVALID_PARAMETER_VALUE);
+        }
+      }
```
gnusi: safeCall?

server/query/config.cpp
```diff
@@ -103,10 +103,32 @@ bool Config::GetReadYourOwnWrites() const {
...
+}
+
 std::optional<std::string> Config::Get(std::string_view key) const {
   duckdb::Value value;
   if (_client_ctx.TryGetCurrentSetting(std::string{key}, value)) {
-    return value.ToString();
+    if (value.type().id() == duckdb::LogicalTypeId::BOOLEAN) {
+      // PostgreSQL protocol uses on/off not true/false. And some client tools
+      // like JDBC expects exactly on/off
+      return duckdb::BooleanValue::Get(value) ? "on" : "off";
```
gnusi: check out FormatDisplayValue, maybe it can work

server/pg/pg_comm_task.cpp
```diff
@@ -370,6 +363,20 @@ void PgSQLCommTaskBase::HandleClientHello(std::string_view packet) {
       _connection_ctx->SetSetting(
         "is_superuser", _connection_ctx->isSuperuser() ? "on" : "off", false);
 
+      // Apply all user settings from startup packet
+      for (const auto& user_setting : _client_parameters) {
+        try {
+          if (_connection_ctx->HasSetting(user_setting.first)) {
+            _connection_ctx->SetSettingChecked(user_setting.first,
+                                               user_setting.second, false);
+          }
```
gnusi: Maybe compound call?

HasSetting followed by SetSettingsChecked does lookup twice.

server/pg/pg_catalog/pg_type.cpp
```diff
@@ -168,6 +168,40 @@ constexpr auto kSampleData = std::to_array<PgType>({
...
+    .typalign = PgType::Typalign::Int,
+    .typstorage = PgType::Typstorage::Plain,
+    .typnotnull = false,
+    .typbasetype = 0,
+    .typtypmod = -1,
+    .typndims = 0,
+    .typcollation = 0,
+    .typdefaultbin = {},
+    .typdefault = {},
+    .typacl = {},
```
MBkkt, line 339: we need to add other oid-like types here, that we support

server/pg/pg_comm_task.cpp
MBkkt: Also maybe return some error, instead of throw + catch

server/pg/pg_types.cpp
```diff
@@ -606,11 +606,9 @@ std::expected<duckdb::Value, DeserializeError> DeserializeBinaryParameter(
       }
     } break;
     case BIGINT: {
-      if (IsOidLike(type)) {
-        if (data.size() == 4) {
-          return duckdb::Value::BIGINT(
-            static_cast<int64_t>(absl::big_endian::Load<int32_t>(data.data())));
```
MBkkt, line 612: revert?

server/query/config_variables.cpp
```diff
@@ -47,6 +50,56 @@ void Readonly(duckdb::ClientContext&, duckdb::SetScope, duckdb::Value&) {
...
+  }
+  throw duckdb::InvalidInputException{
+    "parameter \"%s\" cannot be changed from \"%s\" to \"%s\"", kName.data(),
+    current.ToString(), value.ToString()};
+}
+
+// PostgreSQL drivers supply client_encoding in many forms (UTF8, UTF-8, utf_8,
+// utf8, ...). To tackle this we have a special overload for encoding.
+void NoOverwriteClientEncoding(duckdb::ClientContext& ctx, duckdb::SetScope,
+                               duckdb::Value& value) {
```
MBkkt, line 82: This looks messy, why can't we just accept whatever driver supply?

MBkkt (changes requested):
I think we shouldn't have such hack for oid-like types

server/query/config_variables.cpp
Dronplane: We can't "just accept". As if driver ask something non utf8 we are obliged to do encoding/decoding of all strings. That we don't plan for now. And as most of the clients uses utf8 we just make sure we are still utf8 in any form. In future we can add encoding support. It is just out of scope for this PR.

server/pg/pg_comm_task.cpp
Dronplane: Actually that was neuro-slope.  Checked with real PG - it does not ignores unknown args but rather raises error. So pre-check existance removed

server/pg/pg_comm_task.cpp
Dronplane: Done

server/pg/pg_catalog/pg_type.cpp
Dronplane: Added tid, cid, xid, xid8. Looks like that's all we currently have

server/pg/pg_comm_task.cpp
```diff
@@ -507,12 +528,10 @@ void PgSQLCommTaskBase::DescribeAnalyzedQuery(
...
-  // that matches PG, which treats a client-supplied OID as an assertion
-  // (e.g. `SELECT $1::int4` with client OID=text reports text here, since
-  // the cast coerces a client-sent text into int4 at Execute). Otherwise
-  // fall back to DuckDB's planner-resolved type, and finally to text.
+  // Emit (count, [oid]*num). Client OIDs from Parse are passed to DuckDB's
+  // binder as type hints, so expected_types reflects them for non-zero slots
+  // (and the binder's inference for OID-0 slots). Fall back to text when the
+  // binder couldn't infer.
   const auto& prepared = *stmt.prepared;
   const auto expected_types = prepared.GetExpectedParameterTypes();
```
gnusi: unrelated, to the PR but is it ok, that we copy map here?

server/pg/pg_comm_task.cpp
```diff
@@ -524,17 +543,12 @@ void DescribeParameters(const DuckDBStatement& stmt, message::Buffer& buffer) {
...
-               type_it->second.id() != duckdb::LogicalTypeId::UNKNOWN &&
-               type_it->second.id() != duckdb::LogicalTypeId::INVALID) {
-      oid = Type2Oid(type_it->second);
-    } else {
-      oid = PgTypeOID::kText;
-    }
+    auto type_it = expected_types.find(absl::StrCat(i));
+    int32_t oid = (type_it != expected_types.end() &&
+                   type_it->second.id() != duckdb::LogicalTypeId::UNKNOWN &&
+                   type_it->second.id() != duckdb::LogicalTypeId::INVALID)
```
gnusi: Maybe dedup this? smth like
```
int32_t GetOid(const auto& expected_types, uint16_t id, auto fallback) {
    auto type_it = expected_types.find(absl::StrCat(id));
    return (type_it != expected_types.end() &&
                 type_it->second.id() != duckdb::LogicalTypeId::UNKNOWN &&
                 type_it->second.id() != duckdb::LogicalTypeId::INVALID)
                  ? type_it->second
                  : fallback;
}
```

server/pg/pg_comm_task.cpp
Dronplane: No reason. Will fix.

server/pg/pg_comm_task.cpp
Dronplane: Done

gnusi (approved):

server/pg/pg_catalog/pg_type.cpp
MBkkt: Also reg* classes and name class

Dronplane: Done

MBkkt (approved):
@Dronplane 

But rebase plz, I updated duckdb branch

PR #592, opened 2026-05-01
feat: add WAL recovery

pashandor789:


tests/sqllogic/recovery/wal_index_recovery_huge.test_slow
MBkkt, line 1: test_slow instead of disabled!

server/search/wal_recovery.h
```diff
@@ -0,0 +1,36 @@
...
+#pragma once
+
+namespace sdb::search {
+
+// Stateless one-shot. On Run(), walks the catalog snapshot, picks up every
+// inverted-index shard whose persisted iresearch tick lags the engine's
+// recovered tick, then reads the RocksDB WAL once and fans entries out to
+// each lagging shard's accumulator (routing key = table ObjectId).
+class WalRecovery {
+ public:
```
MBkkt: Why this class? maybe just function?

server/search/inverted_index_shard.h
```diff
@@ -234,6 +234,13 @@ class InvertedIndexShard final
   Result CleanupUnsafeImpl();
   void InitPostRecovery(bool is_new);
 
+ public:
```
MBkkt: public should be once

server/search/inverted_index_shard.cpp
```diff
@@ -303,11 +306,13 @@ void InvertedIndexShard::InitPostRecovery(bool is_new) {
...
         }
 
-        // Register flush subscription if we are loading existing index
-        // If not finishCreation would be called later when indexing finishes
+        // Lagging shards defer FinishCreation+StartTasks to WalRecovery::Run.
         if (!is_new) {
-          self->FinishCreation();
-          self->StartTasks();
+          const bool needs_recovery = self->_recovery_tick < recovery_tick;
+          if (!needs_recovery) {
```
MBkkt: we want commit/consolidate tasks during recovery

server/catalog/index.cpp
```diff
@@ -174,12 +175,34 @@ Result ApplyHNSWOptions(
...
+  // time. TIMESTAMP is supported by the writer but not by the search filter
+  // path yet, so it stays rejected explicitly.
   for (auto c : indexed_columns) {
     SDB_ASSERT(c.catalog_column);
-    if (c.catalog_column->type.id() == duckdb::LogicalTypeId::TIMESTAMP ||
-        c.catalog_column->type.id() == duckdb::LogicalTypeId::HUGEINT) {
-      return {ERROR_BAD_PARAMETER, "Column ", c.name,
-              " has unsupported kind and can not be indexed"};
+    auto kind = c.catalog_column->type.id();
+    bool supported = kind == duckdb::LogicalTypeId::SQLNULL ||
```
MBkkt, line 186: add test plz

server/search/wal_recovery.cpp
```diff
@@ -0,0 +1,527 @@
...
+
+  Row& GetRow(std::string_view pk) {
+    auto [it, inserted] = pk2row.try_emplace(pk);
+    if (inserted) {
+      it->second.indexed_cols.resize(indexed_columns.size());
+    }
+    return it->second;
+  }
+
+  static auto GetComparator() {
```
MBkkt: better name

server/search/wal_recovery.cpp
```diff
@@ -0,0 +1,527 @@
...
+  const uint32_t default_cf_id = default_cf->GetID();
+
+  SDB_INFO("xxxxx", Logger::SEARCH,
+           "Starting WAL recovery, to skip it use the flag: "
+           "\"--search.skip-wal-recovery\"");
+
+  std::unique_ptr<rocksdb::TransactionLogIterator> iter;
+
+  auto s = db->GetUpdatesSince(
+    min_start_tick, &iter, rocksdb::TransactionLogIterator::ReadOptions(true));
```
MBkkt: Create separate read options here, with explicit settings, like async reads, etc. For an example we know that no-one write to rocksdb during this recovery.

Measure different options plz

server/search/wal_recovery.cpp
```diff
@@ -0,0 +1,527 @@
...
+  SDB_FATAL_IF("xxxxx", Logger::SEARCH, !s.ok(),
+               "WAL recovery: failed to open WAL iterator from tick ",
+               min_start_tick, ": ", s.ToString());
```
MBkkt: Maybe add option to ignore failures during recovery?

server/search/wal_recovery.cpp
```diff
@@ -0,0 +1,527 @@
...
+  SDB_ASSERT(iter);
+
+  std::vector<rocksdb::BatchResult> live_batches;
+  auto flush_all = [&]() {
```
MBkkt:
```suggestion
  auto flush_all = [&] {
```

server/search/wal_recovery.cpp
```diff
@@ -0,0 +1,527 @@
...
+    live_batches.clear();
+  };
+
+  while (iter->Valid()) {
+    auto status = iter->status();
+    SDB_FATAL_IF("xxxxx", Logger::SEARCH, !status.ok(),
+                 "WAL recovery: iterator error: ", status.ToString());
+
+    auto batch = iter->GetBatch();
+    if (batch.sequence > end_tick) {
```
MBkkt: is this possible? Looks like SDB_FATAL to me

server/search/wal_recovery.cpp
```diff
@@ -0,0 +1,527 @@
...
+      break;
+    }
+
+    FanoutBatchHandler handler{table2shards, default_cf_id, batch.sequence};
```
MBkkt: better name instead of FanoutBatchHandler

server/search/wal_recovery.cpp
```diff
@@ -0,0 +1,527 @@
...
+    status = batch.writeBatchPtr->Iterate(&handler);
+    SDB_FATAL_IF("xxxxx", Logger::SEARCH, !status.ok(),
+                 "WAL recovery: batch iterate failed at seq ", batch.sequence,
+                 ": ", status.ToString());
+
+    live_batches.emplace_back(std::move(batch));
+
+    const auto need_flush = absl::c_any_of(shards, [&](const ShardState& s) {
+      return s.pk2row.size() >= kFlushThreshold;
+    });
```
MBkkt: cache?

server/search/wal_recovery.cpp
```diff
@@ -0,0 +1,527 @@
...
+      "xxxxx", Logger::SEARCH, "WAL recovery: index '", s.shard->GetId().id(),
+      "' replayed (", s.start_tick, ", ", s.end_tick,
+      "], inserted=", s.total_inserted, ", deleted=", s.total_deleted,
+      ", commit_code=", static_cast<int>(code), ", commit_time_ms=", r.time_ms);
+  }
+
+  for (auto& s : shards) {
+    s.shard->FinishCreation();
+    s.shard->StartTasks();
+  }
```
MBkkt: normal Commit (still with wait=true) and make FinishCreation/StartTasks before this

MBkkt (commented):
Also add tests with mix DDL (index/table/schema/database drop inside recovery range)
and same about inverted index create during recovery.

server/search/wal_recovery.cpp
```diff
@@ -0,0 +1,527 @@
...
+
+  uint64_t recovery_time_ms =
+    std::chrono::duration_cast<std::chrono::milliseconds>(
+      std::chrono::steady_clock::now() - recovery_begin)
+      .count();
+  SDB_INFO("xxxxx", Logger::SEARCH, "WAL recovery: completed in ",
+           recovery_time_ms, " ms, indexes=", shards.size());
```
MBkkt: more human readable time

server/search/wal_recovery.cpp
```diff
@@ -0,0 +1,527 @@
...
+        continue;
+      }
+      auto& row = s->GetRow(pk);
+      row.full_key = full_key;
+      fn(row, col_it->second);
+    }
+  }
+
+  TableToShards& _table2shards;
+  uint32_t _default_cf_id;
```
MBkkt, line 394: bool needs flush?

server/search/wal_recovery.cpp
```diff
@@ -0,0 +1,527 @@
...
+class FanoutBatchHandler final : public rocksdb::WriteBatch::Handler {
+ public:
+  FanoutBatchHandler(TableToShards& table2shards, uint32_t default_cf_id,
+                     rocksdb::SequenceNumber batch_sequence)
+    : _table2shards{table2shards},
+      _default_cf_id{default_cf_id},
+      _batch_sequence{batch_sequence} {}
+
+  rocksdb::Status PutCF(uint32_t cf_id, const rocksdb::Slice& key,
+                        const rocksdb::Slice& value) override {
```
MBkkt:
```suggestion
                        const rocksdb::Slice& value) final {
```

server/search/wal_recovery.cpp
```diff
@@ -0,0 +1,527 @@
...
+    return rocksdb::Status::OK();
+  }
+
+  rocksdb::Status SingleDeleteCF(uint32_t cf_id,
+                                 const rocksdb::Slice& key) override {
+    return DeleteCF(cf_id, key);
+  }
+
+  rocksdb::Status DeleteRangeCF(uint32_t, const rocksdb::Slice&,
+                                const rocksdb::Slice&) override {
```
MBkkt: delete range happens when table removed, but it's probably should mean index deleted.

But this call exists for truncate (not implemented yet ~~by you~~), irs::IndexWriter has Clear, check old code about truncate

server/search/wal_recovery.cpp
```diff
@@ -0,0 +1,527 @@
...
+  }
+  rocksdb::Status MarkEndPrepare(const rocksdb::Slice&) override {
+    return rocksdb::Status::OK();
+  }
+  rocksdb::Status MarkCommit(const rocksdb::Slice&) override {
+    return rocksdb::Status::OK();
+  }
+  rocksdb::Status MarkRollback(const rocksdb::Slice&) override {
+    return rocksdb::Status::OK();
+  }
```
MBkkt: SDB_ASSERT(false) if method not called and return some invalid status, to avoid cases when it's called and works incorrectly

server/search/wal_recovery.cpp
```diff
@@ -0,0 +1,527 @@
...
+  FanoutBatchHandler(TableToShards& table2shards, uint32_t default_cf_id,
+                     rocksdb::SequenceNumber batch_sequence)
+    : _table2shards{table2shards},
+      _default_cf_id{default_cf_id},
+      _batch_sequence{batch_sequence} {}
+
+  rocksdb::Status PutCF(uint32_t cf_id, const rocksdb::Slice& key,
+                        const rocksdb::Slice& value) override {
+    ForEachMatchingShard(cf_id, key, [&](Row& row, size_t col_idx) {
+      row.indexed_cols[col_idx] = {value.data(), value.size()};
```
MBkkt: create vector on first access?

server/search/wal_recovery.cpp
```diff
@@ -0,0 +1,527 @@
...
+    }
+    if (started == 0) {
+      return;
+    }
+
+    const auto col_id = absl::big_endian::Load<catalog::Column::Id>(
+      key.data() + sizeof(ObjectId));
+    std::string_view pk{key.data() + kPrefix, key.size() - kPrefix};
+    std::string_view full_key{key.data(), key.size()};
+    for (auto* s : std::span{shards.data(), started}) {
```
MBkkt: combine these two cycles into one?

server/search/wal_recovery.cpp
```diff
@@ -0,0 +1,527 @@
...
+      return;
+    }
+    ObjectId id{absl::big_endian::Load64(key.data())};
+    auto table_it = _table2shards.find(id);
+    if (table_it == _table2shards.end()) {
+      return;
+    }
+
+    auto& [started, shards] = table_it->second;
+    SDB_ASSERT(std::ranges::is_sorted(shards, ShardState::GetComparator()));
```
MBkkt: check it once, not for each Batch

server/search/wal_recovery.cpp
```diff
@@ -0,0 +1,527 @@
...
+ private:
+  template<typename Fn>
+  void ForEachMatchingShard(uint32_t cf_id, const rocksdb::Slice& key,
+                            Fn&& fn) {
+    if (cf_id != _default_cf_id) {
+      return;
+    }
+    static constexpr auto kPrefix =
+      sizeof(ObjectId) + sizeof(catalog::Column::Id);
+    if (key.size() <= kPrefix) {
```
MBkkt: comment about this needed for our ugly hack with empty key written

server/search/wal_recovery.cpp
```diff
@@ -0,0 +1,527 @@
...
+    ForEachMatchingShard(cf_id, key, [&](Row& row, size_t col_idx) {
+      row.indexed_cols[col_idx] = {value.data(), value.size()};
+      row.op = RowOp::Put;
+    });
+    return rocksdb::Status::OK();
+  }
+
+  rocksdb::Status DeleteCF(uint32_t cf_id, const rocksdb::Slice& key) override {
+    ForEachMatchingShard(
+      cf_id, key, [&](Row& row, size_t col_idx) { row.op = RowOp::Delete; });
```
MBkkt: We should clear row.indexed_cols when delete happened. Looks like bug, create test for this first (ensure it's failed and then fix code and rerun test).

test probably looks like this
```
INSERT ...
DELETE row
UPDATE
```

Actually seems like your code works, because such test is impossible (after delete always should be insert). But lets left comment about this and still add test that verify this.

server/search/wal_recovery.cpp
```diff
@@ -0,0 +1,527 @@
...
+          // to be able to write it to iresearch.
+          std::string_view pk = row->full_key.substr(kPrefix);
+          get_key_buffer.resize(kPrefix);
+          get_key_buffer.append(pk.data(), pk.size());
+          value_buffer.Reset();
+          auto status = db.Get(read_opts, &cf, get_key_buffer, &value_buffer);
+          SDB_FATAL_IF("xxxxx", Logger::SEARCH, !status.ok(),
+                       "WAL recovery: rocksdb Get failed for index '",
+                       s.shard->GetId().id(), "' col=", col.id, ": ",
+                       status.ToString());
```
MBkkt: Option to skip this?

server/search/wal_recovery.cpp
```diff
@@ -0,0 +1,527 @@
...
+  s.table_object_id = s.table->GetId();
+
+  s.full_key_prefix.resize(sizeof(ObjectId) + sizeof(catalog::Column::Id));
+  absl::big_endian::Store64(s.full_key_prefix.data(), s.table_object_id.id());
+  return true;
+}
+
+void EnsureTrxOpen(ShardState& s,
+                   const std::shared_ptr<const catalog::Snapshot>& snapshot) {
+  if (s.trx.has_value()) {
```
MBkkt: codestyle!
```suggestion
  if (s.trx) {
```

server/search/wal_recovery.cpp
```diff
@@ -0,0 +1,527 @@
...
+bool ResolveShardMetadata(ShardState& s, const catalog::Snapshot& snapshot) {
+  auto index_obj =
+    snapshot.template GetObject<catalog::Index>(s.shard->GetIndexId());
+  if (!index_obj) {
+    return false;
+  }
+  auto inverted =
+    std::dynamic_pointer_cast<const catalog::InvertedIndex>(index_obj);
+  if (!inverted) {
+    return false;
```
MBkkt, line 126: We can do this without dynamic cast? With enum check? ObjectType::InvertedIndex?

server/search/wal_recovery.cpp
```diff
@@ -0,0 +1,527 @@
...
+#include "rest_server/serened_single.h"
+#include "rocksdb_engine_catalog/rocksdb_column_family_manager.h"
+#include "rocksdb_engine_catalog/rocksdb_engine_catalog.h"
+#include "search/inverted_index_shard.h"
+#include "storage_engine/engine_feature.h"
+
+namespace sdb::search {
+namespace {
+
+constexpr size_t kFlushThreshold = 1 << 16;
```
MBkkt: lets make smallest number that isn't slow. Also good to account case (for perf measure) with large columns count and with large strings in columns

MBkkt (approved):
Looks good, few things that needs to be fixed written above

PR #594, opened 2026-05-01
feat: View indexing

MBkkt:
- https://github.com/serenedb/duckdb-iceberg/pull/1
- https://github.com/serenedb/duckdb/pull/8

## Summary

Adds the ability to `CREATE INDEX ... USING inverted/secondary` on a `VIEW` whose body is a simple `SELECT [cols] FROM <reader>(...)` over one of the recognised fast-path sources, and rewires the query side to materialize via a new `IndexSource` abstraction.

The PR has two halves:

1. **DuckDB submodule** — extends the `CREATE INDEX` path so views can opt in (today only SerendDB's catalog does), and adds a new lookup-mode TF API (`pk_lookups` + `pk_output_positions`) that lets a caller pull specific rows out of read_parquet/read_csv/read_{json,ndjson,json_objects,ndjson_objects} at caller-positional output slots, with no data round-trip through scratch chunks.

2. **SerendDB main repo** — drops the previous `external_scan` machinery, introduces a polymorphic `IndexSource` (`PrimaryKeyBatch` variant for the three pk shapes), centralizes file-/glob-/iceberg-/rocksdb-view materialization, threads pk-spec metadata through CREATE INDEX, and skips WAL recovery for view-backed indexes (they're static).

Tests cover read_parquet single + glob (including S3), read_csv/json/ndjson + their `_auto` aliases and `_objects` variants, iceberg (REST + MinIO fixture), the views-over-rocksdb-table fast path, and a secondary-index-over-view smoke. The slow tests stay in `*.test_slow`.

## DuckDB-side API additions

`TableFunctionInput` now carries two parallel spans for lookup TFs:

```cpp
//! Per-call sorted file-row-numbers / byte-offsets to look up.
std::span<const int64_t> pk_lookups;
//! Output slot for `pk_lookups[i]`. The TF writes directly there; in glob
//! mode multiple per-file calls share `output` and write at disjoint slots.
std::span<const idx_t> pk_output_positions;
```

Three lookup TF entry points, one per format:

- `MakeParquetLookupTableFunction()` — file metadata + reader cached in gstate; per call narrows row groups via `upper_bound` over `row_group_starts`, then per-row `VectorOperations::Copy` from native `scan_chunk` to the caller's output at `pk_output_positions[k]`. No NATIVE→DECLARED cast inside the TF — the caller composes that with its own NATIVE→PROJECTED cast in one pass.
- `MakeCSVLookupTableFunction()` — `parse_chunk` is rebound to share buffers with `output` via `RebindParseChunkVectors`; setting `result.number_of_rows = pk_output_positions[i]` before each `ParseChunkAppend` makes the parser write each pk's row directly at the caller's slot. Glob mode: per-file calls share the same output, disjoint pk_output_positions accumulate without coordination.
- `MakeJSONLookupTableFunction()` / `MakeJSONObjectsLookupTableFunction()` — `JSONReader::FetchRow` reads ONE record at a given byte offset via a SCAN_ENTIRE_FILE seek-path, using `JSONReaderScanState::units[0]/values[0]` as a one-slot ring; per-pk Transform with count=1 writes via `Vector::Slice` onto `output[row..row+1]`.

The streaming `read_json` / `read_json_objects` paths also pick up a small fix: `file_row_number` is now written at `output.data[output.ColumnCount()-1]`, since the global column index is wrong once constant virtuals are dropped from the local layout.

## SerendDB-side `IndexSource` abstraction

`server/connector/index_source.h` declares:

```cpp
using PrimaryKeyBatch = std::variant<std::monostate, PrimaryKeysBytes,
                                     PrimaryKeyI64, PrimaryKeyI64I64>;

class IndexSource {
 public:
  virtual ~IndexSource() = default;
  virtual PrimaryKeyBatch CreatePkBatch() const = 0;
  virtual void Materialize(duckdb::ClientContext&, PrimaryKeyBatch&,
                           duckdb::idx_t start, duckdb::idx_t count,
                           duckdb::DataChunk& output) = 0;
};
```

Three implementations:

- `RocksDBIndexSource` — reads from the local rocksdb. Picks Seek (large batch), MultiGet (medium), or Get-loop (small) per call. `PrimaryKeysBytes` uses an arena that reserves `kPrefixGap=16` bytes BEFORE each pk; the column-key prefix `[ObjectId][ColumnId]` is stamped in place per column so MultiGet's slice array doesn't need a staging copy of pk bytes.
- `ViewRocksDBIndexSource` — composes a `RocksDBIndexSource` for the base table; matching cols decode straight into `output`, cast cols decode into a per-col scratch `Vector` whose `VectorCache` is reset per call (DeserializeValueIntoDuckDB only ever SetInvalid; without reset stale invalids would persist).
- `ViewFileIndexSourceBase` (single + glob) — sorts pk_bytes within a slice, calls the lookup TF with `pk_output_positions = sort_perm` so the TF lands rows at caller-positional slots; the cast pass at end fans `_tf_target` cast columns into `output[real_proj_slots[c]]`. Glob mode caches per-file `bind_data + gstate` keyed by file index.

`MakeIndexSource(...)` is the factory; chosen at first scan via `bind_data.IsViewBacked()` and `fast_path.pk_spec`. `ResolveViewFastPath` accepts:

| view body                                       | pk_spec                              |
|-------------------------------------------------|--------------------------------------|
| `SELECT * FROM read_parquet('file.parquet')`    | `FileRowNumber`                      |
| `SELECT * FROM read_parquet('glob/*.parquet')`  | `FileIndexPlusRowNumber`             |
| `SELECT * FROM read_csv('file.csv')`            | `FileOffset`                         |
| `SELECT * FROM read_json('file.json')`          | `FileOffset`                         |
| `SELECT * FROM iceberg.<table>`                 | `FileIndexPlusRowNumber`             |
| `SELECT * FROM <serenedb-table>`                | `RocksDBExplicitPK` / `…GeneratedRowId` |

Iceberg snapshot is pinned at `CREATE INDEX` time and stored in the index commit payload; rebind reads the same manifest via the `version` AT-clause so an index built against snapshot N keeps reading snapshot N until rebuilt.

## Caller wiring

`SearchFullScan` / `SearchAnnScan` / `SearchRangeScan` and the SK paths (`SkFullScan` / `SkPointLookup` / `SkRangeScan`) now keep one `PrimaryKeyBatch` per scan in their gstate, switching it to the right alternative on first use via `index_source->CreatePkBatch()`. Per-batch PK collection is `std::visit`-driven so the inner loop is fully type-specialised on the active alternative.

`ANNFilter::is_member` collects a single-row PK and calls `Materialize(batch, 0, 1, scratch)` per HNSW candidate.

The streaming `SearchFullScanFunction` writes scores and `OFFSETS()` LIST(BIGINT) data inline into output during the scan loop; virtual cols (rowid / tableoid) are filled in `WriteVirtualColumns` after.

## WAL recovery for view-backed indexes

`server/search/wal_recovery.cpp` now skips WAL replay for indexes whose relation is a `PgSqlView` — the underlying view is static, there's nothing to replay. The shard still goes through `StartTasks()` + `FinishCreation()` so search threads pick it up.

## Test plan

- [x] `tests/sqllogic/sdb/pg/index/inverted_index_view.test` — single-file parquet/csv/json/ndjson, `*_auto` aliases, `_objects` variants, multi-row-group parquet, top-K BM25 + real-column projection.
- [x] `tests/sqllogic/sdb/pg/index/inverted_index_view_glob.test` — globs across each format, with cast/rename projections.
- [x] `tests/sqllogic/sdb/pg/index/inverted_index_view_variants.test` — variant + nested + json_objects schema.
- [x] `tests/sqllogic/sdb/pg/index/inverted_index_view_s3.test_slow` — S3 fixture for parquet glob.
- [x] `tests/sqllogic/sdb/pg/index/inverted_index_view_iceberg.test_slow` — REST catalog + MinIO; snapshot-pinned reads + schema-evolution scenarios.
- [x] `tests/sqllogic/sdb/pg/index/secondary_index_view.test` — secondary index over view (smoke).
- [x] `make tests/sqllogic` green on CI.
- [x] Build clean with ASAN preset.

---

## Follow-ups

These are known limitations / structural cleanups deferred to follow-up PRs.

### F-2. Naming for scan / lookup / count TFs

The current names mix abstraction layers:

- `iresearch_scan`, `iresearch_count`, `iresearch_ann_fullscan`, `iresearch_ann_range_scan` (TF-level, prefixed by engine)
- `MakeParquetLookupTableFunction` / `MakeCSVLookupTableFunction` / `MakeJSONLookupTableFunction` / `MakeJSONObjectsLookupTableFunction` (in DuckDB submodule, prefixed by file format)
- `RocksDBIndexSource` / `ViewRocksDBIndexSource` / `ViewFileSingleFileIndexSource` / `ViewFileGlobIndexSource` (caller-side)
- `BindFastPathSource` / `MakeFastPathLookupFunction` / `ResolveViewFastPath` ("FastPath" used inconsistently — sometimes meaning "view fast path", sometimes "the lookup path that bypasses MultiFileFunction")

**Action:** standardize on a single vocabulary across the three layers. Suggested split: `<format>_lookup` for the TF, `<source>IndexLookup` for the materializer, `<source>BindLookup` for the bind helper.

### F-4. Performance: per-row `VectorOperations::Copy` in parquet's lookup TF

`ParquetLookupScan` does one `VectorOperations::Copy(src, dst, /*size=*/1, /*src_off=*/k, /*tgt_off=*/pk_output_positions[total+k])` per (row, column). For lookup batches up to STANDARD_VECTOR_SIZE this is fine; for larger batches the per-call fixed cost dominates over the memcpy.

**Action:** if parquet lookup ever shows up in profiles, batch-copy via a target-side selection vector instead of one row at a time. Lower priority — current callers are bounded by STANDARD_VECTOR_SIZE.


server/connector/duckdb_catalog.cpp
```diff
@@ -853,74 +978,129 @@ duckdb::unique_ptr<duckdb::LogicalOperator> SereneDBCatalog::BindCreateIndex(
...
+      expressions.push_back(duckdb::make_uniq<duckdb::BoundColumnRefExpression>(
+        rel_columns[i].second,
+        duckdb::ColumnBinding(get.table_index, duckdb::ProjectionIndex(i))));
+    }
+  } else {
+    create_index_info->names.clear();
+    create_index_info->names.reserve(rel_columns.size());
+    for (const auto& c : rel_columns) {
+      create_index_info->names.push_back(c.first);
+    }
```
gnusi: NIT:
assign_range?

server/connector/duckdb_catalog.cpp
```diff
@@ -853,74 +978,129 @@ duckdb::unique_ptr<duckdb::LogicalOperator> SereneDBCatalog::BindCreateIndex(
...
+      rel_columns.emplace_back(column_info->names[i], column_info->types[i]);
+    }
+  } else {
+    auto& sdb_entry = RequireBaseTable(*resolved_table);
+    auto sdb_table = sdb_entry.GetSereneDBTable();
+    const auto& columns = sdb_table->Columns();
+    rel_columns.reserve(columns.size());
+    for (const auto& c : columns) {
+      rel_columns.emplace_back(c.name, c.type);
+    }
```
gnusi: NIT: 
convert to lambda + return range?

server/connector/duckdb_catalog.cpp
```diff
@@ -853,74 +978,129 @@ duckdb::unique_ptr<duckdb::LogicalOperator> SereneDBCatalog::BindCreateIndex(
...
+    auto column_info = view_entry.GetColumnInfo();
+    if (!column_info) {
+      THROW_SQL_ERROR(ERR_CODE(ERRCODE_FEATURE_NOT_SUPPORTED),
+                      ERR_MSG("view \"", target.name,
+                              "\" must be bound before it can be indexed"));
+    }
+    rel_columns.reserve(column_info->names.size());
+    for (size_t i = 0; i < column_info->names.size(); ++i) {
+      rel_columns.emplace_back(column_info->names[i], column_info->types[i]);
+    }
```
gnusi: NIT: convert to lambda + return range?

server/catalog/local_catalog.cpp
```diff
@@ -1237,6 +1258,44 @@ Result LocalCatalog::CreateIndexImpl(
...
+    const auto& view_info = view.GetInfo();
+    std::vector<Column> columns;
+    columns.reserve(view_info.names.size());
+    for (size_t i = 0; i < view_info.names.size(); ++i) {
+      columns.push_back(Column{
+        .id = static_cast<Column::Id>(i),
+        .type = view_info.types[i],
+        .name = view_info.names[i],
+      });
+    }
```
gnusi: NIT:
use range?

gnusi: NIT:
use range?

server/connector/duckdb_entry_cache.cpp
```diff
@@ -294,10 +294,59 @@ duckdb::unique_ptr<duckdb::CatalogEntry> DuckDBEntryCache::BuildIndexScanEntry(
...
+    info->table = std::string{name};
+    info->schema = std::string{schema.name};
+    for (size_t i = 0; i < vinfo.names.size(); ++i) {
+      info->columns.AddColumn(
+        duckdb::ColumnDefinition(vinfo.names[i], vinfo.types[i]));
+    }
+    std::vector<size_t> indexed_col_indices;
+    for (auto col_id : index.GetColumnIds()) {
+      indexed_col_indices.push_back(static_cast<size_t>(col_id));
+    }
```
gnusi: maybe use uint64_t and just pass it to make_uniq?

server/connector/duckdb_entry_cache.cpp
```diff
@@ -294,10 +294,59 @@ duckdb::unique_ptr<duckdb::CatalogEntry> DuckDBEntryCache::BuildIndexScanEntry(
...
+      ObjectId sk_shard_id;
+      for (auto& shard : snapshot.GetIndexShardsByRelation(view->GetId())) {
+        if (shard->GetIndexId() == index.GetId()) {
+          sk_shard_id = shard->GetId();
+          break;
+        }
+      }
+      if (sk_shard_id == ObjectId{}) {
+        return nullptr;
+      }
```
gnusi: const auto& indexes = snapshot.GetIndexShardsByRelation(view->GetId())
auto it = absl::c_find_if(indexes, [&](auto id){return id == index.GetId(); });
if (it == indexes.end()) {
    return nullptr;
}

server/connector/duckdb_index_scan_entry.cpp
```diff
@@ -23,103 +23,261 @@
...
+}
+
+duckdb::vector<duckdb::column_t> ViewInvertedIndexScanEntry::GetRowIdColumns()
+  const {
+  return {duckdb::COLUMN_IDENTIFIER_ROW_ID};
+}
+
+duckdb::virtual_column_map_t ViewInvertedIndexScanEntry::GetVirtualColumns()
+  const {
+  duckdb::virtual_column_map_t result;
```
gnusi, line 154: reserve?

server/connector/duckdb_index_scan_entry.cpp
```diff
@@ -23,103 +23,261 @@
...
+}
+
+duckdb::vector<duckdb::column_t> ViewSecondaryIndexScanEntry::GetRowIdColumns()
+  const {
+  return {duckdb::COLUMN_IDENTIFIER_ROW_ID};
+}
+
+duckdb::virtual_column_map_t ViewSecondaryIndexScanEntry::GetVirtualColumns()
+  const {
+  duckdb::virtual_column_map_t result;
```
gnusi, line 276: reserve

server/connector/duckdb_physical_create_index.cpp
```diff
@@ -92,27 +90,27 @@ struct CreateIndexGlobalState : public duckdb::GlobalSinkState {
...
+  bool is_glob_external = false;
   duckdb::idx_t file_row_number_col_idx = 0;
+  duckdb::idx_t file_index_col_idx = 0;
   int64_t external_row_counter = 0;
   bool has_generated_pk_col = false;
   duckdb::idx_t generated_pk_col_idx = 0;
+  // No PK column in the chunk -- Sink synthesises a monotonic counter.
+  bool is_view_synth_pk = false;
+  int64_t view_row_counter = 0;
+  bool is_view_rocksdb_pk = false;
```
gnusi: regroup bool?

gnusi (approved):

PR #595, opened 2026-05-01
feat: More FTS sugar

gnusi:
## Summary

Two-phase expansion of FTS surface area on top of the existing `col @@ ts_*(...)` machinery.

### 1. SQL-native sugar predicates

New top-level predicate functions that rewrite at filter-build time to `col @@ ts_*(...)`. Same lowering, friendlier surface for users who haven't discovered the `@@` constructors.

| Predicate | Lowers to |
|---|---|
| `phrase_matches(col, text)` | `col @@ ts_phrase(text)` |
| `ngram_matches(col, text [, threshold])` | `col @@ ts_ngram(text, threshold)` |
| `levenshtein_matches(col, text [, distance, with_transpositions, prefix])` | `col @@ ts_levenshtein(...)` |
| `has_any_token(col, ARRAY[...])` | `col @@ ts_any_of(...)` |
| `has_all_tokens(col, ARRAY[...])` | `col @@ ts_all_of(...)` |

### 2. DuckDB built-in claims

Recognised by name in the filter builder and rewritten to `col @@ ts_*(...)`. Each claim requires the column to be inverted-indexed with the **keyword analyzer** (`irs::StringTokenizer`) so the indexed term equals the column value byte-for-byte and predicate semantics map exactly to per-row evaluation. On a segmenting analyzer the claim is silently declined (returns `Result` error) and DuckDB evaluates per-row with a hint pointing to the equivalent `ts_*` form.

| DuckDB built-in | Lowers to |
|---|---|
| `contains(col, s)` | `col @@ ts_like('%' \|\| escape(s) \|\| '%')` |
| `starts_with` / `^@` / `prefix` | `col @@ ts_starts_with(p)` |
| `ends_with` / `suffix` | `col @@ ts_like('%' \|\| escape(s))` |
| `regexp_matches` / `regexp_like` (2-arg) | `col @@ ts_regexp(re)` |
| `~~` / `like_escape` | `col @@ ts_like(pattern)` (3-arg form normalised by extracting `escape_char`) |

### DIFF
--- a/server/connector/functions/string.cpp
+++ b/server/connector/functions/string.cpp
@@ -769,81 +769,6 @@ void LikeEscapeFunction(duckdb::DataChunk& args, duckdb::ExpressionState&,
     });
 }
 
-// similar_to_escape(pattern[, escape]) -> varchar
-// Converts a SQL SIMILAR TO pattern into a POSIX regex.
-// Ported from PG's similar_escape in regexp.c.
-std::string SimilarToEscapePattern(std::string_view pattern, char escape_char) {
-  std::string result;
-  result.reserve(pattern.size() + 6);
-
-  bool afterescape = false;
-  int nquotes = 0;
-  int bracket_depth = 0;
-  int charclass_pos = 0;
-
-  result += "^(?:";
-
-  for (size_t i = 0; i < pattern.size(); ++i) {
-    char pchar = pattern[i];
-
-    if (afterescape) {
-      if (pchar == '"' && bracket_depth < 1) {
-        if (nquotes == 0) {
-          result += "){1,1}?(";
-        } else if (nquotes == 1) {
-          result += "){1,1}(?:";
-        } else {
-          THROW_SQL_ERROR(
-            ERR_CODE(ERRCODE_INVALID_USE_OF_ESCAPE_CHARACTER),
-            ERR_MSG("SQL regular expression may not contain more than "
-                    "two escape-double-quote separators"));
-        }
-        nquotes++;
-      } else {
-        result += '\\';
-        result += pchar;
-        charclass_pos = 3;
-      }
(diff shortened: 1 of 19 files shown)


server/connector/search_filter_builder.cpp
```diff
@@ -1222,77 +1002,461 @@ bool TryDispatchSqlBoostCast(irs::BooleanFilter& filter,
...
+      return {};
+    case duckdb::LogicalTypeId::TINYINT:
+    case duckdb::LogicalTypeId::SMALLINT:
+    case duckdb::LogicalTypeId::INTEGER:
+    case duckdb::LogicalTypeId::BIGINT:
+    case duckdb::LogicalTypeId::FLOAT:
+    case duckdb::LogicalTypeId::DOUBLE:
+    case duckdb::LogicalTypeId::DATE:
+    case duckdb::LogicalTypeId::TIMESTAMP:
+    case duckdb::LogicalTypeId::TIMESTAMP_TZ:
```
MBkkt, line 1249: I doubt we want such support for this ttypes, better explicit error about this?

server/connector/functions/string.cpp
```diff
@@ -769,81 +769,6 @@ void LikeEscapeFunction(duckdb::DataChunk& args, duckdb::ExpressionState&,
     });
 }
 
-// similar_to_escape(pattern[, escape]) -> varchar
-// Converts a SQL SIMILAR TO pattern into a POSIX regex.
-// Ported from PG's similar_escape in regexp.c.
-std::string SimilarToEscapePattern(std::string_view pattern, char escape_char) {
```
MBkkt, line 775: because it's just moved, it's quite difficult to check actual diff, if presented

MBkkt (approved):

PR #599, opened 2026-05-04
[WIP] feat: Adding executor to IndexWriter

mag7ster:


MBkkt (commented):
Lets use it for parallel deletes (only for search deletes in existing segments).
And write microbench for this

Issue #597, opened 2026-05-04
Json indexing follow ups


mkornaukhov:
- [x] Do no store json path explicitly, it should be some kind of expression that may be executed on rows, or even separate column
- [ ] Index `duckdb::variant` instead of json, prohibit json later
- [ ] Support indexing complex types (arrays, objects). It may have some options (like `includeAllFields`), so we need to make json as a separate opclass, so we need new mechanism to create opclass constructors

PR #607, opened 2026-05-05
misc: Improve json indexing

mkornaukhov:
## General
Fixes some bullets from #597
- Replaces the path-vector representation with a normalized serialized `BoundExpression` shared by `CREATE INDEX`, `INSERT`/`UPDATE`/`DELETE` and WAL recovery, so the field-name bytes match across binders. 
- Adds `#>` / `#>>` operator support, a per-chunk JSON-extract evaluation pipeline

## Limitation
- Only text extractors at root level: as a result, JSON null and SQL null are not distinguished and stored in the same field
- Complex types (arrays and objects) are not supported for indexed expression

Follow ups: #597, #667


server/catalog/inverted_index.h
```diff
@@ -93,11 +94,13 @@ class InvertedIndex final : public Index {
...
     const std::shared_ptr<const Snapshot>& snapshot,
-    catalog::Column::Id columnd_id) const;
+    catalog::Column::Id column_id) const;
 
+  // Resolves a configured JSON-path entry by its normalized serialized
+  // expression and returns the per-path tokenizer.
   std::optional<ColumnTokenizer> GetJsonPathTokenizer(
     const std::shared_ptr<const Snapshot>& snapshot,
-    catalog::Column::Id column_id, std::span<const std::string> path) const;
+    catalog::Column::Id column_id, const std::string& serialized_expr) const;
```
MBkkt: std::string_view?

server/catalog/inverted_index.cpp
```diff
@@ -138,23 +138,22 @@ ColumnTokenizer InvertedIndex::GetColumnTokenizer(
...
 std::optional<ColumnTokenizer> InvertedIndex::GetJsonPathTokenizer(
   const std::shared_ptr<const Snapshot>& snapshot,
-  catalog::Column::Id column_id, std::span<const std::string> path) const {
+  catalog::Column::Id column_id, const std::string& serialized_expr) const {
   const auto* info = FindColumnInfo(column_id);
   if (!info) {
     return std::nullopt;
   }
-
   if (auto it = absl::c_find_if(info->json_paths,
```
MBkkt: hashset?

server/catalog/index.h
```diff
@@ -47,7 +47,8 @@ struct CreateIndexColumn {
   const catalog::Column* catalog_column{nullptr};
   std::string_view name;
   std::string opclass;
-  std::vector<std::string> json_path;
+  // non empty -> indexing json expression
+  std::string serialized_json_expr;
```
MBkkt: why json expr, why not any expr?

server/connector/optimizer/iresearch_plan.cpp
```diff
@@ -316,17 +317,21 @@ bool RewriteFilterColumnRefs(
...
-  std::function<std::optional<catalog::ColumnTokenizer>(
-    catalog::Column::Id, std::span<const std::string>)>
+  // resolved against the catalog, or nullopt if the column is not indexed
+  // at `serialized_expr`. The serialized form is the same normalised-bytes
+  // representation CREATE INDEX persisted in the catalog -- callers
+  // produce it via `connector::SerializeBoundExpression` on a
+  // `connector::NormalizeBoundExpression`-normalised tree.
+  std::function<std::optional<catalog::ColumnTokenizer>(catalog::Column::Id,
+                                                        const std::string&)>
     json_path_tokenizer_provider;
```
MBkkt: string_view + maybe optional is some other form?

server/pg/json.tpp
```diff
@@ -121,6 +121,13 @@ class JsonParser {
       if (value.type() == simdjson::ondemand::json_type::string) {
         return value.get_string().value();
       }
+      // PG `->>` returns SQL NULL for JSON null leaves -- mirror that here
+      // by returning an empty string_view with `data()==nullptr`, which the
+      // caller (JsonExtractFieldTextFunction et al.) detects via
+      // `str.empty() && !str.data()` and turns into mask.SetInvalid.
```
MBkkt:
```suggestion
```

server/catalog/inverted_index.cpp
mkornaukhov: fixed

server/connector/optimizer/iresearch_plan.cpp
mkornaukhov: fixed

server/connector/search_sink_writer.cpp
```diff
@@ -114,18 +112,41 @@ inline constexpr bool kIsNumericKind =
   Kind == duckdb::LogicalTypeId::TIMESTAMP ||
   Kind == duckdb::LogicalTypeId::TIMESTAMP_TZ;
 
+// Strip the kStringPrefix byte (added on the Insert path) when present.
+// Mirrors the slice-shape handling in WriteStringValue.
```
MBkkt: please avoid such comments

server/connector/search_sink_writer.cpp
```diff
@@ -214,6 +230,123 @@ bool SearchSinkInsertBaseImpl::SwitchColumnImpl(const ColumnDescriptor& col) {
...
+#endif
+    return false;
+  }
+
+  switch (return_type.id()) {
+    case duckdb::LogicalTypeId::VARCHAR:
+      SetupColumnWriter<duckdb::LogicalTypeId::VARCHAR>(column_id, have_nulls,
+                                                        serialized_expr);
+      // Also emit numeric/bool variants so cast filters (e.g. ::int) match.
+      SetupExprAuxTypedFields(column_id, serialized_expr);
```
MBkkt, line 249: This doesn't look correct, we should do this only for json, right?

server/connector/search_sink_writer.cpp
```diff
@@ -214,6 +230,123 @@ bool SearchSinkInsertBaseImpl::SwitchColumnImpl(const ColumnDescriptor& col) {
   return true;
 }
 
+bool SearchSinkInsertBaseImpl::SwitchExpressionImpl(
+  const ExpressionDescriptor& json_desc) {
+  const auto column_id = json_desc.column_id;
+  const auto& return_type = json_desc.type;
+  const auto have_nulls = json_desc.have_nulls;
+  const auto serialized_expr = json_desc.serialized_expr;
```
MBkkt: expr_desc?

server/connector/search_sink_writer.cpp
mkornaukhov: We really should do this for json-typed expressions only. Just for now we support json text extractors only, so we setup aux fields and parse numberic/boolean. In next iteration, we will:
- support json typed expr and only their use such a logics
- otherwise we will see at logical type and init only corresponding fields

mkornaukhov:
Closed in favor of #670

PR #630, opened 2026-05-07
[WIP] perf: New deleted docs implementation

the-devid:
Changing DocumentMask from being type alias to interface and a bunch of realizations:
- `absl::flat_hash_set` brings overheads especially when there are a lot of deleted documents in a segment
- We need to choose optimal implementation of document mask based on deletion ratio when reading/creating/updating segment meta
- We want to eliminate virtual calls on a query hot path when masking documents but still store different implementations as a field in a filter/doc-iterator


libs/iresearch/include/iresearch/index/document_mask.hpp
```diff
@@ -0,0 +1,215 @@
...
+  const DocumentMask* operator->() const { return mask.get(); }
+  const DocumentMask& operator*() const { return *mask; }
+};
+
+inline DocumentMaskKind ChooseImmutableRepresentation(
+  size_t doc_count, size_t deleted_doc_count) {
+  if (deleted_doc_count < doc_count / 100) {  // 0 <= x <1% of documents
+    return DocumentMaskKind::DeletedHashSet;
+  } else {  // 1 <= x <= 100% of documents
+    return DocumentMaskKind::DenseBitset;
```
MBkkt: >= 99% also sparse but set of alive docs

Also for dense bitset, maybe it's better to store/load alive bitset

PR #633, opened 2026-05-08
feat: Column with index only storage

Dronplane:
Adds sdb_indexonly option for column. That makes WAL-only column. The operations exists only at WAL (custom LogData markers for put/delete) + inverted index indexes value. So column is searchable but not "readable".
 "Read" part would be handled later by iresearch backed tables.
 VIRTUAL columns can not be indexonly


server/catalog/local_catalog.cpp
```diff
@@ -1326,6 +1326,11 @@ Result LocalCatalog::CreateSecondaryIndex(
       return Result{ERROR_BAD_PARAMETER, "column \"", c.name,
                     "\" does not exist"};
     }
+    if (it->store_mode == ColumnStoreMode::kIndexOnly) {
+      return Result{ERROR_BAD_PARAMETER, "cannot include column \"", c.name,
+                    "\" in a secondary index: column has sdb_indexonly "
```
MBkkt, line 1414:
```suggestion
                    "\" in a secondary index: column has sdb_index_only "
```
maybe `sdb_index_only`? or even just `index_only`

tests/sqllogic/sdb/pg/index/inverted_index.test
```diff
@@ -428,9 +428,9 @@ EXPLAIN ANALYZE CREATE INDEX donsimon_idx ON biksalen_memes USING inverted(b tes
 │   Table: biksalen_memes   │
 │                           │
 │        Projections:       │
-│             a             │
 │             b             │
 │             c             │
+│             a             │
 │                           │
```
MBkkt, line 434: looks strange, why is it changed, should be in other order I think?

Dronplane: It was changed because I want to change how index backfill read works. For now it always reads all table columns. That is messed up by having "index only column" that can not be read yet. But instead of kludge I want to change this in general - Index backfill will read only what is needed by index + PK columns that were not explicitly included by index but needed to make index record are appended at the end. So that is why order was changed. "a" here is PK  but not needed by index. So it is added later.

server/catalog/local_catalog.cpp
Dronplane: More honest would be "sdb_inverted_index_only" but that is a bit long. I wanted to have "sdb_" prefix to not conflict with any future options.

MBkkt: it's not inverted index only, it's just columns that will be written only in wal. In general name doesn't really matter, I think we will use this later internally for iresearch based tables, before we drop rocksdb WAL

server/catalog/table_options.h
MBkkt, line 54: index only cannot be virtual

server/connector/duckdb_index_utils.cpp
```diff
@@ -224,4 +224,57 @@ CreateDuckDBIndexWriters<DuckDBWriteKind::Update>(
...
+bool NeedsRowDeleteMarkers(
+  std::span<const std::shared_ptr<catalog::Index>> indexes,
+  std::span<const catalog::Column> columns) {
+  for (const auto& index : indexes) {
+    if (index->GetType() != catalog::ObjectType::InvertedIndex) {
+      continue;
+    }
+    bool all_indexonly = true;
+    for (auto col_id : index->GetColumnIds()) {
+      auto it = std::ranges::find_if(
```
MBkkt:
```suggestion
      auto it = absl::c_find_if(
```

server/connector/duckdb_index_utils.cpp
```diff
@@ -224,4 +224,57 @@ CreateDuckDBIndexWriters<DuckDBWriteKind::Update>(
...
+}
+
+std::vector<size_t> BuildCreateIndexProjection(
+  std::span<const catalog::Column> columns,
+  std::span<const catalog::Column::Id> pk_column_ids,
+  std::span<const duckdb::idx_t> index_column_positions) {
+  std::vector<size_t> projection;
+  projection.reserve(index_column_positions.size() + pk_column_ids.size());
+  containers::FlatHashSet<size_t> seen;
+  seen.reserve(index_column_positions.size() + pk_column_ids.size());
```
MBkkt: put everything => sort => unique

server/connector/duckdb_scan_base.cpp
```diff
@@ -138,6 +138,42 @@ void InitCommonState(CommonScanGlobalState& state,
...
+          const auto& tbd = bind_data.As<TableScanBindData>();
+          for (const auto& col : tbd.table->Columns()) {
+            if (col.id != catalog_col_id ||
+                col.store_mode != catalog::ColumnStoreMode::kIndexOnly) {
+              continue;
+            }
+            if (bind_data.is_create_index) {
+              state.finished = true;
+            } else {
+              throw duckdb::CatalogException(
```
MBkkt: THROW_SQL_ERROR

server/search/inverted_index_shard.cpp
```diff
@@ -480,6 +480,16 @@ InvertedIndexShard::ResultWithTime InvertedIndexShard::CommitUnsafe(
     // intentionally mark the commit as failed
     result.reset(ERROR_DEBUG);
   }
+  SDB_IF_FAILURE("Search::CrashAfterCommit") {
+    // Crash AFTER a successful iresearch commit (segment + committed_tick
+    // are durable on disk). On restart, wal_recovery starts from the new
+    // committed_tick and must NOT re-apply any IndexOnly markers that
+    // sit at sequence numbers <= committed_tick -- regression test in
+    // tests/sqllogic/recovery/.
+    if (result.ok()) {
```
MBkkt: why only ok?

MBkkt (approved):

MBkkt:
Also update tests

server/connector/duckdb_scan_base.cpp
Dronplane: Done

server/connector/duckdb_index_utils.cpp
Dronplane: done

server/connector/duckdb_index_utils.cpp
Dronplane: Done

server/catalog/table_options.h
Dronplane: Done

server/search/inverted_index_shard.cpp
Dronplane: It should be ok. if removed

PR #634, opened 2026-05-08
Support text search inside ann query

codeworse:
Pushes iresearch-indexable predicates out of the row-by-row ANNFilter and into a cheap TextScanFilter evaluated against the inverted index, so HNSW candidates get rejected before any RocksDB lookup.
Example:
```sql
SELECT id FROM mixed_idx WHERE body @@ ts_phrase('red') ORDER BY emb <-> :query LIMIT :k;
```

### DIFF
--- a/server/connector/duckdb_ann_filter.cpp
+++ b/server/connector/duckdb_ann_filter.cpp
@@ -20,9 +20,6 @@
 
 #include "connector/duckdb_ann_filter.h"
 
-#include <algorithm>
-#include <array>
-
 #include "basics/assert.h"
 #include "connector/duckdb_client_state.h"
 #include "connector/duckdb_table_function.h"
@@ -170,24 +167,21 @@ bool TextScanFilter::Accept(faiss::idx_t id) const {
   return _it->seek(doc_id) == doc_id;
 }
 
-CompositeScanFilter::CompositeScanFilter(
-  std::vector<std::unique_ptr<ScanFilter>> fs)
-  : _filters{std::move(fs)} {
-  std::sort(_filters.begin(), _filters.end(),
-            [](const auto& a, const auto& b) { return a->Cost() < b->Cost(); });
-}
-
 void CompositeScanFilter::Reset(const irs::SubReader& segment) {
-  for (auto& f : _filters) {
-    f->Reset(segment);
+  if (_text) {
+    _text->Reset(segment);
+  }
+  if (_ann) {
+    _ann->Reset(segment);
   }
 }
 
 bool CompositeScanFilter::is_member(faiss::idx_t id) const {
-  for (const auto& f : _filters) {
-    if (!f->Accept(id)) {
-      return false;
-    }
+  if (_text && !_text->Accept(id)) {
+    return false;
+  }
+  if (_ann && !_ann->Accept(id)) {
+    return false;
   }
   return true;
 }
--- a/server/connector/duckdb_ann_filter.cpp
+++ b/server/connector/duckdb_ann_filter.cpp
@@ -20,6 +20,7 @@
 
 #include "connector/duckdb_ann_filter.h"
 
+#include <algorithm>
 #include <array>
 
 #include "basics/assert.h"
@@ -79,8 +80,8 @@ void InitAnnFilterContext(
   });
 }
 
-ANNFilter::ANNFilter(const ANNFilterContext& ctx, const irs::SubReader& segment)
-  : _ctx{ctx}, _segment{segment}, _executor{ctx.context} {
+ANNFilter::ANNFilter(const ANNFilterContext& ctx)
+  : _ctx{ctx}, _executor{ctx.context} {
   _executor.AddExpression(*ctx.filter_expr);
   duckdb::vector<duckdb::LogicalType> scratch_types{ctx.filter_types.begin(),
                                                     ctx.filter_types.end()};
@@ -89,12 +90,14 @@ ANNFilter::ANNFilter(const ANNFilterContext& ctx, const irs::SubReader& segment)
   duckdb::vector<duckdb::LogicalType> bool_types{1,
                                                  duckdb::LogicalType::BOOLEAN};
   _bool_out.Initialize(duckdb::Allocator::Get(ctx.context), bool_types);
+}
 
-  auto opened = OpenSegmentPkIterator(_segment, _it);
+void ANNFilter::Reset(const irs::SubReader& segment) {
+  auto opened = OpenSegmentPkIterator(segment, _it);
   SDB_ASSERT(opened);
 }
 
-bool ANNFilter::is_member(faiss::idx_t id) const {
+bool ANNFilter::Accept(faiss::idx_t id) const {
   SDB_ASSERT(_it);
   auto [_, doc_id] = irs::UnpackSegmentWithDoc(id);
   if (_it.iter->value() > doc_id) {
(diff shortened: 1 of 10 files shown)


server/connector/optimizer/iresearch_plan.cpp
```diff
@@ -447,6 +640,11 @@ bool TryAnnTopk(duckdb::unique_ptr<duckdb::LogicalOperator>& plan) {
     ann->filter_expression =
       CombineFilterExpressions(std::move(rewritten_exprs));
     ann->filter_column_ids = std::move(filter_col_ids);
+    if (any_claimed) {
+      ann->text_filter_summary =
+        irs::ToStringDemangled(and_root, MakeColumnNameLookup(bind_data));
```
gnusi: maybe make it lazy? I think we only need it in case of `EXPLAIN`

server/connector/optimizer/iresearch_plan.cpp
```diff
@@ -867,17 +987,8 @@ bool TrySearchFilter(duckdb::unique_ptr<duckdb::LogicalOperator>& plan,
...
-    static thread_local std::string fallback;
-    auto name = bind_data.ColumnNameById(col_id);
-    if (!name.empty()) {
-      return name;
-    }
-    fallback = absl::StrCat("col", col_id);
-    return fallback;
-  };
-  std::string filter_summary = irs::ToStringDemangled(*root, col_name_lookup);
+  std::string filter_summary =
```
gnusi, line 987: maybe make it lazy? I think we only need it in case of `EXPLAIN`

server/connector/duckdb_ann_filter.cpp
```diff
@@ -154,4 +154,36 @@ bool ANNFilter::is_member(faiss::idx_t id) const {
...
+  : _query{query} {}
+
+void TextScanFilter::Reset(const irs::SubReader& segment) {
+  _it = _query.execute({.segment = segment});
+  SDB_ASSERT(_it);
+}
+
+bool TextScanFilter::Accept(faiss::idx_t id) const {
+  auto [_, doc_id] = irs::UnpackSegmentWithDoc(id);
+  return _it->seek(doc_id) == doc_id;
```
gnusi, line 167: as discussed I think we can use LazySeek,
LazySeek improvement can be done separately

server/connector/duckdb_ann_filter.h
```diff
@@ -82,4 +82,34 @@ void InitAnnFilterContext(
...
+class TextScanFilter {
+ public:
+  TextScanFilter(const irs::Filter::Query& query);
+
+  bool Accept(faiss::idx_t id) const;
+  void Reset(const irs::SubReader& segment);
+
+ private:
+  const irs::Filter::Query& _query;
+  mutable irs::DocIterator::ptr _it;
```
gnusi, line 94: Maybe down cast?

server/connector/optimizer/iresearch_plan.cpp
```diff
@@ -308,7 +311,165 @@ bool RewriteFilterColumnRefs(
...
+  }
+  ctx.table_index = get.table_index;
+  ctx.projected_column_ids = projected_ids_storage;
+  bind_data.IterateColumns(
+    [&](catalog::Column::Id id, const duckdb::LogicalType& type) {
+      ctx.column_type_by_id.emplace(id, type);
+    });
+  for (auto col_id : resolved.index->GetColumnIds()) {
+    ctx.indexed_column_ids.insert(col_id);
+  }
```
gnusi: auto& columns = resolved.index->GetColumnIds();
ctx.indexed_column_ids.insert(columns.begin(), columns.end());

server/connector/optimizer/iresearch_plan.cpp
```diff
@@ -427,12 +588,44 @@ bool TryAnnTopk(duckdb::unique_ptr<duckdb::LogicalOperator>& plan) {
...
+  for (auto* f : residual_filters) {
+    std::vector<bool> claimed(f->expressions.size(), false);
+    for (size_t i = 0; i < f->expressions.size(); ++i) {
+      if (TryClaimIresearchConjunct(and_root, f->expressions[i], getter,
+                                    json_getter, options)) {
+        claimed[i] = true;
+        any_claimed = true;
+      }
+    }
+    claimed_per_filter.push_back(std::move(claimed));
```
gnusi: emplace_back

server/connector/optimizer/iresearch_plan.cpp
```diff
@@ -632,6 +857,11 @@ bool TryAnnRange(duckdb::unique_ptr<duckdb::LogicalOperator>& plan) {
     rss->filter_expression =
       CombineFilterExpressions(std::move(rewritten_exprs));
     rss->filter_column_ids = std::move(filter_col_ids);
+    if (any_claimed) {
+      rss->text_filter_summary =
```
gnusi: make it lazy?

server/connector/optimizer/iresearch_plan.cpp
MBkkt: Yes, it's meh and should be fixed, the issue if I remember correctly that these explain-strings relies on some data, that became freed after this code gone

server/connector/duckdb_ann_filter.h
```diff
@@ -52,26 +54,24 @@ struct ANNFilterContext {
   std::vector<catalog::Column::Id> filter_column_ids;
 };
 
-// faiss IDSelector: per HNSW candidate, materializes the row's filter
-// columns and evaluates the filter expressions to gate inclusion.
-class ANNFilter final : public faiss::IDSelector {
+// Row-by-row filter: per HNSW candidate, materializes the row's filter
+// columns and evaluates the filter expression. Expensive (RocksDB lookup
+// per candidate).
```
MBkkt, line 59: does it work for views? if so it's not only rocksdb, also can be parquet/anything else.
If there's no such tests, add plz

server/connector/duckdb_ann_filter.cpp
codeworse: https://github.com/serenedb/serenedb/issues/635

server/connector/duckdb_ann_filter.h
codeworse: added tests into vector_search_view.test_slow

gnusi (approved):

server/connector/duckdb_ann_filter.h
MBkkt: so comment should be updated???

server/connector/duckdb_table_function.h
```diff
@@ -229,6 +229,9 @@ struct VectorSearchScan : ScanSource {
   std::vector<float> query_vector;
   duckdb::unique_ptr<duckdb::Expression> filter_expression;
   std::vector<catalog::Column::Id> filter_column_ids;
+
+  std::unique_ptr<irs::Filter> stored_text_filter;
+  irs::Filter* text_filter_root = nullptr;
```
MBkkt, line 234: why do we need both variables

server/connector/duckdb_ann_filter.h
```diff
@@ -82,4 +82,34 @@ void InitAnnFilterContext(
...
+  void EnableAnn(const ANNFilterContext& ctx) { _ann.emplace(ctx); }
+
+  bool Empty() const { return !_text && !_ann; }
+
+  bool is_member(faiss::idx_t id) const final;
+
+  void Reset(const irs::SubReader& reader);
+
+ private:
+  std::optional<TextScanFilter> _text;
```
MBkkt, line 111: better naming, maybe: IndexFilter + RelationFilter?

server/connector/duckdb_ann_filter.h
```diff
@@ -52,26 +54,24 @@ struct ANNFilterContext {
   std::vector<catalog::Column::Id> filter_column_ids;
 };
 
-// faiss IDSelector: per HNSW candidate, materializes the row's filter
-// columns and evaluates the filter expressions to gate inclusion.
-class ANNFilter final : public faiss::IDSelector {
+// Row-by-row filter: per HNSW candidate, materializes the row's filter
+// columns and evaluates the filter expression. Expensive (RocksDB lookup
+// per candidate).
+class ANNFilter {
```
MBkkt, line 60: filters, both should have const methods, because all members are immutable. Instead whole filter should be mutable

MBkkt (approved):
please fix in followups

PR #642, opened 2026-05-10
fix: Sequences leak 

pashandor789:


server/catalog/local_catalog.cpp
```diff
@@ -2323,14 +2323,14 @@ Result LocalCatalog::DropSequence(std::string_view database,
...
+      ctx.DropDefinition(*schema_id, ObjectType::Sequence, seq_id);
+      ctx.DropSequence(seq_id);
+    });
     if (!r.ok()) {
       return r;
     }
-    if (auto cr = _engine->DropSequence(seq->GetId()); !cr.ok()) {
-      return cr;
-    }
     clone->UnregisterObject(std::move(seq), *schema_id);
```
MBkkt, line 2334: drop table
1. if presented drop some childs
2. if presented drop some childs
3. final step: drop definition

MBkkt (approved):
Please create an issue about better sequneces drop

PR #653, opened 2026-05-11
fix: don't mutate shared LogicalType alias in PgSqlType ctor

MBkkt:
`CreateTypeInfo::Copy()` shallow-copies the embedded `LogicalType`, so the `type_info_` shared_ptr is shared across copies. Calling `SetAlias` on the nested type rewrote that shared state, leaving every cloned PgSqlType holding the most recent alias.

Build a fresh `ExtraTypeInfo` (DeepCopy from the existing one when present, otherwise generic) and assign a new LogicalType with that unique shared_ptr.


mkornaukhov (approved):
OK

PR #655, opened 2026-05-11
feat: surface CREATE INDEX progress via pg_stat_progress_create_index

MBkkt:
Add `IndexProgressReporter::SetIndexRelid` so the relid can be patched in mid-flight (it isn't known when the build starts -- the catalog assigns it a moment later). `SereneDBPhysicalCreateIndex` now owns a reporter on the global sink state and advances phases: `Initializing` during planning, `BuildingIndex` once the catalog row exists, `Committing` while iresearch runs CommitWait, `Finalizing` for the tombstone-removal step. `ReportBatch` fires per Sink chunk so total_rows / done_rows reflect ingest progress.


codeworse (approved):
Is there a way to test it using sqllogictest?

MBkkt:
@codeworse yes, but it will be in original PR (just create index in one connection, async, in main connection run get from pg view/table)

PR #662, opened 2026-05-12
fix: load view-backed indexes on startup

MBkkt:
## Summary

Startup walked indexes parented by tables only, leaving view-parented
rows in rocksdb orphaned. After restart `SELECT FROM <idx>` errored
("Table with name <idx> does not exist"), and `CREATE INDEX IF NOT
EXISTS` then silently rebuilt a duplicate segment set under the same
name because the existence check found nothing.

`RegisterViews` now mirrors `AddTable`: after registering each view,
collect its index tombstones and call the existing `RegisterIndexes`
with the view id as the relation parent. `RelationDependency` is already
the common base of `TableDependency` / `ViewDependency`, so the snapshot
APIs (`VisitIndexes`, `GetIndexesByRelation`) already handled
view-parented indexes — only the recovery walk was missing.

## Test plan

- [x] New recovery test `tests/sqllogic/recovery/catalog_view_inverted_index.test`:
  create view → create inverted index on view → crash → verify alias
  still resolves, `pg_class` lists it, and `CREATE INDEX IF NOT EXISTS`
  is a no-op (no segment doubling).
- [x] Manual repro against `build/bin/serened` on a clean data dir:
  without the patch, the test fails at the post-restart `SELECT FROM
  vi_idx` with the same error the bug report shows; with the patch,
  it returns the same row count as before the crash.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -16,8 +16,13 @@ Pick a free `<port>`. You own it for the session -- when you're done, kill the
 process so the next contributor / agent doesn't inherit a stale serened.
 
 ```bash
+# If you background the process (`&`), add --log.foreground-tty=true so
+# logs still print to your terminal -- otherwise serened drops them.
 ./build/bin/serened ./build_data --server.endpoint='pgsql+tcp://0.0.0.0:<port>'
-psql -h 127.0.0.1 -p <port> -U postgres
+
+# Default database is `postgres` (serenedb has no separate default).
+psql -h 127.0.0.1 -p <port> -U postgres -d postgres
+
 # when finished:
 kill -9 $(lsof -t -i:<port>) 2>/dev/null
 rm -rf build_data    # only if you don't need the datadir afterwards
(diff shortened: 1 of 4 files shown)


CLAUDE.md
```diff
@@ -0,0 +1,29 @@
...
+
+Read `CONTRIBUTING.md` -- it covers build, tests, branches, commits, PRs, and
+C++ style. This file only flags traps that aren't in there.
+
+## Before writing tests
+
+- Sqllogic: read a sibling `.test` first. Control directives, retry patterns,
+  connection naming, and the right subtree (`any/sdb/pg/recovery`) are set by
+  example, not docs.
+- gtest / microbench: mirror an existing one in the same `tests/<area>/` subtree.
```
Dronplane, line 11: I'd add a note about using "multiline error" mode in testing statement/query errors. At least for me it is necessary to pinpoint it every time.
Also worth noting - avoid term "pudge" in test dara

Dronplane (approved):
LGTM

CLAUDE.md
MBkkt: I or Pasha will add format for sqlogictest-rs, no need for agent help in this area

MBkkt: so every test will be multiline beauty sql about vedernikoff and pudge

PR #666, opened 2026-05-13
chore: Prune views columns for index backfill

Dronplane:
Forces view backed indexes to do pruning of unused columns for backfill task.
The problem was that Create Index binder uses surrogate Table to reference its columns. That is useful so we can create index over view without building full plan and wiring all columns like we do when we need to execute SELECT. And that is specifically handled in the RemoveUnusedColumns rule in the optimizer - CREATE INDEX is not projection boundary. So we still on the "root" pruner that effectively counts columns referenced at root level as "used" and prunes nothing. That makes views to read all columns  no matter if they are needed for indexing or not. Index writer then just happily skips unindexed columns and from user perspective indexing works. 

PR addresses this in two steps
First it gathers what index indeed needs - all directly indexed columns + PK columns + service columns for iceberg/parquet cases.
if not entire view output is needed it creates LIST with used columns and stores it in the options so physical index create later knows that it has reduced set of columns in chunks.

Secondly we insert an additional Projection node between create index and other part of the view. That projection references as identity only needed columns. That makes optimizer to "see the usage boundary" and prune unused columns in a normal way. After playing its role projection is anyway discarded by optimizer as it is just identity transfer.

That is done for both fast-path and regular views as optimizer can handle all the query forms and we are not limited here.


server/connector/duckdb_catalog.cpp
```diff
@@ -1109,7 +1246,10 @@ duckdb::unique_ptr<duckdb::LogicalOperator> SereneDBCatalog::BindCreateIndex(
             duckdb::Value("rocksdb_rowid");
           break;
         default: {
-          const auto vcols = BackfillPkVirtualColumns(*view_fast_path);
+          SDB_ASSERT(vcols_opt.has_value(),
```
MBkkt:
```suggestion
          SDB_ASSERT(vcols_opt,
```

MBkkt (approved):
Please cleanup comments

PR #671, opened 2026-05-15
chore: Allow using AT as column name and table  alias

Dronplane:
DuckDB grammar PR https://github.com/serenedb/duckdb/pull/16
Allows using AT as table alias - needed to remove "at"->"att" kludge in system view + allow running client tools that use "AT" as alias in queries.


scripts/update_system_catalog.py
```diff
@@ -479,7 +479,7 @@ def qualify_info_schema_refs(sql, view_names, func_names):
             sql
         )
     # "at" is a reserved keyword in DuckDB (time-travel)
-    sql = re.sub(r'\bat\b', 'att', sql)
+    #sql = re.sub(r'\bat\b', 'att', sql)
```
MBkkt: remove these lines

MBkkt (approved):

PR #675, opened 2026-05-15
feat: support cascade drop

pashandor789:


server/pg/pg_catalog/pg_class.cpp
```diff
@@ -74,8 +74,8 @@ constexpr uint64_t kNullMask = MaskFromNonNulls({
 
 }  // namespace
 
-PgClass MakeBaseRow(ObjectId schema_id, const catalog::SchemaObject& object) {
-  auto owner = object.GetOwnerId();
+PgClass MakeBaseRow(ObjectId schema_id, const catalog::Object& object) {
+  auto owner = object.GetParentId();
   if (!owner) {
     owner = id::kRootUser;
```
MBkkt, line 80: it's not correct, here they wanted owner, right now rbac not implemented so it's ok to always return kRootId from owner, but Parent is probably not correct?

Please add tests for this and other catalogs that do joins on this owner id, and see it's wrong

MBkkt (approved):

PR #683, opened 2026-05-18
Embeddings demos

codeworse:


server/connector/functions/embedding/provider_openai.cpp
```diff
@@ -0,0 +1,213 @@
...
+          throw duckdb::InvalidInputException(
+            "OpenAI embedding contains non-numeric value: %s", body);
+        }
+        data[size++] = static_cast<float>(d);
+      }
+      parsed++;
+      next_row++;
+    }
+    if (parsed != expected) {
+      throw duckdb::InvalidInputException(
```
MBkkt: throw sql error instead of duckdb exceptions for our code plz

MBkkt (approved):

PR #703, opened 2026-05-21
fix: Fix remote JSON read

codeworse:
duckdb PR: https://github.com/serenedb/duckdb/pull/19

In duckdb's json_transform.cpp::GetValueString, YYJSON_TYPE_RAW values were wrapped in a string_t pointing directly at yyjson's internal buffer. With remote JSON reads that buffer can be recycled before the string is consumed, producing corrupted values. The fix copies the bytes into an allocator-owned buffer first, so the string_t stays valid.


third_party/duckdb
MBkkt, line 1: patch looks wrong

server/connector/search_sink_writer.cpp
```diff
@@ -739,7 +739,9 @@ void SearchSinkInsertBaseImpl::InitImpl(size_t batch_size) {
 SearchSinkInsertBaseImpl::Field& SearchSinkInsertBaseImpl::WriteStringValue(
   std::string_view, std::span<const rocksdb::Slice> cell_slices,
   SearchSinkInsertBaseImpl::Field& field) {
-  SDB_ASSERT(!cell_slices.empty());
+  if (cell_slices.size() == 0) {
+    SDB_ASSERT(!cell_slices.empty());
+  }
```
MBkkt: just assert non-empty?!

codeworse: yes, it was just for the debug

MBkkt (approved):

Issue #317, opened 2026-03-12
Sort terms by frequency in interval phrase matching


pashandor789:
## What we already have

In `FixedPhraseFrequency::NextPosition()` there's a compile-time split:

```cpp
IRS_FORCE_INLINE uint32_t NextPosition() {
    if constexpr (HasIntervals) {
      return NextPositionGeneric();
    } else {
      return NextPositionOptimized();
    }
}
```

The non-interval path (`NextPositionOptimized`) sorts terms by doc frequency before matching — rarest term becomes the lead. The interval path (`NextPositionGeneric`) doesn't do this. It should.

## How `NextPositionOptimized` works

The idea is simple: pick the rarest term as lead so we iterate over fewer positions.

Take a phrase `"konstantin vedernikoff is intern"`. Say the document looks like this:

- `konstantin` (doc_freq=50) -> positions [3, 20, 100]
- `vedernikoff` (doc_freq=5) -> positions [4, 21]
- `is` (doc_freq=10000) -> positions [1, 5, 10, 15, 22, 50, 80, 101, ...]
- `intern` (doc_freq=30) -> positions [6, 102]

Each term knows its `lead_offset` — distance from phrase start: `konstantin=0, vedernikoff=1, is=2, intern=3`.

First we sort by doc frequency:

```
vedernikoff(5)  lead_offset=1  ← new lead
intern(30)      lead_offset=3
konstantin(50)  lead_offset=0
is(10000)       lead_offset=2
```

Then we iterate from vedernikoff (only 2 positions to check!):

```
new_lead_offset = 1

lead_pos = 4 (vedernikoff)
  intern:      (4-1)+3 = 6  -> seek(6) -> 6, match
  konstantin:  (4-1)+0 = 3  -> seek(3) -> 3, match
  is:          (4-1)+2 = 5  -> seek(5) -> 5, match
  -> match, phrase_freq = 1

lead_pos = 21 (vedernikoff)
  intern:      (21-1)+3 = 23 -> seek(23) -> 102 != 23
  -> mismatch, jump lead to seek((102-3)+1) = 100 -> EOF
  -> done, phrase_freq = 1
```

If `is` were the lead we'd be looping through thousands of positions. With sorting we check 2.

The matching itself is trivial — `sought != target` means mismatch, `goto restart`.

## How intervals work and why they don't sort yet

[...]

PR #704, opened 2026-05-21
perf: optimize intervals search (#317)

deymon-d:
1. Improved interval search algorithm – The previous unidirectional traversal has been replaced with a bidirectional approach, initiating from the rarest term within the phrase. This optimization reduces the search space and improves overall matching efficiency.
2. Added golden benchmark tests for interval search – A new suite of benchmark tests has been introduced, including support for the `--report` flag. This enables detailed time-based performance reporting, facilitating regression detection and performance tracking over time.

### DIFF
--- a/libs/iresearch/include/iresearch/search/phrase_iterator.hpp
+++ b/libs/iresearch/include/iresearch/search/phrase_iterator.hpp
@@ -132,11 +132,27 @@ class SinglePositionStrategy {
     return seek == sought;
   }
 
+  bool AdvanceIterators(bool match, PosAttr::value_t sought,
+                        const Iterator& end, Iterator& it,
+                        PosAttr::value_t& lead_bound) {
+    if (!match) {
+      SDB_ASSERT(sought + Traits::Interval(*_lead_it).lead_offset >
+                 Traits::Interval(*it).lead_offset);
+      lead_bound = sought - Traits::Interval(*it).lead_offset +
+                   Traits::Interval(*_lead_it).lead_offset;
+    }
+    _base_position = sought;
+    ++it;
+    return match;
+  }
+
   bool AdvanceIterators(bool match, PosAttr::value_t sought, const Iterator&,
                         Iterator& it) {
     if (!match) {
-      SDB_ASSERT(sought > Traits::Interval(*it).lead_offset);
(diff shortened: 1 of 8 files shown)


libs/iresearch/include/iresearch/search/phrase_iterator.hpp
```diff
@@ -375,6 +375,8 @@ class FixedPhraseFrequency {
     std::conditional_t<HasIntervals,
                        IntervalPositionStrategy<typename Positions::iterator>,
                        SinglePositionStrategy<typename Positions::iterator>>;
+  using IntervalBoundType = decltype(std::declval<TermPosition>().second.offs_min);
```
deymon-d, line 378: maybe define with using in TermInterval

MBkkt:
@deymon-d It will be nice to implement some benchmark

deymon-d:
@MBkkt , Good afternoon!


I have made benchmarks, here is a [link ](https://drive.google.com/file/d/1_MACg_a1-Fi9zwp6TJUi8l5lz1pnnL9T/view?usp=sharing)to the results. 

Link to the bench tests, they covered more stress tests than the units that are available now, I saw the cases that only they caught - https://github.com/deymon-d/serenedb/commit/c75fbbe294763bcd4dcea4dd5509b59988ca4953 .
Link to the code with 2 terms: https://github.com/deymon-d/serenedb/commit/6af2c08f695964acbd2ba772b7b9da9d4f885a69 and https://github.com/deymon-d/serenedb/commit/e5aaefb1b5e2a9aec3dde7c1938c989bacbd86fa .
Link to the code with the search tree, which is close to the version without intervals: https://github.com/deymon-d/serenedb/commit/64251b4a49f103c2b5343fe840e7932283785eb3 .

The option with 1 support element performed better than the others on tests with different probability events, and it shows himself to be worse on equally probable ones. The other options proved to be quite bad: I assume that due to the search in the distance, the probability of using a rarer element is compensated by the additional expansion of the search range. It's like in the example: there are 4 terms, the distance between each neighboring one is 1, 2 or 3, technically the distance between the first and the last one can be 9, but for this all the others must be 3 in distance, which happens quite rarely.

PR #710, opened 2026-05-25
fix: better error reporting for wrong opclass type/name

mkornaukhov:


server/catalog/index.cpp
```diff
@@ -562,6 +562,7 @@ Result ApplyHNSWOpclass(
   std::string_view owner_label, const duckdb::LogicalType& value_type,
   const std::optional<duckdb::case_insensitive_map_t<duckdb::Value>>& opts,
   InvertedIndexEntryInfo& entry) {
+  SDB_ASSERT(opts.has_value());
```
MBkkt:
```suggestion
  SDB_ASSERT(opts);
```

tests/sqllogic/sdb/pg/index/vector_search.test
```diff
@@ -821,31 +821,31 @@ DROP INDEX m_real_real;
 statement error
 CREATE INDEX m_real_bare ON t_err USING inverted(emb hnsw);
 ----
-db error: ERROR: Failed to create index: Built-in opclass 'hnsw' on column 'emb' requires options; use 'hnsw (...)'
+db error: ERROR: Failed to create index: Column 'emb' has unsupported type FLOAT[3]; ARRAY values can only be indexed with the built-in 'hnsw' opclass; use 'hnsw (...)'
```
codeworse: I guess this error message should be about the lack of options, not about type of emb field

tests/sqllogic/sdb/pg/index/vector_search.test
```diff
@@ -821,31 +821,31 @@ DROP INDEX m_real_real;
...
 ----
-db error: ERROR: Failed to create index: Unknown built-in opclass 'fake_idx' on column 'emb' (known: included, hnsw)
+db error: ERROR: Failed to create index: Unknown built-in opclass 'fake_idx' on 'emb' (known: included, hnsw)
 
 
 statement error
 CREATE INDEX m_fake_name ON t_err USING inverted(emb fake_dict);
 ----
-db error: ERROR: Failed to create index: Unknown opclass 'fake_dict' on column 'emb': not a built-in type (known: included, hnsw) and no text dictionary by that name in schema 'public'
+db error: ERROR: Failed to create index: Column 'emb' has unsupported type FLOAT[3]; ARRAY values can only be indexed with the built-in 'hnsw' opclass; use 'hnsw (...)'
```
codeworse: Incorrect error message, should be about unknown opclass

tests/sqllogic/sdb/pg/index/opclass_name_resolution.test
```diff
@@ -0,0 +1,323 @@
...
+
+
+statement ok
+CREATE TABLE shadow_hnsw_arr_expr(id INT PRIMARY KEY, x FLOAT, y FLOAT, z FLOAT);
+
+statement error
+CREATE INDEX bad_shadow_hnsw_arr_expr_idx ON shadow_hnsw_arr_expr USING inverted(
+    id, (array_value(x, y, z)) hnsw);
+----
+db error: ERROR: Failed to create index: Column 'array_value(x, y, z)' has unsupported type FLOAT[3]; ARRAY values can only be indexed with the built-in 'hnsw' opclass; use 'hnsw (...)'
```
codeworse: Maybe describe that currently we are using keyword ts dictionary and not built-in class? It's not obvious that user should add parameters, so the query starts to work correctly

tests/sqllogic/sdb/pg/index/vector_search.test
mkornaukhov: I do not agree that it is incorrect. For now we do not support arrays indexing except from HNSW. HNSW is not used here, so error message is correct. We can specify known dictionary here, but error still be the same.
But it may be not really descriptive, yes.

server/catalog/index.cpp
mkornaukhov: fixed

tests/sqllogic/sdb/pg/index/inverted_index_json_path_synthetic.test
```diff
@@ -184,17 +184,11 @@ id
...
 statement ok
 DROP TABLE jp_norm;
 
 statement ok
 DROP TEXT SEARCH DICTIONARY js_wild;
 
-statement ok
-DROP TEXT SEARCH DICTIONARY js_geo;
-
 statement ok
```
MBkkt, line 199: why?

server/catalog/index.cpp
```diff
@@ -415,14 +390,50 @@ Result ValidateInvertedIndexColumns(
         kind == duckdb::LogicalTypeId::POINTER;
       if (unsupported) {
         return {ERROR_BAD_PARAMETER,
-                "Column ",
-                c.name,
-                " has kind ",
+                "Column '",
+                label,
+                "' has kind ",
                 duckdb::EnumUtil::ToString(kind),
```
MBkkt: better to make error about type, not about kind

server/catalog/index.cpp
```diff
@@ -415,14 +390,45 @@ Result ValidateInvertedIndexColumns(
...
+    if (c.IsBuiltin(kHNSWKind)) {
+      if (kind != duckdb::LogicalTypeId::ARRAY) {
+        return {ERROR_BAD_PARAMETER, "Column '", label,
+                "' must be an ARRAY type to use the 'hnsw' opclass"};
+      }
+      if (duckdb::ArrayType::GetChildType(value_type).id() !=
+          duckdb::LogicalTypeId::FLOAT) {
+        return {ERROR_BAD_PARAMETER, "Column '", label,
+                "' must be ARRAY(FLOAT, N) to use the 'hnsw' opclass"};
+      }
```
MBkkt: I think we want to combine this error to something like `Column mas be ARRAY(FLOAT, N) ... not 'user-type'`

server/catalog/index.cpp
```diff
@@ -435,13 +441,12 @@ Result ValidateInvertedIndexColumns(
...
                            kind == duckdb::LogicalTypeId::GEOMETRY;
     if (!supported) {
       return {ERROR_BAD_PARAMETER,
-              "Column ",
-              c.name,
-              " has unsupported kind ",
+              "Column '",
+              label,
+              "' has unsupported kind ",
               duckdb::EnumUtil::ToString(kind),
```
MBkkt: still not fixed.

MBkkt (approved):

PR #734, opened 2026-05-27
Index expressions and arrays

gnusi:


server/pg/commands/create_tsdictionary.cpp
```diff
@@ -710,6 +711,13 @@ class CreateTSDictionaryOptions : public OptionsParser {
 void CreateTokenizer(ConnectionContext& conn_ctx, std::string_view name,
                      std::string_view schema, bool if_not_exists,
                      const duckdb::named_parameter_map_t& options) {
+  if (catalog::IsBuiltinOpclassName(name)) {
+    THROW_SQL_ERROR(
+      ERR_CODE(ERRCODE_RESERVED_NAME),
+      ERR_MSG("text search dictionary name \"", name,
+              "\" is reserved for the built-in '", name, "' opclass"));
+  }
+
```
mkornaukhov: That looks wrong. We had a point to allow user to write smth like
```SQL
create text search dictionary hnsw {
   template = '...' 
 };
create my_table_idx using inverted (col hnsw);
```
for backward compatibility.

gnusi: Fixed

MBkkt (approved):

PR #735, opened 2026-05-27
fix: Create or replace drop cascade

pashandor789:


gnusi (approved):
I don't like what I see

PR #749, opened 2026-05-29
Remove checking active segments in the drop

codeworse:


server/catalog/drop_task.h
MBkkt: Why do we need this at all?

server/catalog/drop_task.h
```diff
@@ -82,16 +82,17 @@ class DropTask {
     return _object.expired() && AllowToDropDependencies();
   }
 
+  virtual bool AllowToDropDependencies() const noexcept { return true; }
```
MBkkt: maybe keep it `= 0`?

MBkkt (approved):

PR #773, opened 2026-06-04
Catalog subscription

k-morozov:


server/connector/duckdb_subscription.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+#include <duckdb/main/database.hpp>
+
+namespace sdb::connector {
+namespace {
+
+void CreateSubscriptionPragma(duckdb::ClientContext& context,
+                              const duckdb::FunctionParameters& params) {
+  throw duckdb::InvalidInputException(
+    "create_subscription is not implemented yet");
```
MBkkt: don't use duckdb exceptions, THROW_SQL_ERROR instead

PR #778, opened 2026-06-05
chore: Simplier `ByWildcard` and `ByRegexp` filters

codeworse:
Before shipping PR with filter optimizer it would be helpful to make `ByWildcard` and `ByRegexp` easier. So, in this PR these two filters actually storing only automaton (they don't fallback for prefix or term filters).
Since we don't have optimizers yet these filters is created via `CreateByWildcard/ByRegexp`, but later (in optimizer PR) these optimizations will be moved into optimizer.
Also, wildcard and regexp can be created via `mutable_options()`, but then they won't be optimized. Mutable options also should be dropped as a follow-up


libs/iresearch/include/iresearch/search/regexp_filter.cpp
```diff
@@ -22,124 +22,102 @@
...
+  return ExecuteRegexp(
+    buf, pattern,
+    [&](bytes_view term) -> Filter::ptr {
+      auto filter = std::make_unique<ByTerm>();
+      *filter->mutable_field() = field;
+      filter->mutable_options()->term = term;
+      filter->boost(boost);
+      return filter;
+    },
+    [&](bytes_view term) -> Filter::ptr {
```
MBkkt: prefix

libs/iresearch/include/iresearch/search/regexp_filter.hpp
```diff
@@ -31,33 +35,41 @@ struct FilterVisitor;
...
+  using ByRegexpFilterOptions::ByRegexpFilterOptions;
 
   size_t scored_terms_limit{1024};
-  RegexpSyntax syntax{RegexpSyntax::Perl};
 
-  bool operator==(const ByRegexpOptions&) const noexcept = default;
+  bool operator==(const ByRegexpOptions& rhs) const noexcept {
+    return filter_options::operator==(rhs) &&
+           scored_terms_limit == rhs.scored_terms_limit;
+  }
```
MBkkt: why non default?

MBkkt (approved):

PR #784, opened 2026-06-08
fix: enable LOAD only on unittests

codeworse:


MBkkt (commented):
something goes wrong

MBkkt:
Fixed differently in mine recent PR about enable all sqllogic duckdb tests.

The idea for fix was following: if load/install/etc written for extension that we statically linked, it's just noop instead of throw

PR #786, opened 2026-06-10
feat: Rework Top-K follow-ups

pashandor789:
The two lines (doc_collector.hpp:75 and :121, both ExecuteTopK*):

  // old (bug):
  std::sort(hits.data(), hits.data() + hits.size(), ...);            // whole 2k buffer
  // new (fix):
  std::sort(hits.data(), hits.data() + collector.AcceptedCount(), ...); // live entries only

  Why the buffer has dead entries (iterators.hpp:213-225, Push):

  *_hits_it = {score, doc, ...};   // write at cursor
  ++_hits_it;
  if (_hits_it == _hits_end) {     // buffer full →
    _hits_it = _hits_pivot;        //   cursor jumps back to slot k
    std::nth_element(begin, pivot, end);  // best k to front, losers stay behind
    *_score_threshold = pivot->score;
  }

  After nth_element, the losers physically remain in slots [k, 2k) — not erased, just awaiting overwrite. AcceptedCount()
  (iterators.hpp:130) = _hits_it - _hits_begin = how many slots are real.

  Example, k=2, buffer 4:

  push A9,B5,C5,D5 (threshold still -inf, all enter):  [A9][B5][C5][D5]  full
  nth_element keeps 2 of them:                         [A9][C5] | [B5][D5]   threshold=5
                                                        cursor ↑    dead, tied at 5
  stream ends, nothing overwrote B,D.

  old: sort all 4   → A9, then coin-flip among C5/B5/D5 → top-2 can be {A9, B5}  ← dead doc returns
  new: sort first 2 → top-2 is {A9, C5}, always

  nth_element had to reject two of the three score-5 docs arbitrarily — so B5/D5 are ties of C5, and the old whole-buffer sort could
  resurrect one of them into the answer. Same score, different doc → different id list → the 44 changed hashes. New code never looks past
  the cursor, so the answer depends only on what the collector accepted.


server/connector/optimizer/iresearch_plan.cpp
```diff
@@ -551,43 +511,27 @@ duckdb::unique_ptr<duckdb::Expression> PushdownOffsetsCall(
...
-      if (phys < found->bind_data->column_ids.size()) {
-        target_col_id = found->bind_data->column_ids[phys];
-      }
-    }
+  std::string_view col_name{col_ref.GetAlias()};
+  if (const auto& cids = found->get->GetColumnIds();
+      resolved.column_index.GetIndex() < cids.size()) {
+    col_name =
+      found->get->GetColumnName(cids[resolved.column_index.GetIndex()]);
   }
```
MBkkt: this should be lazy? it's used only in THROW?

MBkkt (approved):

PR #788, opened 2026-06-11
refactor: Make duckdb read/write stream base classes for iresearch data input/output

MBkkt:
This will simplify code and decrease count of virtual calls, also we don't need whole footer in single array (for cases when it's not mmap)


libs/iresearch/include/iresearch/index/directory_reader_impl.cpp
MBkkt, line 1: this change is bugfix, was revealed by flaky test

MBkkt (commented):
I a little dislike that ReadData doesn't return count of read bytes, maybe it means we need to change it to throw if it's unable to read requested count of bytes

mkornaukhov (approved):
🕕 🕖

PR #793, opened 2026-06-11
[WIP] perf: tune some stuff in column store

mkornaukhov:
- Change codec picking algorithm (as for native)
- Implement projection pushdown for struct fields
- Get rid of extra copies for `mmap`-ed segment


mkornaukhov:
```
query             cs cold  native cold   cold x |       cs hot   native hot    hot x
------------ ------------ ------------ -------- | ------------ ------------ --------
structF          277.8 ms     321.9 ms    1.16x |      39.1 ms      38.6 ms    0.99x
variantF         297.1 ms     304.5 ms    1.02x |      40.6 ms      39.6 ms    0.98x
```
for now

scripts/perf/profile_types.sh
```diff
@@ -129,20 +146,28 @@ profile_one() {
...
-	# `perf report` minute-long, and the binary has no frame pointers
-	# so `--call-graph fp` would be useless.  Flat profile is enough
-	# to identify hot symbols.
+	# Callgraph via frame pointers: build_perf is RelWithDebInfo which
+	# compiles with -fno-omit-frame-pointer, so fp unwinding works and
+	# stays cheap (dwarf would be 100+MB per profile).
 	run_loop "${table}" "${expr}"
 	sleep 1
-	perf record -F 999 --pid="${pid}" --inherit \
+	perf record -F 999 --call-graph fp --pid="${pid}" --inherit \
```
MBkkt: call-graph fp is default

server/connector/duckdb_scan_base.hpp
```diff
@@ -402,6 +402,14 @@ void RunStreamingScan(duckdb::ClientContext& ctx, Gstate& g, Lstate& l,
     if (added != 0) {
       continue;
     }
+    // Segment exhausted. Emit the short chunk instead of refilling from the
+    // next segment: one straddling chunk would shift every later scan in
+    // that segment off the vector alignment that the SCAN_ENTIRE_VECTOR /
+    // direct-decode codec paths require (duckdb likewise cuts chunks at
+    // row-group ends).
+    if (collected != 0) {
+      break;
```
MBkkt: should be smarter

libs/iresearch/include/iresearch/formats/column/column_reader.hpp
```diff
@@ -347,9 +389,25 @@ class ColumnReader final {
...
+  // per-query staging. Immutable after construction, so opens are lock-free.
+  std::vector<duckdb::shared_ptr<duckdb::BlockHandle>> _data_blocks;
+  std::vector<duckdb::shared_ptr<duckdb::BlockHandle>> _validity_blocks;
+  // One-shot MADV_WILLNEED flags, one per prebuilt block. The first open of
+  // a block (cold page cache) wants kernel readahead; afterwards the pages
+  // are resident and the syscall is pure overhead (mmap_lock + range walk,
+  // ~5% of a hot query when issued per open). Racing opens may both advise;
+  // harmless. Sized with the block vectors at construction.
+  mutable std::unique_ptr<std::atomic<bool>[]> _data_advised;
+  mutable std::unique_ptr<std::atomic<bool>[]> _validity_advised;
```
MBkkt: meh

libs/iresearch/include/iresearch/formats/column/column_reader.cpp
```diff
@@ -281,6 +384,24 @@ RgWindow ColumnReader::LocateVariantRg(uint64_t row,
...
+
+bool CodecPreservesValidityOnEntireScan(duckdb::CompressionType t) noexcept {
+  switch (t) {
+    case duckdb::CompressionType::COMPRESSION_UNCOMPRESSED:
+    case duckdb::CompressionType::COMPRESSION_ALP:
+    case duckdb::CompressionType::COMPRESSION_ALPRD:
+    case duckdb::CompressionType::COMPRESSION_BITPACKING:
+    case duckdb::CompressionType::COMPRESSION_CHIMP:
+    case duckdb::CompressionType::COMPRESSION_PATAS:
+      return true;
```
MBkkt, line 391: hmm, a little strange, I think something wrong here, although it's good catch.

I think duckdb has two options
1. compression methods that can hold nulls with values (dict_fsst -- great example), in such case there's no separate validity, validity will be written with data
2. otherwise validity will be written as separate row group column

I think we want entire scan independently on validity (same as duckdb itself?)

server/connector/duckdb_scan_base.hpp
mkornaukhov: Give an example. This variant gives 1-2% improvement

libs/iresearch/include/iresearch/formats/column/column_reader.hpp
mkornaukhov: ?

scripts/perf/profile_types.sh
```diff
@@ -73,10 +88,12 @@ setup_session() {
 	psql "${PSQL_CONN}" -v ON_ERROR_STOP=1 <<EOF
 ATTACH IF NOT EXISTS '${NATIVE_DB}' AS native_db (TYPE duckdb, STORAGE_VERSION latest);
 SET search_path TO public, native_db.main;
-CREATE INDEX IF NOT EXISTS bench_idx ON bench_view USING inverted()
+CREATE INDEX IF NOT EXISTS bench_idx ON bench_view USING inverted(bool_col)
```
MBkkt, line 91:
```suggestion
CREATE INDEX IF NOT EXISTS bench_idx ON bench_view USING inverted()
```

mkornaukhov:
Implemented in https://github.com/serenedb/serenedb/pull/1210

PR #804, opened 2026-06-12
feat: Provide column statistics for inverted index scans

pashandor789:
Expose per-column statistics from the columnstore to the DuckDB optimizer via `statistics_extended` (the legacy `statistics` callback makes DuckDB disable extract pushdown, which would regress shredded VARIANT queries ~100x — see `RemoveUnusedColumns::CheckPushdownExtract`).

## How it works

- `ColumnReader` folds its per-rowgroup `DataPointer` stats into one cached `BaseStatistics` per column at segment open (recursive child stats for STRUCT/LIST/ARRAY, `CreateUnknown` for VARIANT, null flags from the validity stream).
- `DirectoryReaderImpl` merges the per-segment cached stats across segments at reader open, in the same pass that sums `docs_count`/`live_docs_count`; `DirectoryReader::GetColumnStats(field_id)` is a lock-free lookup, mirroring `live_docs_count` (computed once per reader open, reused across queries).
- Optional `approx_distinct` INCLUDE option (`CREATE INDEX ... INCLUDE (v included (approx_distinct = true))`): a 64-byte HyperLogLog per column per segment, hashed at rowgroup flush, rebuilt on consolidation, persisted as one optional `.col` footer property, unioned across segments, clamped to `live_docs_count`, exposed as `approx_unique` for join ordering.

## Effect

- `SUM` → `sum_no_overflow` rewrite fires for integer INCLUDE scans: 23–27 ms → 6–7 ms on the 25M-row types bench (parity with native duckdb tables).
- `IS NULL` constant folding and join-order cardinalities get real min/max/null/distinct inputs.
- Plan time for a 9-column aggregate is ~1 ms (stats are a hash lookup + one `Copy` per column at bind).

## Tests

- `inverted_index_statistics.test` — `sum_no_overflow` positive/negative controls, exact `stats()` output incl. has_null from validity, multi-segment merge.
- `inverted_index_approx_distinct_option.test` — `approx_unique` ranges, option parsing, consolidation rebuild.
- `inverted_index_summarize.test` — `SUMMARIZE` over index entries incl. VARIANT/STRUCT/LIST INCLUDE columns.
- 103 column/shredding/index gtests green; full index sqllogic suite green (except the pre-existing `inverted_index_matrix.test` environment failure).


pashandor789:
It's not neuro slop

MBkkt (commented):
merge writer can reuse existing hll and merge it into new one, in case when no removal was for this (old) segment?

MBkkt (approved):

PR #805, opened 2026-06-13
fix: Remove code [1/N]

MBkkt:
## Summary

Retire RocksDB as the data plane: all table data and catalog now live in a single
embedded DuckDB store, with copy-on-write catalog snapshots and iresearch-backed
inverted/vector indexes (WAL-offset durable cursor + bounded replay on recovery).

Also in this PR:
- Removed the universal `*Shard` layer — the iresearch runtime now lives as a
  `mutable shared_ptr<InvertedIndexStorage>` inside the immutable `InvertedIndex`
  catalog object (reached via `GetData()`); no `*Shard` class, no shard registry.
- Removed the dead `TableStats` / `num_rows` machinery (`pg_class.reltuples` is
  live-counted).
- Simplified the `wal_cursor` segment-meta format (single current layout) and ran a
  `commit`-terminology rename audit (`commit` kept only on `TruncateCommit` /
  `RefreshCommit`).
- ALTER TABLE ADD / DROP / ALTER COLUMN go through the DuckDB store table.
- The `third_party/duckdb` fork carries the serenedb patches for delegated
  storage/catalog and per-WAL-offset index replay.

## Follow-ups (next PR — explicitly out of scope here)

These are intentionally deferred to keep this PR focused:

1. **Perf: fix regressions** — make the cases that got slower fast again (`nextval`
   with cache=1, write-commit throughput, lookup). This PR only *remeasures* vs the
   baseline (`aa1595ddb`).
2. **Split the single DB into two** — separate catalog from data.
3. **Online index build under concurrent DML** — make CREATE INDEX backfill correct
   when writers race the build. (Crash-recovery of a half-built index is already
   handled here via the WAL-offset cursor + `BoundIndex::FinishReplay`.)
4. **Recovery speedup** — downstream of (2): a catalog-only DB recovers
   near-instantly and the data WAL becomes pure data, which tightens index replay.
   Depends on (2).
5. **Nested-type PK/UNIQUE via encoded-key index (`sdb_encode_key`)** — such columns
   are currently rejected cleanly; this lifts that restriction.
6. **DROP CONSTRAINT for UNIQUE / PRIMARY KEY / FOREIGN KEY** — PostgreSQL supports
   `ALTER TABLE ... DROP CONSTRAINT <name>` for every constraint kind; serenedb
   implements CHECK only. UNIQUE/PK/FK are stored anonymously in the catalog (no
   name or id), so this needs named constraints — an on-disk catalog format change —
   plus per-kind drop and store-table mirroring.


tests/sqllogic/sdb/pg/simple/variant.test
```diff
@@ -226,11 +232,12 @@ EXPLAIN SELECT id FROM variant_pk WHERE id = 1;
...
 QUERY PLAN
 ┌───────────────────────────┐
-│  ROCKSDB_PK_POINTS_LOOKUP │
+│          SEQ_SCAN         │
 │    ────────────────────   │
 │     Table: variant_pk     │
-│       Filter: (id=1)      │
+│   Type: Sequential Scan   │
 │      Projections: id      │
+│     Filters: (id = 1)     │
```
MBkkt, line 240: it's not index scan because small table size?

tests/server/connector/duckdb_column_serializer_test.cpp
MBkkt, line 1: maybe we want some part of this test to test our complex-pk materialization?

tests/server/connector/duckdb_key_builder_test.cpp
MBkkt, line 1: maybe we want some part of this test to test our complex-pk materialization?

server/catalog/identifiers/CMakeLists.txt
```diff
@@ -1 +1 @@
-target_sources(sereneserver PRIVATE shard_id.cpp)
+# Identifier types are header-only; no sources to compile.
```
gnusi: delete?

gnusi (approved):

PR #846, opened 2026-06-24
feat: New vector search 

codeworse:
# Native IVF vector index (replaces HNSW)

Branch: `codeworse/new-vector-search`

## What it does

Replaces the HNSW vector index with a native **IVF** (inverted-file) index built
directly on the iresearch columnstore. Vector similarity search is expressed as
ordinary SQL — an `ORDER BY <distance> LIMIT k` kNN, or a `WHERE <distance> <
threshold` range — and the planner routes it through the IVF index.

Indexes are declared inline on the column via the `ivf` opclass inside an
`inverted` index:

```sql
CREATE INDEX idx ON t
  USING inverted(pk, emb ivf (metric = 'l2', quant = 'sq8', nlist = 256));
```

- **Metrics** (`metric`): `l2`, `ip`, `cosine`, matched by the distance
  operators `<->`, `<#>`, `<=>`.
- **Quantization** (`quant`): `none` (raw f32), `sq8`, `sq4`, `pq`
  (`pq_m` subquantizers), `rabitq` (`rabitq_bits`). Quantized codes are stored
  per cluster; `pq`/`rabitq` are restricted to `l2`/`ip`.
- **Clustering** (`nlist`): number of coarse cluster lists; defaults to
  `2 * sqrt(n)` when unset.

Query-time knobs are session settings:

- `sdb_nprobe` — number of cluster lists scanned per query (recall vs. latency).
- `sdb_rerank_factor` — for quantized indexes, size the exact-distance rerank
  pool at `factor * k` (0 disables; `quant = 'none'` never reranks). Exposed as
  `sdb_rerank`.

Compared to HNSW: builds ~60× faster, and recall is tunable at query time via
`nprobe` (and rerank for quantized codes) rather than baked into the graph.

## How it is implemented

Everything lives under
[libs/iresearch/include/iresearch/formats/ivf/](../libs/iresearch/include/iresearch/formats/ivf/).
The IVF postings reuse the existing terms/postings machinery: each **cluster is a
term** (a 4-byte big-endian centroid id, `EncodeCentroidTerm`), its member docs
are the postings list, and quantized codes ride along in the postings `.pay`
stream per cluster.

**Build (`ivf_writer`).** At flush, `IvfBuilder::Build` scans the vector column,
trains centroids, assigns each doc to its nearest cluster, and emits one term per
cluster. `ResolveCentroidShape(valid_count, info)` picks a centroid layout by row
count:

- `BruteForce` — tiny inputs: a single "cluster", no real clustering.
- `Flat` — single flat centroid list (currently disabled via `kFlatMaxRows = 0`,
  branch and footer encoding retained so it can be re-enabled).
- `TwoLayer` — an L1 coarse list held in RAM plus lazily-read L2 fine centroids
  on disk (`TwoLayerCentroids`); L2 is trained on a reservoir sample, not the
  full dataset.

**Clustering (`clustering`).** Centroid training is metric-aware: `l2` uses
L2-train + L1-assign, `ip`/`cosine` use spherical (angular) k-means, with faiss's
default of 25 iterations.

**Quantizers (`quantizer`).** `MakeQuantizerWriter` / `MakeQuantizerCodebook` /
`MakeQuantizerReader` form the write/read split. The writer trains once, then
`EncodeCluster` streams codes into `.pay` per cluster; `pq` is backed by
`faiss::ProductQuantizer`. On read, a shared `QuantizerCodebook` is parsed once
per query and reused across clusters, and a `QuantizerReader` computes distances
over a code block.

**Centroids on disk (`centroids`).** `TwoLayerCentroids` serializes as a footer
(metric, shape kind, `d`, `n_l1`, L1 centroids, per-L1 body offsets, quantizer
stats). `SearchGlobal` probes L1, then reads only the selected L2 bodies from
disk to produce the `nprobe` clusters to scan.

**Query (`ivf_reader` + connector).** `IvfVectorReader` reads raw doc vectors
(single or batched) for exact reranking. The connector materializes the query
vector, scans the chosen clusters via the cluster term reader, scores by
quantized distance, and — for quantized indexes — reranks the top
`rerank_factor * k` with exact distances. A fast block-scan path is used when
there is no deletion mask; otherwise it falls back to the masked similarity
iterator.


libs/iresearch/include/iresearch/formats/ivf/centroids.hpp
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+ private:
+  const float* L1Centroid(uint32_t i) const noexcept {
+    return _l1_centroids.data() + static_cast<size_t>(i) * _d;
+  }
+
+  VectorMetric _metric = VectorMetric::L2Sqr;
+  uint32_t _d = 0;
+  uint32_t _n_l1 = 0;
```
MBkkt, line 110: segment size => sqrt(n) * some_coeff => count of leaf clusters => 


```
2 level: 1 scan root, 2 scan leaf, 3 scan posting
1 level: 1 scan root, 2 scan posting
0 level: 1 scan posting
```

MBkkt (approved):

PR #852, opened 2026-06-25
fix: copy non-inline strings out of columnstore scan to avoid UAF

pashandor789:
A multi-row columnstore scan that crosses index segments could emit non-inline string_t that are zero-copy views into a per-segment scan state's decompression buffer. RangeScan::Scan replaces the cursor when it crosses to the next segment, destroying that scan state while the output Vector still references the freed buffer -> heap-use-after-free. It surfaces as corrupted bytes in a later join compare or pg-wire serialization of a multi-segment scan (ASAN: freed StringHeap).

Codec: this affects VARCHAR/BLOB columns stored with dict_fsst. Its dictionary scan decodes into a reusable dictionary owned by the segment scan state, so the flat output only holds views into it. Other string codecs are unaffected -- dictionary and uncompressed are block-backed (already healed by the block pin), and plain FSST decodes into the output's own string heap. dict_fsst is auto-selected for compressible text (e.g. source code), which is why it reproduces with real data but not with synthetic/random strings.

Temporary fix: copy each non-inline string into the output Vector's own heap so it owns its bytes. The intended follow-up keeps the borrowed buffer alive without copying (hand the outgoing scan state to the output on a segment crossing).


gnusi (approved):

libs/iresearch/include/iresearch/formats/column/column_reader.hpp
```diff
@@ -194,7 +196,26 @@ class ColumnReader final {
...
-      if (_pin) {
+
+      // TEMP FIX: copy each non-inline string into the output's own heap so it
+      // owns its bytes (inline <=12B strings are self-contained). This adds a
+      // memcpy per long string on every scan. The intended fix is to keep the
+      // borrowed heap alive without copying -- hand the outgoing scan state to
+      // the output on a segment crossing -- which is to be wired in later.
+      // Fixed-width / non-flat output cannot borrow a heap, only the block, so
+      // it just needs the pin.
+      if (out_vec.GetType().InternalType() == duckdb::PhysicalType::VARCHAR &&
```
gnusi: Condition looks strange

gnusi (changes requested):

gnusi (approved):

PR #853, opened 2026-06-25
fix: make attached dbs unique

Dronplane:
Some tests used instance global aliases for attaching databases. That broke parallel run on CI if unlucky enough as one test file may detach database used by another.

also workaround for https://github.com/serenedb/serenedb/issues/857


gnusi (approved):

MBkkt (commented):
Use `__RUN_ID__` instead of custom names/database-id

Dronplane:
> Use `__RUN_ID__` instead of custom names/database-id

ok

PR #866, opened 2026-06-29
feat: alphanumeric tokenizing function

mkornaukhov:
It's widely used and much faster than regexp and segmentation tokenizers.
0 -- with lowercasing, 1 -- without:
```
-------------------------------------------------------------------------------------------
Benchmark                                 Time             CPU   Iterations UserCounters...
-------------------------------------------------------------------------------------------
MixedCorpus/BmFunction/0             483872 ns       483661 ns         1423 bytes_per_second=2.01911Gi/s
MixedCorpus/BmFunction/1            1652206 ns      1652116 ns          420 bytes_per_second=605.284Mi/s
MixedCorpus/BmPattern/0             5731875 ns      5730241 ns          123 bytes_per_second=174.513Mi/s
MixedCorpus/BmSegmentation/1       15569731 ns     15557739 ns           45 bytes_per_second=64.2767Mi/s
LongTokenCorpus/BmFunction/0         108918 ns       108883 ns         6430 bytes_per_second=8.96895Gi/s
LongTokenCorpus/BmFunction/1         135755 ns       135655 ns         5163 bytes_per_second=7.19888Gi/s
LongTokenCorpus/BmPattern/0         1504239 ns      1503640 ns          463 bytes_per_second=665.053Mi/s
LongTokenCorpus/BmSegmentation/1   12386485 ns     12386080 ns           56 bytes_per_second=80.7358Mi/s
BmSmallInput/8                         4.44 ns         4.43 ns    199430774 bytes_per_second=1.68043Gi/s
BmSmallInput/16                        7.38 ns         7.38 ns     97183808 bytes_per_second=2.01856Gi/s
BmSmallInput/24                        9.80 ns         9.80 ns     66101159 bytes_per_second=2.28171Gi/s
BmSmallInput/31                        14.0 ns         14.0 ns     62621162 bytes_per_second=2.06517Gi/s
BmSmallInput/48                        12.5 ns         12.5 ns     73554412 bytes_per_second=3.56637Gi/s
BmSmallInput/64                        14.1 ns         14.1 ns     59238257 bytes_per_second=4.23711Gi/s
```


server/connector/functions/split_by_non_alpha.cpp
```diff
@@ -0,0 +1,134 @@
...
+    tl_data = duckdb::UnifiedVectorFormat::GetData<bool>(tl_format);
+  }
+
+  result.SetVectorType(duckdb::VectorType::FLAT_VECTOR);
+  duckdb::ListVector::SetListSize(result, 0);
+  auto* list_entries =
+    duckdb::FlatVector::GetDataMutable<duckdb::list_entry_t>(result);
+  auto& result_validity = duckdb::FlatVector::ValidityMutable(result);
+  ListSink sink{result};
+  std::string buf;
```
gnusi: It seems we can avoid copying data into intermediate `buf`. ToLower can be done directly on the bytes in chunk.

server/connector/functions/split_by_non_alpha.cpp
```diff
@@ -0,0 +1,134 @@
...
+  for (duckdb::idx_t i = 0; i < count; i++) {
+    auto text_idx = text_format.sel->get_index(i);
+    if (!text_format.validity.RowIsValid(text_idx)) {
+      result_validity.SetInvalid(i);
+      list_entries[i] = {sink.Offset(), 0};
+      continue;
+    }
+    bool to_lower = false;
+    if (has_to_lower) {
+      auto tl_idx = tl_format.sel->get_index(i);
```
gnusi: We could resolve to_lower at bind time in most cases

libs/iresearch/include/iresearch/analysis/split_by_non_alpha_tokenizer.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+bool SplitByNonAlphaTokenizer::next() {
+  const char* const base = _data.data();
+  const size_t size = _data.size();
+
+  // Skip separators: any byte that is not an ASCII letter or digit. Non-ASCII
+  // bytes (>= 0x80) are not alphanumeric, so they act as separators too, which
+  // matches a `[^A-Za-z0-9]+` split.
+  while (_pos < size &&
+         !absl::ascii_isalnum(static_cast<unsigned char>(base[_pos]))) {
```
gnusi: Gut says it such loops can be vectorized (relevant for other places as well)

server/connector/functions/split_by_non_alpha.cpp
mkornaukhov: fixed

server/connector/functions/split_by_non_alpha.cpp
mkornaukhov: fixed

libs/iresearch/include/iresearch/analysis/split_by_non_alpha_tokenizer.cpp
mkornaukhov: fixed

libs/iresearch/include/iresearch/analysis/split_by_non_alpha.hpp
```diff
@@ -0,0 +1,157 @@
...
+  };
+  Mask bitmask = 0;
+  for (unsigned i = 0; i < std::numeric_limits<Mask>::digits; ++i) {
+    const unsigned char byte = static_cast<unsigned char>(block[i]);
+    const unsigned char lowered = byte | 0x20;
+    const uint32_t is_digit = static_cast<unsigned char>(byte - '0') < 10;
+    const uint32_t is_letter = static_cast<unsigned char>(lowered - 'a') < 26;
+    bitmask = as_mask(bitmask | as_mask((is_digit | is_letter) << i));
+  }
+  return bitmask;
```
gnusi: `as_mask` here is only guarding against integer promotion, but in this function every use is redundant — assigning back into `Mask` truncates to the block width for you. Drop the lambda entirely:

```suggestion
  Mask bitmask = 0;
  for (int i = 0; i < std::numeric_limits<Mask>::digits; ++i) {
    const unsigned char byte = static_cast<unsigned char>(block[i]);
    const unsigned char lowered = byte | 0x20;
    const unsigned is_digit = static_cast<unsigned char>(byte - '0') < 10;
    const unsigned is_letter = static_cast<unsigned char>(lowered - 'a') < 26;
    bitmask |= static_cast<Mask>((is_digit | is_letter) << i);
  }
  return bitmask;
```

libs/iresearch/include/iresearch/analysis/split_by_non_alpha.hpp
```diff
@@ -0,0 +1,157 @@
...
+      const auto run = as_mask(bitmask >> token_begin);
+      const auto token_length = as_mask(std::countr_zero(as_mask(~run)));
+      if (token_begin + token_length >= kBits) {
+        OpenToken(offset + token_begin);
+        return;
+      }
+      EmitToken(offset + token_begin, token_length);
+      bitmask = as_mask(
+        bitmask & ~as_mask(LowBitsMask<Mask>(token_length) << token_begin));
+    }
```
gnusi, line 77: Same here — most `as_mask` calls aren't load-bearing. Only three casts actually matter, and they're all the same shape: a unary `~` whose result feeds `countr_zero` or is stored as a mask (otherwise `countr_zero` walks a promoted 32-bit `int` instead of the block width). Everywhere else the value is either stored straight back into a `Mask` (assignment truncates) or is a bit index that should just be `int`. This drops the lambda and the `// Just not to bloat` note:

```suggestion
    constexpr int kBits = std::numeric_limits<Mask>::digits;

    if (_token_open) {
      const Mask separators = ~bitmask;
      if (separators == 0) {
        return;
      }
      const int token_end = std::countr_zero(separators);
      CloseToken(offset + token_end);
      bitmask &= ~LowBitsMask<Mask>(token_end);
    }

    while (bitmask != 0) {
      const int token_begin = std::countr_zero(bitmask);
      const Mask run = bitmask >> token_begin;
      const int token_length = std::countr_zero(static_cast<Mask>(~run));
      if (token_begin + token_length >= kBits) {
        OpenToken(offset + token_begin);
        return;
      }
      EmitToken(offset + token_begin, token_length);
      bitmask &= ~static_cast<Mask>(LowBitsMask<Mask>(token_length) << token_begin);
    }
```

gnusi (commented):
A couple of cleanups around `as_mask` — the tokenizer logic itself looks correct (traced the block/boundary and width cases), these just cut the integer-promotion-guard noise.

MBkkt (commented):
this isn't tokenizer, it's function, we maybe want to make tokenizers functions

But maybe as separate PR

server/connector/functions/split_by_non_alpha.cpp
```diff
@@ -0,0 +1,172 @@
...
+    std::move(args), duckdb::LogicalType::LIST(duckdb::LogicalType::VARCHAR),
+    fn, bind};
+  f.SetNullHandling(duckdb::FunctionNullHandling::SPECIAL_HANDLING);
+  return f;
+}
+
+}  // namespace
+
+void RegisterSplitByNonAlpha(duckdb::ExtensionLoader& loader) {
+  duckdb::ScalarFunctionSet set{"split_by_non_alpha"};
```
MBkkt: I think all such functions should have some shitty prefix.
Otherwise they can conflict with other systems functions

Also it's quite convenient if all such functions (tokenizers, analyzers, etc) will be prefixed same way.

I also think we need to think about more generic solution, e.g. 
first argument input, second argument is tokenizer config (json)?

also constant argument can be handled in bind.
dynamic argument can be quite slow.

also it probably make sense to make such function with generic name?
`ts_tokenize(input: varchar, config: json/tokenizer-name)`?

mkornaukhov: > also constant argument can be handled in bind.

It's handled now

> I think all such functions should have some shitty prefix

👍 

> I also think we need to think about more generic solution, e.g.
first argument input, second argument is tokenizer config (json)?

Sounds good, but out of scope of this PR

> ts_tokenize(input: varchar, config: json/tokenizer-name)

It does not look very convenient.
```
ts_tokenize(input, '{name: "split_by_non_alpha", params: {to_lower: true}}' 
VS
ts_split_by_non_alpha(input, true)
```

MBkkt:
> It does not look very convenient.

it shouldn't, convenient is create tokenizer and use it

gnusi (approved):

PR #867, opened 2026-07-01
feat: Search backed tables part 3

Dronplane:
Add background COMMIT/COMPACTION/CLEANUP threads to search tables. Execution generalizes and shares machinery with the inverted index. Same options control background cadence both for inverted index and search table (with possible override for each entity) 
 
Follow-ups from previous PR:
Search table path now includes schema_id.
DropTask for search table is separated, is_root handling fixed.  

+ Fix shutdown order assertion


server/search/search_table.cpp
```diff
@@ -21,26 +21,49 @@
...
-std::filesystem::path SearchTable::GetPath(ObjectId db_id, ObjectId table_id) {
+std::filesystem::path SearchTable::GetPath(ObjectId db_id, ObjectId schema_id,
+                                           ObjectId table_id) {
   SDB_ASSERT(db_id.isSet());
+  SDB_ASSERT(schema_id.isSet());
   SDB_ASSERT(table_id.isSet());
   auto path = GetSearchEngine().GetPersistedPath(db_id);
+  path /= absl::StrCat(schema_id);
   path /= absl::StrCat(table_id);
   return path;
```
MBkkt: we can reuse code for index (it already has case when index_id is 0)

server/search/search_table.cpp
```diff
@@ -21,26 +21,49 @@
...
+ABSL_FLAG(uint32_t, server_search_table_refresh_interval_ms, 1000,
+          "Interval between background commits (RefreshCommit) for search "
+          "tables, making inserts searchable. 0 disables background commit "
+          "(VACUUM is then the only way to flush); set both this and the "
+          "compaction interval to 0 to turn background maintenance off.");
+ABSL_FLAG(uint32_t, server_search_table_compaction_interval_ms, 10000,
+          "Interval between background segment consolidations for search "
+          "tables. 0 disables background consolidation.");
+ABSL_FLAG(uint32_t, server_search_table_cleanup_interval_step, 10,
+          "Run iresearch directory GC every Nth background commit for search "
```
MBkkt: maybe we want to reuse current flags? just make them more generic, I don't think we want more flags

server/search/search_table.cpp
Dronplane: done

server/search/search_table.cpp
Dronplane: Actually we have session options (with defaults) for inverted indexes that have  already generic enough naming so I'd switch to this existing pattern

server/catalog/store/store.cpp
```diff
@@ -1032,9 +1032,7 @@ Result CatalogStore::ExecuteCreateStoreTableImpl(const StoreTableDef& def,
     duckdb::MetaTransaction::Get(context).ModifyDatabase(
       catalog.GetAttached(),
       duckdb::DatabaseModificationType::CREATE_CATALOG_ENTRY);
-    SDB_IF_FAILURE("pause_store_create_table") {
-      sdb::WaitWhileFailurePointDebugging("pause_store_create_table");
-    }
+    sdb::WaitWhileFailurePointDebugging("pause_store_create_table");
```
MBkkt: it's bad because it's exists when failure points disabled, better to be macros

Dronplane: when failpoints are disabled it is just 
```
constexpr void WaitWhileFailurePointDebugging(std::string_view) noexcept {}
```
So I expect it to be just discarded by compiler

MBkkt: why can't it be macro instead?

Dronplane: I guess as function it looks more readable + it helps us to keep bookkeeping structs private in the debugging.cpp. Like all other failpoint related functions around.
What benefits do you see in rewriting it (or all functions) as macros?

MBkkt: we will be sure they're (function call and string literal) gone, it's also quite convenient if we have macros for fault injection, we already have one for `if`, why wait should be function?

Dronplane: I don't think we should be that cautious with compiler not discarding such function call entirely.
But I see your point. Then I can suggest to just wrap a function into the macro like

```
#ifdef SDB_FAULT_INJECTION
#define SDB_WAIT_ON_FAILURE(what) sdb::WaitWhileFailurePointDebugging(what)
#else
#define SDB_WAIT_ON_FAILURE(what)
#endif
```

If you are ok with that I'll commit this.

MBkkt: ok

MBkkt (approved):

PR #871, opened 2026-07-02
feat : Add foreign server

pashandor789:
PostgreSQL-native FDW DDL — `CREATE/DROP SERVER`, `CREATE/DROP USER MAPPING` — as a **persistent, catalog-backed** layer over the ClickHouse and postgres connectors, plus the vendored ClickHouse connector itself. A plain `ATTACH (TYPE clickhouse)` is in-memory instance state that dies with the process; a foreign server is a catalog object that survives restart via boot replay.

## ClickHouse connector (`third_party/duckdb_clickhouse`)

First-class connector modeled on `postgres_scanner`, speaking ClickHouse's **native TCP protocol** via vendored `clickhouse-cpp` (columnar blocks, LZ4/ZSTD, reusing our in-tree abseil/lz4/zstd/openssl):

- `ATTACH ... (TYPE clickhouse)` with catalog/schema discovery, plus ad-hoc `clickhouse_scan()` / `clickhouse_query()` / `clickhouse_execute()`.
- Full CH→DuckDB type map (Nullable/LowCardinality/Array/Map/Tuple/Enum/Decimal/DateTime64/UUID/IPv4...), read+write (INSERT via native blocks, UPDATE/DELETE/MERGE INTO as `ALTER TABLE ... mutations_sync=1`, PK-as-rowid with a non-unique-key guard).
- Pushdown on the **shared dbconnector layers** (serenedb/database-connector#1): projection, dialect-aware filters with a divergence corpus keeping CH-incompatible predicates local, ORDER BY/LIMIT/TopN with engine-agnostic NULLS placement.
- Shared connection pool (`ch_pool_*` settings, Ping health checks, reaper), per-query `ch_statement_timeout_millis` (`max_execution_time`), cardinality from `system.tables`, scan progress, and auto-rebind on remote schema drift (`COLUMN_NOT_FOUND` → clear caches → rebind once).
- Index integration: an inverted index over an attached CH/pg table re-fetches matched rows by the table's own primary key (`PkSpec::ExternalDBKey`, `WHERE pk IN (...)`); `WITH (on_conflict = 'throw'|'nothing')` governs duplicate lookup keys.

## FDW catalog model — PG's exact shape

```
Database ── ForeignServer "chs" ─┬─ UserMapping "public"
                                 └─ UserMapping "alice"
```

- Servers are **database children** (no schema, name unique per database — `pg_foreign_server` has no namespace column); `DROP SCHEMA` can never touch one; `DROP DATABASE` sweeps and detaches its servers.
- A mapping's parent is **its server** and its name is **the mapped role** — `(parent, name)` is PG's `(umserver, umuser)`, no synthetic name encoding.
- Store rows stay flat under the database id: one range-delete sweeps them, boot loads per database.

## Attach mechanics

- Credentials never enter SQL text: attach goes through a **transient DuckDB secret** (`ATTACH '' AS "chs" (TYPE ..., SECRET __sdb_fdw_secret_chs)`), registered for the handshake and dropped immediately. No connstr escaping, nothing to redact in errors.
- Live credentials = server OPTIONS ⊕ PUBLIC mapping (mapping wins per key, keys canonicalised per dialect). Changing the PUBLIC mapping **probes on a throwaway alias first**, then atomically swaps — a bad mapping never detaches a working server.
- **Boot replay** re-attaches every server after catalog load; a down remote warns and boot continues.
- `DROP SERVER` / `DROP DATABASE` detach the live attachment (a leaked attachment stayed readable with no ACLs and blocked name reuse).

## RBAC + exposure

- `CREATE SERVER` requires **CREATE on the database** (PG gates on FDW USAGE, which we don't model); drops require ownership; mapping authority follows the server owner; `DROP ROLE` is refused while a mapping references or is owned by the role; the create gate fires **before** the remote is contacted.
- `pg_user_mapping` is superuser-only with **cleartext** `umoptions` (PG's model: the read gate is the guard); world-readable `pg_foreign_server` **redacts secret values** (unlike PG, our server options may carry credentials); the `pg_user_mappings` view's per-viewer CASE serves everyone else. Passwords are write-only through SQL for non-superusers.

## Testing

sqllogic throughout (per team convention): the full `sdb/clickhouse` suite (types, pushdown incl. EXPLAIN plan shapes, writes, MERGE INTO, errors, pool, FDW permissions/scenarios/creds/cascade-detach), `create_server_pgscan` + `inverted_index_pgscan` for the postgres side, crash-recovery tests for owner/ACL and credential round-trip through boot replay (the fixture password carries a load-bearing space), RBAC/index/system-catalog suites for blast radius, and an attach-bridge gtest for the dialect canonicalisation sqllogic can't observe (trust auth ignores credentials).

## Known limitations / follow-ups

- **Per-user runtime credentials (“2b”)**: per-user mappings are persisted and visible but the live attachment uses one resolved identity — ATTACH is instance-global.
- The attachment **alias namespace is instance-global**, so equally-named servers in two databases can't be live simultaneously (catalog scoping itself is per-database, like PG).
- `GRANT USAGE ON FOREIGN SERVER` is not in the grammar yet (the mapped-role branch of `pg_user_mappings` is reachable via owner-role membership only).
- Boot-time attach is eager and serial; lazy attach-on-first-use via the rebind hook is a designed follow-up.

Stack: serenedb/duckdb#43 (FDW grammar, submodule `third_party/duckdb`), serenedb/database-connector#1 (shared pushdown, submodule `third_party/database-connector`), serenedb/duckdb-postgres branch `wip/fdw-clickhouse` (a two-line adaptation: the shared CreateConfigs now require an explicit dialect, so the postgres connector states `Dialect::Postgres`) — land those first, this PR pins their pointers.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


server/connector/index_source_external_db_key.h
```diff
@@ -0,0 +1,73 @@
...
+                            duckdb::idx_t count,
+                            duckdb::DataChunk& output) final;
+
+ private:
+  // Quoted "catalog"."schema"."table" and the quoted PK column name.
+  std::string _qualified_table;
+  std::string _pk_quoted;
+  // Cached "SELECT <pk>, <cols> FROM <table> WHERE <pk> IN (" -- everything but
+  // the per-call IN list, which is all that varies between Materialize() calls.
+  std::string _sql_prefix;
```
MBkkt: prepared statement + bind parameters for in?

MBkkt (changes requested):
1. For postgres we should make it `ctid` (https://www.postgresql.org/docs/current/ddl-system-columns.html) at least by default, and I think as separate effort we should allow user to specify what columns to store (these columns should be stored as struct column probably for simplicity) and then we can use this struct column to lookup from datasource with this ability user can specify either primary key/secondary key, in general just anything, and this approach should be very generic, it should work for any query I think! I suggest struct instead of separate columns for simplicity reasons (single column reader abstraction/etc)
2. For clickhouse unfortunately impossible to do something like this, their PK generally same as ours: file id + id in file (part + part offset), so it's possible to use them only if table not modified + table not compacting (merged), so I think only one thing for clickhouse that we can try, using `prewhere` instead of `where`, although I think it should be automatically
3. lets make duckdb-clickhouse repo separate repo?

cmake/Runtime.cmake
```diff
@@ -112,4 +112,12 @@ include(UpdateModule)
 
 sdb_update_module(${GIT_EXECUTABLE} "llvm-project" ${CMAKE_SOURCE_DIR}/third_party "LICENSE.TXT")
 
+# The runtimes are configured before the top-level -nostdinc++ (see the root
+# CMakeLists), so set it here too: build libc++/libc++abi/libunwind against
+# their own in-tree headers (passed via -isystem) rather than the system
+# libstdc++. Without this, clang's default search reaches GCC's
+# /usr/include/c++/NN and libc++'s own `#include_next <stdlib.h>` resolves to
+# libstdc++'s stdlib.h (`using std::abort;`), which fails to compile.
+add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-nostdinc++>)
```
MBkkt: lets remove this

libs/iresearch/include/iresearch/index/index_source.hpp
MBkkt: remove it

libs/iresearch/include/iresearch/index/pk_batch_helpers.hpp
```diff
@@ -33,9 +33,16 @@
 namespace sdb::connector {
 
 inline void AppendPrimaryKeysFromVector(PrimaryKeyBatch& pk,
```
MBkkt: remove whole function

server/catalog/persistence/index.h
```diff
@@ -36,6 +36,9 @@ enum class PkColumnKind : uint8_t {
   I64,
   I64I64,
   Unable,
```
MBkkt: maybe unable first

server/catalog/persistence/index.h
```diff
@@ -36,6 +36,9 @@ enum class PkColumnKind : uint8_t {
   I64,
   I64I64,
```
MBkkt: remove this and struct?
make `Stored,` or `AutoStored` and `UserStored`

server/catalog/foreign_server.cpp
```diff
@@ -0,0 +1,258 @@
...
+#include "basics/serializer.h"
+#include "catalog/persistence/foreign_server.h"
+
+namespace sdb::catalog {
+namespace {
+
+using persistence::ForeignServerData;
+
+// Map an FDW name to the DuckDB connector storage type; "" if unsupported.
+std::string StorageTypeForFdw(std::string_view fdw) {
```
MBkkt:
```suggestion
std::string_view StorageTypeForFdw(std::string_view fdw) {
```

server/catalog/foreign_server.cpp
```diff
@@ -0,0 +1,258 @@
...
+  if (fdw == "clickhouse_fdw" || fdw == "clickhouse") {
+    return "clickhouse";
+  }
+  if (fdw == "postgres_fdw" || fdw == "postgres") {
```
MBkkt: strange?

server/catalog/foreign_server.cpp
```diff
@@ -0,0 +1,258 @@
...
+#include "catalog/persistence/foreign_server.h"
+
+namespace sdb::catalog {
+namespace {
+
+using persistence::ForeignServerData;
+
+// Map an FDW name to the DuckDB connector storage type; "" if unsupported.
+std::string StorageTypeForFdw(std::string_view fdw) {
+  if (fdw == "clickhouse_fdw" || fdw == "clickhouse") {
```
MBkkt: strange?

server/catalog/foreign_server.cpp
```diff
@@ -0,0 +1,258 @@
...
+  if (fdw == "postgres_fdw" || fdw == "postgres") {
+    return "postgres";
+  }
+  return {};
+}
+
+// Canonicalise a (lower-cased) option key to the connector's secret parameter
+// name. The connectors resolve aliases in their connstr parser, but their
+// secret overlays read exact keys, so we resolve the aliases here instead.
+std::string CanonicalOptionKey(std::string_view storage, std::string key) {
```
MBkkt:
```suggestion
std::string_view CanonicalOptionKey(std::string_view storage, std::string_view key) {
```

server/catalog/foreign_server.cpp
```diff
@@ -0,0 +1,258 @@
...
+}
+
+// Drops a transient secret registered by PrepareForeignServerAttach; a missing
+// secret is ignored (best-effort cleanup).
+void DropForeignServerSecret(duckdb::ClientContext& context,
+                             std::string_view secret_name) {
+  auto& secret_manager = duckdb::SecretManager::Get(context);
+  // Like RegisterSecret, dropping needs an active transaction; wrap it (the
+  // ATTACH query has already returned the connection to autocommit).
+  context.RunFunctionInTransaction([&]() {
```
MBkkt:
```suggestion
  context.RunFunctionInTransaction([&] {
```

server/catalog/foreign_server.cpp
```diff
@@ -0,0 +1,258 @@
...
+}
+
+}  // namespace
+
+bool IsSupportedFdw(std::string_view fdw_name) {
+  return !StorageTypeForFdw(fdw_name).empty();
+}
+
+std::string QuoteSqlIdentifier(std::string_view name) {
+  return duckdb::KeywordHelper::WriteQuoted(std::string{name}, '"');
```
MBkkt: make in duckdb string_view

server/catalog/foreign_server.cpp
```diff
@@ -0,0 +1,258 @@
...
+    // The secret's values are internal to the attach; hide every one from
+    // duckdb_secrets() for the attach window (the connector reads secret_map
+    // directly, so redact_keys only affects the display).
+    secret->redact_keys.insert(key);
+  }
+
+  auto& secret_manager = duckdb::SecretManager::Get(context);
+  // RegisterSecret needs an active transaction; these attach paths run on a
+  // fresh connection with none, so wrap it (begins + commits one).
+  context.RunFunctionInTransaction([&]() {
```
MBkkt:
```suggestion
  context.RunFunctionInTransaction([&] {
```

pre-commit

server/catalog/foreign_server.cpp
```diff
@@ -0,0 +1,258 @@
...
+  DropForeignServerSecret(*conn.context, secret);
+  if (result->HasError()) {
+    // The attach carries credentials in a TEMPORARY secret, not the SQL text,
+    // and neither connector echoes them on connect failure (the postgres error
+    // renders an empty attach path; PQerrorMessage never prints the password),
+    // so the connector error is safe to surface verbatim.
+    return std::string{result->GetError()};
+  }
+  return "";
+}
```
MBkkt: looks bad, what if error is empty string

server/catalog/object.h
```diff
@@ -59,6 +59,8 @@ enum class ObjectType : uint8_t {
   Schema,
   // Under schema - dependencies first
   Tokenizer,
+  ForeignServer,
+  UserMapping,
   PgSqlFunction,
   PgSqlType,
   PgSqlView,
```
MBkkt, line 66:
```suggestion
Function,
Type,
```

server/catalog/pk_spec.h
```diff
@@ -32,8 +32,19 @@ enum class PkSpec : uint8_t {
...
   FileIndexPlusDuckDBRowId,
+  // Attached external DB (pg/CH): matched rows re-fetch by value via SQL.
+  //   ExternalPostgresCtid -- the postgres ctid (travels as the duckdb rowid).
+  //   ExternalColumnKey    -- 1..N real columns (engine PK metadata or
+  //   key_columns).
+  ExternalPostgresCtid,
+  ExternalColumnKey,
 };
 
+constexpr bool IsExternalPK(PkSpec spec) noexcept {
```
MBkkt, line 43: strange function

server/connector/optimizer/rbac.cpp
```diff
@@ -150,6 +151,40 @@ void CollectAndEnforce(duckdb::ClientContext& context, duckdb::Binder& binder) {
...
+  // USAGE (or ownership) on that server (PG-style); the remote enforces the
+  // rest. The attachment is instance-global, so resolve the server across ALL
+  // databases (names are globally unique), never scoped to the caller's own
+  // database -- a server created in one database must still be gated when
+  // referenced from a session in another (else the check silently skips ->
+  // cross-database bypass). Checked once per distinct catalog alias, on the
+  // caller's own role. A null resolve means the catalog is a regular database,
+  // a system/temp catalog, or a raw ATTACH foreign catalog with no CREATE
+  // SERVER object -- none are governed by foreign-server USAGE.
+  containers::FlatHashSet<std::string> checked_catalogs;
```
MBkkt, line 165:
```suggestion
  containers::FlatHashSet<std::string_view> checked_catalogs;
```

server/connector/optimizer/rbac.cpp
```diff
@@ -150,6 +151,40 @@ void CollectAndEnforce(duckdb::ClientContext& context, duckdb::Binder& binder) {
...
+  for (const auto& req : reqs) {
+    if (!req.table) {
+      continue;
+    }
+    const auto catalog_name =
```
MBkkt, line 170:
```suggestion
    const auto& catalog_name =
```

server/connector/optimizer/rbac.cpp
```diff
@@ -150,6 +151,40 @@ void CollectAndEnforce(duckdb::ClientContext& context, duckdb::Binder& binder) {
...
+      req.table->ParentCatalog().GetName().GetIdentifierName();
+    if (!checked_catalogs.insert(catalog_name).second) {
```
MBkkt, line 172:
```suggestion
    if (!checked_catalogs.emplace(catalog_name).second) {
```

server/connector/optimizer/rbac.cpp
MBkkt, line 165: why dedup?

server/connector/duckdb_index_scan_entry.cpp
```diff
@@ -139,9 +139,14 @@ duckdb::TableFunction ViewInvertedIndexScanEntry::GetScanFunction(
   data->table_entry = this;
   data->entry_kind = ScanEntryKind::InvertedIndex;
   data->inverted_index = _inverted_index;
-  if (auto fp = ResolveViewFastPath(context, *_sdb_view)) {
-    data->lookup_label = FormatLookupLabel(*fp);
-    data->lookup_supports_filters = fp->supports_filters;
+  std::span<const std::string> key_cols;
+  if (_inverted_index) {
+    key_cols = _inverted_index->GetOptions().key_columns;
+  }
```
MBkkt: what? when we don't have inverted_index?

server/connector/duckdb_physical_create_index.cpp
```diff
@@ -341,15 +350,18 @@ SereneDBPhysicalCreateIndex::GetGlobalSinkState(
       }
     }
     const bool table_backed = IsDuckDBTable();
-    enum class KeyShape { Single, Two, Synth };
+    enum class KeyShape { Single, Two, Struct, Synth };
```
MBkkt:
```suggestion
    enum class KeyShape { Single, Two, Struct };
```
maybe

server/connector/duckdb_physical_create_index.cpp
```diff
@@ -615,6 +630,46 @@ duckdb::SinkResultType SereneDBPhysicalCreateIndex::Sink(
...
+      if (gstate.postgres_ctid) {
+        // Split the duckdb rowid back into the ctid's STRUCT{page, tuple}: the
+        // page half is run-heavy (RLE) and the tuple half is a small sawtooth
+        // (bitpacked), ~2x smaller than the packed int64 as one column.
+        SDB_ASSERT(base + 1 == chunk.ColumnCount());
+        auto& rowid_vec = chunk.data[base];
+        duckdb::UnifiedVectorFormat fmt;
+        rowid_vec.ToUnifiedFormat(num_rows, fmt);
+        const auto* packed = duckdb::UnifiedVectorFormat::GetData<int64_t>(fmt);
+        pk_scratch = std::make_unique<duckdb::Vector>(pg::CTID());
```
MBkkt: pg::CTID shouldn't be int64, it's struct, so we don't need repack

server/connector/duckdb_storage_extension.cpp
```diff
@@ -104,9 +105,24 @@ void SereneDBCatalog::OnDetach(duckdb::ClientContext& context) {
...
+    }
+  }
+
   duckdb::shared_ptr<void> keep_alive = GetAttached().shared_from_this();
   catalog::DropDatabase(ax, GetName().GetIdentifierName(),
                         std::move(keep_alive));
+  for (const auto& server_name : detach_servers) {
+    catalog::DetachForeignServerAttachment(server_name);
+  }
   SDB_IF_FAILURE("crash_on_drop") { SDB_IMMEDIATE_ABORT(); }
```
MBkkt: feels like logical data race possible

server/connector/index_source_external_lookup.h
```diff
@@ -0,0 +1,85 @@
...
+namespace sdb::connector {
+
+class ExternalLookupIndexSource final : public ViewIndexSourceBase {
+ public:
+  ExternalLookupIndexSource(
+    duckdb::ClientContext& context, ViewFastPath fast_path,
+    std::span<const duckdb::idx_t> projected_columns,
+    std::span<const duckdb::LogicalType> projected_types,
+    std::span<const catalog::Column::Id> bind_column_ids);
+  ~ExternalLookupIndexSource() final = default;
```
MBkkt:
```suggestion
```

server/connector/index_source_external_lookup.h
```diff
@@ -0,0 +1,85 @@
...
+    return PrimaryKeyBatch::Kind::Struct;
+  }
+
+  duckdb::idx_t Materialize(duckdb::ClientContext& context,
+                            PrimaryKeyBatch& batch, duckdb::idx_t start,
+                            duckdb::idx_t count,
+                            duckdb::DataChunk& output) final;
+
+ private:
+  enum class Dialect : uint8_t { Postgres, ClickHouse };
```
MBkkt:
```suggestion
  enum class Dialect : uint8_t { Postgres, ClickHouse, };
```

server/connector/pg_logical_types.h
```diff
@@ -117,6 +117,29 @@ inline bool IsOidLike(const duckdb::LogicalType& type) {
...
+inline bool IsCtid(const duckdb::LogicalType& type) {
+  return type.id() == duckdb::LogicalTypeId::STRUCT &&
+         type.GetAlias() == kCtidAlias;
+}
+
+#ifndef SDB_PG_LOGICAL_TYPES_NO_FACTORY
+inline duckdb::LogicalType CTID() {
+  auto type =
+    duckdb::LogicalType::STRUCT({{"page", duckdb::LogicalType::BIGINT},
+                                 {"tuple", duckdb::LogicalType::BIGINT}});
```
MBkkt: (block_number, tuple_offset) and it's integer and short?

server/connector/pg_logical_types.h
```diff
@@ -117,6 +117,29 @@ inline bool IsOidLike(const duckdb::LogicalType& type) {
...
+  type.SetAlias(std::string{kCtidAlias});
```
MBkkt: maybe `static const auto kType = [] {...}();`

server/connector/view_fast_path.cpp
```diff
@@ -199,10 +201,45 @@ duckdb::TableFunction LookupSingleStringReader(duckdb::ClientContext& context,
                           "\" has no (VARCHAR) overload"));
 }
 
+std::vector<std::string> ParseKeyColumns(std::string_view text) {
+  std::vector<std::string> cols;
+  for (auto part : absl::StrSplit(text, ',')) {
+    auto trimmed = absl::StripAsciiWhitespace(part);
+    if (!trimmed.empty()) {
```
MBkkt: I think strsplit have something to skip whitespaces

server/connector/view_fast_path.cpp
```diff
@@ -199,10 +201,45 @@ duckdb::TableFunction LookupSingleStringReader(duckdb::ClientContext& context,
                           "\" has no (VARCHAR) overload"));
 }
 
+std::vector<std::string> ParseKeyColumns(std::string_view text) {
+  std::vector<std::string> cols;
```
MBkkt:
```suggestion
  std::vector<std::string_view> cols;
```

server/connector/view_fast_path.cpp
```diff
@@ -199,10 +201,45 @@ duckdb::TableFunction LookupSingleStringReader(duckdb::ClientContext& context,
...
+  for (auto part : absl::StrSplit(text, ',')) {
```
MBkkt:
```suggestion
  for (std::string_view part : absl::StrSplit(text, ',')) {
```

server/connector/view_fast_path.cpp
```diff
@@ -463,6 +571,20 @@ std::optional<ViewFastPath> ResolveViewFastPath(
...
+  if (fp.pk_spec == catalog::PkSpec::ExternalColumnKey) {
+    // Project the key columns in order so the index sink can key postings by
+    // their values (the ctid rowid as one BIGINT; key columns as a struct).
+    std::vector<duckdb::column_t> ids;
+    ids.reserve(fp.key_columns.size());
+    for (const auto& kc : fp.key_columns) {
+      ids.push_back(kc.source_index);
+    }
+    return ids;
+  }
```
MBkkt, line 603: kal

server/pg/commands/rbac.cpp
```diff
@@ -533,6 +533,11 @@ std::shared_ptr<catalog::Object> ResolveGrantTarget(
     out_name = std::string{raw_name};
     return snap.GetSchema(ctx.GetDatabaseId(), raw_name);
   }
+  if (type == catalog::ObjectType::ForeignServer) {
+    // A foreign server is a database child with no schema (PG-shape).
+    out_name = std::string{raw_name};
```
MBkkt:
```suggestion
    out_name = raw_name;
```

server/pg/pg_catalog/fwd.h
```diff
@@ -84,6 +84,11 @@ struct IsArray : std::false_type {};
 template<typename T>
 struct IsArray<Array<T>> : std::true_type {};
 
+// Owning variant: a row field may hold its array data directly instead of
+// spanning caller-kept storage.
+template<typename T>
+struct IsArray<std::vector<T>> : std::true_type {};
+
```
MBkkt: maybe remove it?

server/pg/pg_catalog/pg_foreign_server.h
```diff
@@ -42,8 +46,11 @@ struct PgForeignServer {
   Text srvtype;
   Text srvversion;
   AclColumn srvacl;
-  Array<Text> srvoptions;
+  std::vector<std::string> srvoptions;
```
MBkkt:
```suggestion
  std::vector<Text> srvoptions;
```

server/pg/sql_utils.h
```diff
@@ -93,6 +97,9 @@ constexpr catalog::ObjectType FromPgObjectTypeName(
   if (word == "TYPE") {
     return PgSqlType;
   }
+  if (word == "FOREIGN SERVER") {
+    return ForeignServer;
+  }
   return Invalid;
```
MBkkt, line 106: add user mapping or remove it entirely

third_party/duckdb_clickhouse/src/include/storage/clickhouse_catalog.hpp
```diff
@@ -0,0 +1,141 @@
...
+	case_insensitive_map_t<unique_ptr<ClickHouseSchemaEntry>> schemas;
+	// Dropped schema entries are parked here (not destroyed) for the catalog's
+	// lifetime, so a concurrently-bound statement holding a raw pointer into a
+	// schema (and its table/retired_tables entries) never dangles.
+	vector<unique_ptr<ClickHouseSchemaEntry>> retired_schemas;
+
+	shared_ptr<ClickHouseConnectionPool> connection_pool;
+
+	std::mutex describe_cache_lock;
+	std::unordered_map<string, DescribeCacheEntry> describe_cache;
```
MBkkt: never use std::mutex and std::unoredered_map!

third_party/duckdb_clickhouse/src/include/storage/clickhouse_table_entry.hpp
```diff
@@ -0,0 +1,70 @@
...
+	//! Approximate row count from system.tables.total_rows, fetched lazily on the first
+	//! scan and cached for the entry's lifetime (clickhouse_clear_cache rebuilds it).
+	//! row_count_known is false when the server reports NULL (non-MergeTree engines).
+	//! Guarded by row_count_lock: concurrent binds of the same entry race otherwise.
+	std::mutex row_count_lock;
+	idx_t cached_row_count = 0;
+	bool cached_row_count_known = false;
+	bool row_count_fetched = false;
+	//! MergeTree-family engine (from system.tables.engine): PREWHERE is legal.
+	bool cached_is_merge_tree = false;
```
MBkkt: maybe remove prewhere shit, clickhouse should optimize it automatically, if not we can create issue maybe?

third_party/duckdb_clickhouse/src/include/clickhouse_connection.hpp
```diff
@@ -0,0 +1,83 @@
...
+#include <clickhouse/client.h>
+
+#include <cstdint>
+#include <memory>
+#include <string>
+
+namespace duckdb {
+
+struct ClickHouseConnectionParams {
+	std::string host;
```
MBkkt: in some place std types, in other duckdb, lets follow duckdb rules

third_party/duckdb_clickhouse/src/storage/clickhouse_catalog.cpp
```diff
@@ -0,0 +1,245 @@
...
+		connection.Invalidate();
+		ClickHouseConnection::ThrowError("reading database size", sql, e);
+	} catch (...) {
+		// Socket-level failures (std::system_error) and exceptions thrown by
+		// block callbacks are not clickhouse::Error; the connection may be
+		// mid-stream either way, so never hand it back to the pool.
+		connection.Invalidate();
+		throw;
+		}
+	return size;
```
MBkkt, line 236: we have ./scripts/format_duckdb.sh it should format this dir too

third_party/duckdb_clickhouse/src/storage/clickhouse_catalog.cpp
```diff
@@ -0,0 +1,245 @@
...
+}
+
+bool ClickHouseCatalog::InMemory() {
+	return false;
+}
+
+string ClickHouseCatalog::GetDBPath() {
+	return params.host + ":" + std::to_string(params.port) + "/" + params.database;
```
MBkkt: absl::StrCat, without std::to_string

third_party/duckdb_clickhouse/src/storage/clickhouse_schema_entry.cpp
```diff
@@ -0,0 +1,609 @@
...
+			sql += ", ";
+		}
+		bool nullable = !is_not_null(i, col.GetName().GetIdentifierName());
+		sql += ClickHouseQuoteIdentifier(col.GetName().GetIdentifierName()) + " " + LogicalTypeToClickHouseType(col.GetType(), nullable);
+		auto default_literal = RenderConstantDefault(col);
+		if (!default_literal.empty()) {
+			sql += " DEFAULT " + default_literal;
+		}
+	}
+	sql += ") ENGINE = MergeTree ORDER BY ";
```
MBkkt, line 164: it's probably some parameter? engine=?

third_party/duckdb_clickhouse/src/clickhouse_connection.cpp
```diff
@@ -0,0 +1,304 @@
...
+#include "clickhouse_connection.hpp"
+
+namespace duckdb {
+
+static void ApplyParam(ClickHouseConnectionParams &result, const string &key, const string &value) {
+	auto lkey = StringUtil::Lower(key);
+	if (lkey == "host" || lkey == "hostname") {
+		result.host = value;
+	} else if (lkey == "port") {
+		result.port = NumericCast<uint16_t>(std::stoul(value));
```
MBkkt: don't use std::stoi/ul/etc

third_party/clickhouse-cpp
MBkkt, line 1: license.md

third_party/CMakeLists.txt
```diff
@@ -538,6 +538,41 @@ set(CURL_INCLUDE_DIRS
...
+)
+
+target_compile_definitions(
+    clickhouse_cpp
+    PUBLIC WITH_OPENSSL=1 CH_MAP_BOOL_TO_UINT8=1
+)
+
+target_link_libraries(
+    clickhouse_cpp
+    PUBLIC lz4::lz4 zstd::zstd absl::int128 OpenSSL::SSL OpenSSL::Crypto
```
MBkkt: after rebase check lz4::lz4 and zstd::zstd still presented

third_party/CMakeLists.txt
```diff
@@ -538,6 +538,41 @@ set(CURL_INCLUDE_DIRS
...
+    clickhouse_cpp
+    SYSTEM
+    PUBLIC
+        "${CMAKE_CURRENT_SOURCE_DIR}/clickhouse-cpp"
+        "${CMAKE_CURRENT_SOURCE_DIR}/clickhouse-cpp/contrib/cityhash/cityhash"
+)
+
+target_compile_definitions(
+    clickhouse_cpp
+    PUBLIC WITH_OPENSSL=1 CH_MAP_BOOL_TO_UINT8=1
```
MBkkt, line 605: why map bool to u8?

MBkkt (approved):

PR #882, opened 2026-07-04
feat: Term dict

gnusi:
## What

Adds a term-dictionary access mode to inverted indexes: a family of
`ts_dict_*` aggregates that enumerate a field's term dictionary straight from
the index instead of scanning documents. This turns "what terms does this
field hold, and how are they distributed" into an index-only read.

```sql
-- every distinct term in body, in byte order, with per-term stats
SELECT unnest(ts_dict_agg(body))   AS term,
       unnest(ts_dict_count(body)) AS docs,
       unnest(ts_dict_freq(body))  AS freq
FROM   docs_idx
ORDER  BY term;

-- term conditions filter the dictionary; WHERE filters the documents
SELECT unnest(ts_dict_agg(body)) AS term
FROM   docs_idx
WHERE  body @@ ts_starts_with('a') AND body @@ ts_regexp('a.*e')
ORDER  BY term;

-- keyword facet: groups are the field's live terms, counts are live doc counts
SELECT cat, count(*) AS n FROM docs_idx GROUP BY cat ORDER BY cat;
```

## SQL surface

| function | returns |
|---|---|
| `ts_dict_agg(col)` | list of terms (decoded text) |
| `ts_dict_raw_agg(col)` | list of raw term bytes |
| `ts_dict_count(col)` | per-term live document counts |
| `ts_dict_freq(col)` | per-term total frequencies |
| `ts_dict_score(col)` | per-term score of the driving acceptor |
| `ts_dict_min(col)` / `ts_dict_max(col)` | least / greatest term |

`count(DISTINCT col)`, `min(col)`, `max(col)` and `array_agg(DISTINCT col)`
over a keyword column, and the `SELECT col, count(*) ... GROUP BY col` facet
shape, are rewritten onto this path automatically — no `ts_dict_*` call
required. Anything the rewrite can't serve exactly (nullable-column WHERE,
extra aggregates, multiple group keys, cross-field WHERE) falls back to the
normal document scan.

## Semantics

The mental model is two independent filter positions:

- **`WHERE` filters documents.** Document conjuncts fuse into a
  `ProxyFilter`-cached per-segment live-doc set, probed once per segment
  through a conjunction; deletes are respected.
- **Term conditions filter terms.** A term-level conjunct (a comparison or
  `@@ ts_*` matcher over the enumerated field, whether in `HAVING` or in an
  outer filter over the `unnest`) compiles into a per-request having filter
  that *drives* the enumeration — via dictionary seeks, an automaton walk, or
  a per-term predicate.

The scan counts per `CountMode`: term-dictionary metadata on clean segments,
mask- or where-conjunction counting under deletes, and `min`/`max` pinned to
seek probes so they don't pay for a full walk.

## Internals

- **Acceptor fusion moved out of the planner into the iresearch filter
  optimizer.** Same-field `OR`s of term/prefix/wildcard/regexp acceptors now
  render a single regex compiled through RE2 (`OrAcceptorFusionRule`), and
  `AND` conjunctions intersect-fold their compiled acceptors post-lowering
  (`FuseIntersections`). Both rules are context-gated so the ordinary search
  path keeps its native seek-based disjunctions; fusion only kicks in where it
  measured as a win (recorded in the microbench ledger).
- `UnionAcceptors` and the old planner-side fusion machinery are deleted
  (~200 lines out of `iresearch_plan.cpp`).
- Rebased onto the DuckDB `v2026.06.29` bump; the connector is ported to the
  new privatized-member accessor API.

## Testing

- 20 `ts_dict_*` sqllogic tests plus the full `sdb/pg/index` sweep, green on
  both `pg-wire-simple` and `pg-wire-extended`.
- 22 fusion gtests (`OrAcceptorFusion` / `AndAcceptorFusion`) plus the
  iresearch filter/automaton/formats suites.
- Microbench rows gate the fusion routing decision.

## Follow-ups

Tracked in #896 (pattern-cache LRU, RE2 DFA build tuning, `NOT` lowering,
multi-field unnest pushdown).


server/connector/optimizer/iresearch_plan.cpp
```diff
@@ -848,67 +1973,497 @@ bool TryClaimSearchFilter(
...
+        for (const auto& req : ss.ts_dicts) {
+          if (req.field_id == field) {
+            return true;
+          }
+        }
+        return false;
+      };
+      const std::function<bool(const duckdb::Expression&, bool&)>
+        refs_term_level = [&](const duckdb::Expression& expr,
+                              bool& saw_analyzed) -> bool {
```
MBkkt: make it deducing this instead of std::function

server/connector/optimizer/iresearch_plan.cpp
```diff
@@ -848,67 +1973,497 @@ bool TryClaimSearchFilter(
...
+      // The four execution modes of the where/having model: term-level
+      // conjuncts drive or predicate the enumeration (having side), the
+      // rest restrict documents (where side, executed once per segment) or
+      // post-filter term rows.
+      enum class Route {
+        kHavingFast,  // term-level acceptor tree -> having_filter
+        kWhere,       // document filter -> where proxy
+        kPostFilter,  // term-level scalar over the emitted term rows
+        kRejected,
+      };
```
MBkkt, line 2335: incorrect codestyle, should be without `k`

MBkkt (approved):

Issue #23, opened 2025-12-06
Parallelizing Index Commit and Document Removal


pashandor789:
The following write-related operations exist:  
1. Segment flush  
2. Segment compaction  
3. Index commit  
4. Applying index deletions  
5. Writing to a segment  

Currently, flushing is parallel when it occurs during insertion, but **not parallel when it happens during commit and compaction**. The goal is to make it **asynchronous for insertions** and **parallel for commit/compaction**.

Applying deletions happens during commit and is therefore **not parallel**. The goal is to make it **parallel**.

PR #884, opened 2026-07-04
feat: Table filters for inverted-index scans

pashandor789:
Inverted-index scans now accept table filters instead of rejecting pushdown: filters are enforced inside the index scan, so a filtered top-k (ORDER BY score LIMIT k) collects passing rows only instead of scoring everything and filtering above the scan.

**Filter planes.** A pushed filter routes by where its column lives:

- Covered (INCLUDE'd) `.col` columns: applied in-scan against the index columnstore — codec-level Filter plus per-row-group zonemap short-circuit (ALWAYS_TRUE / ALWAYS_FALSE), with one decode pass doubling as materialization, RowGroup::Scan-style. Query-driven scans wrap the search DocIterator in a TableFilterDocIterator; the staged emit path applies the same FilterSpecs in HitBatcher when draining hit windows; match-all scans run FullScanner. No `.col` filters ⇒ no wrapper, zero cost.
- The score column: applied to the computed score vector after scoring (it is not a columnstore field).
- Lookup (source-only) columns: these can only be evaluated during the source lookup, so they forbid the collector/bulk fast paths and force the streaming scan; the filter runs through a rowid lookup on the base source using duckdb's native TableFilterSet machinery, and survivors come back compact in pk order. Works for tables and view sources.
- Anything not expressible as a table filter stays a row expression above the scan.

**Streaming single-term WAND (max_score_iterator).** Scored streaming behind a dynamic max-score threshold, used in two places: as the windowed scored drain under filtered collection (TableFilterDocIterator::Collect), and as streaming WAND when TopN orders by (score, tie-breaker column), where heap-only top-k collection does not apply.

**Search tables unified on iresearch_scan.** The bespoke search_table_scan is removed; search tables scan through the same iresearch_scan path as index scans. With a single consumer left, the ScanSource interface is de-virtualized (virtual Clone / EmitDocs / EmitRowOffsets / Materialize / OnSegment hooks replaced by direct calls).

**Parquet lookup.** The lookup plane over parquet view sources accepts pushed filters too, and no longer re-scans the whole row group per output batch: a resumable per-row-group cursor with batched scatter takes dense lookups from ~40–66x the cost of a native scan down to ~1.6–3.1x.

**duckdb submodule:**

- Table functions answer filter pushdown per filter (BeforeLimit / AfterLimit / Reject / Drop) instead of all-or-nothing; pushed and rehosted filters render uniformly in EXPLAIN.
- Rowid lookup scan backing the lookup plane.
- TopN lifts its child projection above the operator, ColumnLifetimeAnalyzer prunes columns through a future TopN, and identity projections under aggregates are elided.
- CSV reader: utf-16le/be and utf-32le/be encodings; base64 encode/decode and UTF-8/ASCII validation via simdutf (simdutf submodule bumped).
- Version-cleanup horizon clamp under the durable snapshot bound — fixes a catalog-entry use-after-free under concurrent duckdb_tables() enumeration and DDL churn.
- Hidden-database enumeration opt-in: catalog-scoped duckdb_* table functions take include_hidden := true, and GetAllSchemas pins enumerated databases against concurrent DETACH; serenedb attaches __sdb_store hidden so store internals stop leaking into catalog listings.
- HistogramAggState's MAP_TYPE template parameter renamed (collides with glibc's MAP_TYPE macro).

**Vector scoring** is unified to larger-is-nearer at the reader/kernel level (quantized/IVF included); user-facing scores map through ScoreEmit (L2 negates, cosine 1−x).

**Index options:** row_group_size is columnstore-wide only (per-column form removed; norm_row_group_size stays per-column). Filtered-out rows and lookup misses behave identically (dropped).

**Coverage:** new large-cardinality (10–100k rows) sqllogic suites for `.col` filters (dict reuse, mixed blocks, multichunk), lookup filters, view/parquet filters, pushdown planes, top-k fallback, WAND + table filter, segment pruning, and index joins.

Co-authored-by: pashandor789 <96681992+pashandor789@users.noreply.github.com>
Co-authored-by: Mikhail Kornaukhov <66915223+mkornaukhov@users.noreply.github.com>


pashandor789:
Review completed; fixes pushed as b2df13882.

**Important status note:** the filtered top-k fast path in this PR is currently **dormant** — the fork's RowGroupPruner invalidates `row_limit` whenever a FILTER remains between ORDER BY and the scan (row_group_pruner.cpp:54/79), so `score_top_k` is never set for filtered queries and they run the streaming path (~55ms on the study dataset vs ~1ms when the path engages). The machinery lands correct and gated, and activates once the filter coordinator work re-claims column filters into the scan.

Fixed in b2df13882: unsafe Collect prologue removed (window-refill iterators like block_disjunction would drop candidates), hard error on unreadable filter columns (was release no-op assert), Complete() bookkeeping, out-of-range row_pos semantics unified (incl. an OOB read in MatchesBatch), per-candidate dynamic-filter stats recomputation eliminated, O(1) merged-stats all-true probe, .h->.hpp rename, clang-format (pinned v21), session artifacts removed.

Known remaining (deliberate, for the coordinator work): no emit-side filter backstop on the top-k path — enforcement relies on DocPruner coverage + the new hard error.

pashandor789:
**Correction to my July-4 dormancy comment — root cause found and fixed in 0a2b47e5f.**

The July-3 rebase dropped `func.filter_pushdown = true` from `CreateIResearchScanFunction` (collateral of removing the adjacent, legitimately-obsolete `filter_to_string` line). Since then DuckDB never converted WHERE conjuncts into table filters for iresearch scans — the residual FILTER operator above the scan enforced them instead, and the entire scan-side machinery in this PR (`prune_filters`, DocPruner row-group skip, in-scan enforcement, Column Filter labels) was unreachable. My July-4 comment blamed `RowGroupPruner` row-limit invalidation; that was the downstream symptom, not the cause.

Restored + regenerated the affected EXPLAIN expectations. Oracle-arm A/B (flag off vs on, interleaved, 500k docs): clustered filters **g>=95 −71%, g<=5 −77%, r<90 AND g>=95 −71%**; non-clustered cases within ±2% noise. Correctness: 30/30 text A/B (both index types), ANN leak gate 12/12, full sdb+any sqllogic suites, 9/9 recovery tests.

Still true from the July-4 comment: filtered **top-k** stays dormant — with filters now pushed into the get, `RowGroupPruner` invalidates `row_limit` (correct for DuckDB's own count-based reorderer, pessimistic for our scan which counts k after the predicate). Re-activating it needs a fork-side capability ("scan enforces pushed filters before the limit counts") — folded into the filter-coordinator work (#21–#23).

pashandor789:
why am i having a conversation with myself ? schizophrenia.

CONTRIBUTING.md
```diff
@@ -309,8 +309,8 @@ Similar to [Google style](https://google.github.io/styleguide/cppguide.html#Func
 ### Error Handling
 
 - PostgreSQL/frontend code: use `THROW_SQL_ERROR`
-- Common/backend code: both `Result<>` and `SDB_THROW` are acceptable
-- Consider performance: `Result` with a default message should not allocate
+- Common/backend code: both `absl::Status` and `throw` are acceptable
```
pashandor789, line 312: THROW_SQL_ERROR

server/connector/index_source_view_file.cpp
```diff
@@ -149,25 +194,44 @@ void ViewFileGlobIndexSource::Materialize(duckdb::ClientContext& context,
...
+    _lookup_func.function(context, in, _file_target);
+
+    const auto file_rows = _file_target.size();
+    for (duckdb::idx_t c = 0; c < _real_proj_slots.size(); ++c) {
+      duckdb::VectorOperations::Copy(_file_target.data[c], _tf_target.data[c],
+                                     file_rows, /*source_offset=*/0,
+                                     /*target_offset=*/total);
+    }
+    // The reader wrote survivors as 0-based indices into this file's pk_lookups
+    // span (which starts at _sorted_rows[i]); shift to batch-global sorted-pk
```
pashandor789, line 222: why did it appear

MBkkt (approved):

PR #908, opened 2026-07-12
feat: Improve RBAC flags

pashandor789:
## Simplify first-connect auth (RBAC flags cleanup)

Makes SereneDB trivial to connect to locally and safe the moment it's exposed —
without the operator touching passwords, HBA, or flags first.

### What changed, in plain terms

- **Root now connects for free — but only locally.** On a fresh install,
  `psql -h 127.0.0.1 -U postgres` just works, no password. The same over a unix
  socket works too.
- **Root over the network needs a password.** Bind `0.0.0.0` and connect
  `postgres` from another host with no password set → **refused** (used to be
  silently trusted — a passwordless superuser on the network). Set one with
  `ALTER ROLE postgres PASSWORD '…'` and remote works.
- **Everyone else always needs a password.** A passwordless non-superuser role
  can't log in anywhere — locally or remotely (SCRAM), same as Postgres.
- **Containers stay one step.** `SERENEDB_INITIAL_PASSWORD=secret` on first boot
  seeds root's password, so a container comes up remote-reachable *and*
  authenticated in one action.
- **A bad HBA config can't lock the admin out.** Root keeps a guaranteed
  loopback `trust` channel no matter what the operator writes.

### How

The whole policy lives in HBA now — no per-connection special-casing. The
default (no-config) ruleset is:

```
local all postgres      trust          # bootstrap superuser: always in, locally
host  all postgres 127.0.0.1/32 trust  #   (anti-lockout escape hatch)
host  all postgres ::1/128     trust
local all all           scram-sha-256  # everyone else: password required...
host  all all 0.0.0.0/0 scram-sha-256  #   ...locally and remotely alike
host  all all ::0/0     scram-sha-256
```

The three superuser `trust` lines (`SafetyRules()`) are force-prepended onto
**every** ruleset — the built-in default and any operator-authored one — so
first-match-wins guarantees root can always authenticate on loopback. Operators
can `reject`/password-gate loopback for non-superusers freely; the old
un-overridable loopback-trust is gone.

### Cleanup

- Deleted pre-RBAC flags `--auth_password` / `--auth_method` / `--auth_user`
  (`--auth_password` shadowed RBAC by flipping the server to one hard-coded
  user; the others were inert). Per-role catalog passwords are now the only
  auth path.
- Deleted the `is_default` flag and the per-session trust fallback — trust is a
  property of the HBA rules, not a code branch.
- Closed the remote-passwordless hole on both pg-wire and HTTP Basic.

### Tests

- gtest `SuperuserLocalTrustEveryoneElseScram` — full matrix (root: trust
  local+loopback, SCRAM remote; other roles: SCRAM everywhere) + `reject`-all
  can't lock root out of loopback.
- python HBA matching suite; rbac sqllogic suite green on both engines.


server/network/CMakeLists.txt
```diff
@@ -2,7 +2,6 @@ target_sources(
     sereneserver
     PRIVATE
         io_context.cpp
-        ada_re2_provider.cpp
```
MBkkt, line 5: hmm, but file isn't removed, I think we need to make ada (url parser) to use re2

libs/geo/geo_json.cpp
```diff
@@ -232,95 +232,83 @@ bool ParseImpl(ondemand::value vertex_value, S2LatLng& vertex) {
...
 template<Type T>
-Result Validate(ondemand::object& object, ondemand::value& coordinates) {
+void Validate(ondemand::object& object, ondemand::value& coordinates) {
   if (ParseType(object) != T) [[unlikely]] {
     // TODO(mbkkt) Unnecessary memcpy,
     //  this string can be constructed at compile time with compile time new
-    return {ERROR_BAD_PARAMETER,
-            absl::StrCat("Require type: '", toString<T>(), "'.")};
+    THROW_SQL_ERROR(
+      ERR_MSG(absl::StrCat("Require type: '", toString<T>(), "'.")));
```
MBkkt: inspired by kaldb

MBkkt (approved):
Followups:
- ada, should use re2 and handle url parsing entirely
- support more auth methods peer/gss/ident/ldap/gss/cert
- remove sdb::basics::Buffer -- check other classes/functions/files for same issue (maybe used in tests only)
- remove generate_exit_codes.py
- investigate about env variables and trust/etc
- doesn't allow just throw in our files, not only duckdb exception

PR #926, opened 2026-07-16
feat: GROUPING SETS + tsquery rework

### LINKED ISSUE #919: Serve GROUPING SETS facet queries from term dictionaries in one index pass
## Workload

Every faceted search UI needs, per query: value counts for each of K facet columns under one match predicate (the sidebar). Today there are two options, both with a measured flaw.

**K separate queries** get the fast plan — `GROUP BY brand` over an inverted index converts to a term-dictionary read:

```
IRESEARCH_SCAN  Index: gs_idx, TsDict: brand
  Projections: sdb_inverted_index_term$brand, sdb_inverted_index_term_count$brand
HASH_GROUP_BY   sum(#1)
```

but the WHERE match predicate is compiled and evaluated K times, once per query, and match evaluation dominates on real corpora (the dictionary walk is nearly free). Sidebar cost ~= K x match-evaluation.

**One GROUPING SETS query** already parses and returns correct results (`GROUPING()` disambiguates the sets), but the conversion does not fire — the plan materializes every matching row's raw columns into a generic hash aggregate:

```
IRESEARCH_SCAN  Index: gs_idx        -- no TsDict: line
  Projections: brand, status, kind   -- row materialization of all matches
HASH_GROUP_BY
```

One pass, but it trades away the term-dictionary path entirely.

## Proposal

Extend the existing single-column conversion to recognize
`GROUP BY GROUPING SETS ((brand), (status), (kind))` where every set is one
bare term-dict-eligible column and the aggregates are those the
single-column path already serves. Plan it as **one IRESEARCH_SCAN walking K
term dictionaries with one shared compiled WHERE per segment**, emitting
`(grouping_id, term, count)`; the upper HASH_GROUP_BY groups by
`(grouping_id, term)` and sums across segments — today's design plus one id
column.

```sql
SELECT GROUPING(brand, status, kind) AS facet, brand, status, kind, count(*) AS n
FROM search_docs_idx
WHERE title @@ $q OR body @@ $q
GROUP BY GROUPING SETS ((brand), (status), (kind));
```

Cost becomes 1 x match-evaluation + K dictionary walks.

## Why this is cheaper to build than it sounds

The scan side is already multi-field: the ts_dict aggregate path serves
several fields in one scan (`SearchScan.ts_dicts` is a vector; the
NULL-padding + `FILTER (IS NOT NULL)` handling exists for exactly that),
and WHERE-restricted counting computes a per-segment live-docs bitset
(`WhereLiveDocs`) that is naturally shareable across the K enumerations —
that sharing is the entire win. The new work is optimizer pattern-matching
(GROUPING SETS -> multi-field TsDict scan) plus synthesizing `GROUPING()`.

## Guardrails (mirror the existing conversion rules)

- Single-column sets only; ROLLUP / CUBE / multi-column sets fall back to
  the row plan (they need row co-occurrence, which independent term
  dictionaries cannot provide).
- NULL groups via the null-marker fields, as in the single-column
  conversion.
- HAVING / ORDER BY / per-set LIMIT remain in the upper plan.

### LINKED ISSUE #920: First-class TSQUERY values: AST with per-node boost and tokenizer bindings
## Problem

`TSQUERY` today is a bind-time marker, not a value. The whole `@@` RHS is an expression tree the optimizer claims and compiles into an iresearch filter during planning; `ts_*` constructors and casts are stubs that error outside that context. Boost is smuggled through two mechanisms (`^` as an expression node, `::boost(K)` as a type modifier on `BOOSTED_TSQUERY`), and the tokenizer binding is ambient (a bare string adopts the target column's analyzer at bind time).

This design bought full plan-time compilation, but four costs have now materialized in practice:

1. **Queries are uninspectable.** The recent multiterm-boost bug (boosts on `ts_any`/multi-token/prefix queries silently ignored in scoring) survived because nothing can render a compiled query; there was no way to `SELECT` a query and see the boost was present but unused.
2. **No parameterization.** `WHERE title @@ $1` with a query parameter is impossible, so applications interpolate SQL literals (the Highsail demo backend does exactly this).
3. **No storage.** Saved searches are re-parsed strings; `ts_rewrite`-style expansion and percolation/alerting (store queries, match arriving documents) are structurally blocked.
4. **PG divergence.** In PostgreSQL `tsquery` is a real type: storable, parameter-bindable, with input/output syntax.

## Proposal

Make `TSQUERY` a first-class value: a small **versioned AST** with text and binary encodings.

- **Node kinds**: term, phrase (slop), prefix, wildcard, regexp, fuzzy (distance/transpositions/prefix), range, boolean (and/or/not, `min_match`).
- **Boost**: a per-node attribute. `^` and `::boost(K)` both construct it; the `BOOSTED_TSQUERY` type-modifier mechanism is retired.
- **Tokenizer**: an *optional per-subtree* reference (what `::tokenize('kw')` becomes). Text in unbound nodes stays **untokenized**; binding to a column at `@@` supplies the default analyzer, preserving today's ergonomics (one value works across columns with different analyzers). An eager, PG-style pre-tokenized form remains available via `to_tsquery(dict, text)`.
- **`@@` becomes a genuine binary operator** over (column, tsquery value).

## Execution model

- **Constant folding preserves today's performance**: a tsquery expression built from literals folds to an AST literal at bind time and compiles to the iresearch filter during planning, exactly as now. Zero regression for every existing query shape.
- **Late binding is the new capability**: a parameter (or column-sourced value) compiles at execution time; the scan accepts a runtime filter. Top-k/WAND eligibility decisions move to execution for this path (v1 may simply not WAND-optimize late-bound queries).
- Structurally: `search_filter_builder` splits into (bound expressions -> AST) and (AST -> irs filter); the AST is the value's storage format.

## Phasing

1. **Value + I/O + inspectability**: the type, text rendering (`SELECT ('a b'::tsquery ^ 2)::text` -> `("a" | "b")^2.0`), constant-folding integration. No behavior change; immediately makes query compilation testable.
2. **Parameters**: extended-protocol binding + late-bound scan path.
3. **Storage**: tsquery columns, saved searches as data, `ts_rewrite`.
4. **Percolation**: stored queries matched against incoming documents.

## Open questions

- Text syntax: PG-compatible where possible vs. the richer surface (fuzzy/ranges/tokenizer bindings have no PG equivalent).
- Wire representation: PG's `tsquery` OID expects lexeme-based binary; likely a distinct type OID with documented divergence.
- Dictionary staleness for stored values: resolve-by-name at use (error on missing) vs. content hash pinning.
- Interaction with #917 (DDL field weights): declared weights apply as defaults where a node carries no explicit boost — cleanly answerable once boost is a node attribute.

gnusi:
Adds `TSQUERY` as a first-class value type so full-text predicates can be built, parameterized, and passed around like any SQL value instead of only as inline `@@ ts_*(...)` calls; fixes SQL three-valued-logic semantics for claimed negations; and widens the existing GROUP BY facet pushdown to many more shapes.

## TSQUERY value type
- Carries `{text, tokenizer, boost}`. Constant constructor trees (`ts_like`, `ts_phrase`, `ts_compound`, `^` boost, `::tokenizer`) fold to a canonical text value at plan time and re-parse structurally when the filter is claimed; non-constant or over-cap inputs fall back to a bag-of-words term filter. Caps: 64 KiB text, depth 64, 4096 nodes.
- `col @@ $1` prepared-statement parameters and arrays of tsquery via `ts_compound` / `ts_any` / `ts_all`. Wire (de)serialization added to the pg protocol.

## NULL semantics for negation
- A NULL row satisfies no comparison, but a bare `irs::Not` runs against *all* live docs and would readmit it. The claim loop threads a null-marker map (null field → column field) into the boolean optimizer, which hoists one `IS NOT NULL` guard per negated column to the conjunction level — elided when a positive sibling on the same column already rejects NULLs, and replaced by a local null exclusion under a disjunction where the guard is unsound. `IS [NOT] NULL` claims against the null-marker term, `IS DISTINCT FROM` keeps NULL rows, and range negation is absorbed by operator inversion (postings-only, null-safe).

## Facet pushdown — extended (`optimizer/ts_dict_plan`)
The `SELECT term, count(*) … GROUP BY <keyword col>` → term-dictionary scan already existed, but only for a single plain NOT-NULL key, one grouping set, and bare same-field WHERE. Extended to cover:
- Multiple group keys over distinct keyword fields.
- GROUPING SETS over singleton keyword keys, collapsed to one multi-field term scan (≤ 1 nullable key — a second nullable key's NULL group would collide with the first's all-NULL group key).
- Indexed keyword *expression* keys, not just plain columns.
- Index-claimable WHERE: cross-field and `@@ ts_*` tsquery conjuncts compile into the scan's document filter and same-field term-acceptor conjuncts prune their own enumeration — vs. only bare same-field predicates (and no WHERE for nullable keys) before.

## Refactor
- Extracted the tsquery surface out of `search.cpp` into `functions/ts_query.{h,cpp}` (registration) and `functions/ts_query_codec.{h,cpp}` (value codec / whitelist re-parse).
- Split the term-dictionary planner out of `iresearch_plan.cpp` into `optimizer/ts_dict_plan.{hpp,cpp}` with shared helpers in `iresearch_plan_common.hpp`.

## Estimation
- The term-dictionary scan's cardinality is the sum of per-segment term counts — the exact number of rows the scan emits — reduced by the term-level having filter but not by the document-level stored filter (a doc filter shrinks the vocabulary only sublinearly, so applying a flat selectivity to a term count is wrong). Since it depends on segment layout, the `ts_dict` tests pin the segment count by disabling background maintenance (`SET refresh_interval = 0, compaction_interval = 0`) and assert exact row counts. EXPLAIN resolves indexed expressions by field id so plan structure stays stable across server lifetime.
- `ColumnIsNotNull` resolves the column via the table's O(1) id index and matches check constraints by name in a single pass (new `CheckConstraint::NotNullColumnName()`), replacing a redundant column scan plus a per-constraint column scan.

## Tests
New `tsquery_type`, `tsquery_params`, `null_semantics` (+ fuzz) sqllogic suites plus C++ coverage in `duckdb_search_filter_builder_test`; facet and EXPLAIN expectations regenerated. Connector suite 489/489.


libs/iresearch/include/iresearch/search/block_disjunction.hpp
```diff
@@ -596,11 +596,15 @@ class BlockDisjunction : public DocIterator {
...
+        // The sub may sit behind the window: one step behind before the
+        // very first next() or after seek() ('kSeekReadahead == false'),
+        // arbitrarily far behind after a LazySeek probe. Recover with
+        // seek(), never advance(): a probe may have moved the sub's
+        // internals past value() (e.g. an exclusion's include), and
+        // advance() would skip that committed position while seek()
+        // re-settles on it. FillBlock requires value >= _doc_base.
         if (value < _doc_base) {
-          value = it.advance();
+          value = it.seek(_doc_base);
```
gnusi, line 607: Seek might be heavier than advance(), is there a way to keep it?

libs/iresearch/include/iresearch/formats/ivf/quantizer.cpp
```diff
@@ -543,7 +543,7 @@ class ProductQuantizerReader final : public QuantizerReader {
   std::vector<uint8_t> _lutq;
   faiss::AlignedTable<uint8_t> _packed_lut;
   std::vector<byte_type> _codes_buf;
-  faiss::AlignedTable<uint16_t> _accu;
+  std::vector<uint16_t> _accu;
```
MBkkt: bad

libs/iresearch/include/iresearch/search/filter.hpp
```diff
@@ -79,6 +78,12 @@ class QueryBuilder : public memory::Managed {
   virtual DocIterator::ptr Execute(const ExecutionContext& ctx,
                                    const StatsBuffer& stats) const = 0;
 
+  virtual bool CollectTopK(ScoreCollector& /*collector*/,
+                           const ExecutionContext& /*ctx*/,
+                           const StatsBuffer& /*stats*/) const {
+    return false;
+  }
+
```
MBkkt: conflicts were not resolved correctly at all

MBkkt (changes requested):

libs/iresearch/include/iresearch/search/boolean_filter.hpp
```diff
@@ -180,6 +182,16 @@ class Exclusion : public FilterWithType<Exclusion> {
...
   }
 
+  template<typename Pred>
+  bool EraseExcludesIf(Pred&& pred) {
+    SDB_ASSERT(!_filters.empty());
+    const auto tail = std::remove_if(std::next(_filters.begin()),
+                                     _filters.end(), std::forward<Pred>(pred));
+    const bool erased = tail != _filters.end();
+    _filters.erase(tail, _filters.end());
+    return erased;
```
MBkkt, line 192: erase_if?

gnusi: std::erase_if accepts a vector, but we need to skip the first element

server/connector/functions/search.cpp
```diff
@@ -74,15 +69,7 @@ void SearchStubFn(duckdb::DataChunk& /*args*/,
...
-
-void TSQueryStubFn(duckdb::DataChunk& /*args*/,
-                   duckdb::ExpressionState& /*state*/,
-                   duckdb::Vector& /*result*/) {
-  THROW_SQL_ERROR(
-    ERR_CODE(ERRCODE_FEATURE_NOT_SUPPORTED),
-    ERR_MSG("TSQUERY expression evaluated outside an `@@` match against an "
-            "inverted-indexed column."));
-}
+namespace {
```
MBkkt, line 73: codestyle?

server/connector/optimizer/iresearch_plan.cpp
```diff
@@ -332,6 +264,194 @@ duckdb::idx_t AppendVirtualGetColumn(connector::SereneDBScanBindData& bind_data,
...
+  const auto it = absl::c_find_if(
+    cols, [&](const auto& col) { return col.GetId() == col_id; });
+  if (it == cols.end()) {
+    return false;
+  }
+  const auto idx = static_cast<size_t>(it - cols.begin());
+  return absl::c_any_of(table.CheckConstraints(), [&](const auto& cc) {
+    return cc.IsNotNull(cols) == idx;
+  });
+}
```
MBkkt, line 276: linear checks are mess

server/network/pg/pg_wire_session.cpp
```diff
@@ -2254,8 +2228,49 @@ void PgWireSession<Kind>::DescribePortal(Portal& portal) {
   if (_txn_state->StatusByte() == 'E') {
     ThrowAbortedTransaction();
   }
+  WriteResolvedRowDescription(prepared, &portal.bind_info.param_values,
+                              portal.bind_info.output_formats);
```
MBkkt, line 2232: do we have test for this?

tests/sqllogic/sdb/pg/index/filter_optimizer_negation.test
```diff
@@ -24,19 +24,20 @@ query
...
+│ ╭─ Not ────────────╮                     │
+│ ╰─────────┬────────╯                     │
+│           ├────────────────────╮         │
+│ ╭─ Term ──┴────────╮  ╭─ Term ─┴───────╮ │
+│ │ Field: t(string) │  │ Field: t(null) │ │
+│ │ Value: alpha     │  │ Value:         │ │
+│ ╰──────────────────╯  ╰────────────────╯ │
+│ Projections: id                          │
+│ ~1 row                                   │
+╰──────────────────────────────────────────╯
```
MBkkt, line 40: looks like bug, same for other queries here, I doubt we want to make everything slower, just to make NULL => NULL

tests/sqllogic/sdb/pg/index/ts_dict.test
```diff
@@ -194,20 +194,20 @@ EXPLAIN SELECT unnest(ts_dict_agg(body)) AS term, unnest(ts_dict_score(body)) AS
...
 ╰──────────────────┬──────────────────╯
 ╭─ IRESEARCH_SCAN ─┴──────────────────╮
 │ Index: td_idx                       │
 │ TsDict: body                        │
 │ Projections:                        │
 │ sdb_inverted_index_term$body,       │
 │ sdb_inverted_index_term_score$body  │
-│ ~5 rows                             │
+│ ~<slt:ignore>│
 ╰─────────────────────────────────────╯
```
MBkkt: revert this, it's awful, slt:ignore barely work, this should be determenistic

MBkkt (commented):
Don't use slt:ignore for approximate rows

server/connector/optimizer/iresearch_plan.cpp
gnusi: Adjusted

server/connector/functions/search.cpp
gnusi: Fixed

server/network/pg/pg_wire_session.cpp
gnusi: Yes, added

MBkkt (approved):

PR #931, opened 2026-07-18
feat: Online parallel inverted CREATE INDEX

pashandor789:
### Online (non-blocking) inverted CREATE INDEX
- The backfill runs as a parallel sink over the live index storage in **commit-on-flush** mode: each sink thread's transaction cuts (flush + commit) whenever its segment fills, so segments become durable and visible **during** the build instead of one giant publish at the end. Backfill transactions acquire private segments (`CreateSegmentContext`, no pool adoption), so DML and backfill never mix docs in one segment and refresh is never pinned by the build.
- `irs::IndexWriter::Transaction` gains `SetCommitOnFlush()` and an out-param on `Insert(..., bool* flushed)` reporting the cut.
- DML runs freely during the build; the published index is exactly consistent with the table.

### Delete-log with a `[begin, end)` rowid window
Deletes racing the backfill are the resurrection hazard (a native remove can be consumed by a mid-build refresh before the backfill copy commits). The log now covers only the shrinking window of rows whose copies may still be uncommitted:
- **end** — the store table's **rowid allocator** (`DataTable::GetNextRowId`) at build start. Using the allocator instead of the row count fixes a real resurrection bug: checkpoint vacuum drops fully-deleted row groups and persists rowid gaps, so live rowids can exceed `GetTotalRows` (reproduced end-to-end, regression test included).
- **begin** — a rising *minimum-uncommitted-rowid* watermark: one single-writer slot per sink (its current transaction's start), advanced only on that sink's commits, `begin = min(slots)` published on each cut. Lock-free: one `fetch_add` per thread, plain stores/loads, zero per-batch shared work. Deletes below `begin` or at/above `end` remove natively; only the window is logged (raw int64 rowids), replayed once at publish.

### Inverted-index options as first-class DDL
- New `WITH (...)` / session settings: `segment_memory_max`, `segment_docs_max`, `compaction_max_segments`, `compaction_max_segments_bytes`, `compaction_floor_segment_bytes`; per-index compaction tier built from them.
- Options always hold **concrete values** (struct field initializers are the defaults; CREATE resolves WITH → session); one shared validator for CREATE and ALTER (unknown names rejected, 0 rejected except where 0 is meaningful: `segment_docs_max` = unlimited, intervals = task disabled; uint32 range-checked). `WITH (row_group_size = 0)` used to abort the server.
- **`ALTER INDEX <name> SET/RESET (...)`** — new PEG grammar + transformer (duckdb submodule bump), catalog COW update, options applied live to the running storage (writer limits + task cadences).
- pg-native surfaces: `pg_class.reloptions` (full persisted option list), `pg_class.relam` → `pg_am`, `pg_index.indisvalid/indisready = false` while the build is in flight.

### Misc
- Secondary (ART) index attach plan uses an empty-result child (attach-first).
- `sdb_faults`-driven sqllogic tests live under `tests/sqllogic/recovery/` (per-file isolated servers); enforced by the fault-points pre-commit hook.
- Shared `catalog::GetStoreTableEntry()` helper replaces three hand-rolled facade→store lookups.
- Catalog persistence fixtures regenerated (options struct change) — old datadirs need a wipe.

### Tests
- `recovery/online_create_index_concurrent_dml.test` — parked build vs. real DML (delete/update/fresh-row/TRUNCATE), exact final state.
- `recovery/online_create_index_rowid_gap.test` — the vacuum-gap resurrection regression (checkpoint + parked build + mid-build delete).
- `recovery/online_create_index_rotation_parked.test` — committed segments visible mid-build.
- `sdb/pg/index/online_create_index_rotation.test` — fault-free rotation via `segment_docs_max`.
- `sdb/pg/index/inverted_index_options.test` — reloptions/pg_am/pg_index, ALTER SET/RESET round-trips, validator error matrix.
- Validated: full index suite green on Debug + Release (282/282), gtests green, plus a 15M-row stress run with concurrent insert/update/delete during a long build — exact table↔index consistency, zero resurrections.


server/connector/duckdb_physical_create_index.cpp
```diff
@@ -282,16 +287,47 @@ SereneDBPhysicalCreateIndex::GetGlobalSinkState(
...
+      }
+      return ResolveUbigintWithOption(context, name, nullptr);
     };
 
+    // Reject typos instead of silently ignoring them (CREATE TABLE already
+    // does); `_sdb_`-prefixed options are internal plumbing.
+    for (const auto& [option, value] : _info->options) {
+      const auto lower = duckdb::StringUtil::Lower(option);
+      if (lower.starts_with("_sdb_") ||
+          std::ranges::contains(kCreateInvertedOptions, lower)) {
```
MBkkt:
```suggestion
          absl::c_contains(kCreateInvertedOptions, lower)) {
```

server/connector/search_sink_writer.hpp
```diff
@@ -315,6 +316,13 @@ class DuckDBSearchSinkInsertWriter final : public DuckDBSinkIndexWriter,
     InitImpl(batch_size, pk);
   }
 
+  // `flushed` reports a segment flush performed by this batch's insert; on a
+  // commit-on-flush transaction that flush is also a commit (see
+  // irs::IndexWriter::Transaction::Insert).
+  void Init(duckdb::idx_t batch_size, const PkChunk& pk, bool* flushed) {
+    InitImpl(batch_size, pk, flushed);
+  }
+
```
MBkkt: why do we need this code at all? why can't we make initimpl public and not-impl instead?

server/pg/pg_catalog/pg_index.cpp
```diff
@@ -99,9 +99,9 @@ catalog::MaterializedData SystemTableSnapshot<PgIndex>::GetTableData() {
         .indisexclusion = false,
         .indimmediate = true,
         .indisclustered = false,
-        .indisvalid = true,
+        .indisvalid = !index.Tombstoned(),
         .indcheckxmin = false,
-        .indisready = true,
+        .indisready = !index.Tombstoned(),
```
MBkkt: maybe false?

server/connector/inverted_store_index.h
```diff
@@ -140,6 +145,8 @@ class InvertedStoreIndex final : public duckdb::BoundIndex {
 
   ObjectId _table_id;
   ObjectId _index_id;
+
+  std::shared_ptr<search::InvertedIndexStorage> _storage;
```
MBkkt: remove this?

server/connector/inverted_store_index.cpp
```diff
@@ -361,24 +362,60 @@ void InvertedStoreIndex::Delete(duckdb::IndexLock&, duckdb::DataChunk& chunk,
...
+    // Build in progress (rare path, so the copy is fine): the delete log
+    // covers the rowid window [begin, end). Below begin the backfill copy is
+    // already committed and above end the row was inserted during the build
+    // through the live writer -- in both cases the native remove masks the
+    // doc, so only the window in between (rows the scan may still copy
+    // uncommitted) defers to the log.
+    std::vector<int64_t> rows(count);
+    for (duckdb::idx_t i = 0; i < count; ++i) {
+      rows[i] = data[fmt.sel->get_index(i)];
+    }
```
MBkkt: memcpy if sel is identity

server/connector/inverted_store_index.cpp
```diff
@@ -361,24 +362,60 @@ void InvertedStoreIndex::Delete(duckdb::IndexLock&, duckdb::DataChunk& chunk,
...
+    const auto log_begin = _storage->DeleteLogRowidBegin();
+    const auto log_end = _storage->DeleteLogRowidEnd();
+    const auto native_end = std::partition(
```
MBkkt: absl::c_partition/std::ranges

server/connector/inverted_index_options_util.h
```diff
@@ -0,0 +1,126 @@
...
+// cleanup step) of 0 disables that background task -- the established idiom
+// deterministic tests rely on.
+inline constexpr bool IsZeroAllowedInvertedOption(std::string_view name) {
+  constexpr std::array<std::string_view, 4> kZeroAllowed = {
+    kSegmentDocsMaxSetting,
+    kRefreshIntervalSetting,
+    kCompactionIntervalSetting,
+    kCleanupIntervalStepSetting,
+  };
+  return std::ranges::contains(kZeroAllowed, name);
```
MBkkt:
```suggestion
  return absl::c_contains(kZeroAllowed, name);
```

server/connector/inverted_index_options_util.h
```diff
@@ -0,0 +1,126 @@
...
+  constexpr std::array<std::string_view, 7> kUint32Options = {
+    "row_group_size",
+    "norm_row_group_size",
+    kRefreshIntervalSetting,
+    kCompactionIntervalSetting,
+    kCleanupIntervalStepSetting,
+    kSegmentDocsMaxSetting,
+    kCompactionMaxSegmentsSetting,
+  };
+  return std::ranges::contains(kUint32Options, name);
```
MBkkt:
```suggestion
  return absl::c_contains(kUint32Options, name);
```

server/connector/duckdb_physical_create_index.cpp
```diff
@@ -619,6 +717,30 @@ duckdb::SinkFinalizeType SereneDBPhysicalCreateIndex::Finalize(
     }
 
     auto& inverted_storage = *gstate.index_storage;
+    auto delete_log = inverted_storage.TakeDeleteLog();
+    if (!delete_log.empty()) {
+      // Sorted rowids encode to lexicographically sorted pk terms, so the
+      // remove filter walks each segment's term dictionary sequentially.
+      std::ranges::sort(delete_log);
```
MBkkt:
```suggestion
      absl::c_sort(delete_log);
```

libs/iresearch/include/iresearch/index/index_writer.cpp
```diff
@@ -770,6 +785,25 @@ bool IndexWriter::FlushRequired(const SegmentWriter& segment) const noexcept {
...
+      meta.name = FileName(NextSegmentId());
+      return meta;
+    },
+    options);
+  // recreate writer if it reserved more memory than allowed by current limits
+  if (_segment_limits.Memory() < segment->writer->memory_reserved()) {
+    segment->writer = SegmentWriter::make(segment->dir, options);
+  }
+  return {std::move(segment), _segments_active};
+}
```
MBkkt: probably drop this

MBkkt (approved):

PR #932, opened 2026-07-18
feat: add multi-column search highlights

### LINKED ISSUE #918: ts_highlights(): per-matched-column snippets in one call
## Problem

A result renderer wants "show where this document matched". Today that requires one `ts_highlight(col)` call per searchable column plus client-side merging:

```sql
SELECT ts_highlight(title)        AS hl_title,
       ts_highlight(related_text) AS hl_related,
       ts_highlight(body)         AS hl_body
FROM idx WHERE title @@ $q OR related_text @@ $q OR body @@ $q;
```

Drawbacks: N calls per row; a column matched via a *different* OR branch returns its plain text, indistinguishable from "matched but no tags" without client-side inspection for the markup; apps end up reimplementing "which fields matched" logic that the index scan already knows.

## Proposal

```sql
SELECT ts_highlights(idx.tableoid) AS hl   -- MAP(VARCHAR, VARCHAR)
FROM idx WHERE title @@ $q OR body @@ $q;
-- {'title': 'FENDT <b>826</b> VARIO', 'body': '... <b>826</b> ...'}
```

Returns only the columns whose predicates produced hits for that row, each rendered with the usual options (optional second argument, same option string as `ts_highlight`).

## Why good-first-issue

The building blocks all exist and this is mostly composition: `ts_highlight` already has a bind-time sugar-rewrite mechanism (`RewriteToPostings` in `server/connector/functions/ts_highlight.cpp`) that synthesizes `ts_offsets` calls, `ts_offsets` works per column, and the scan knows the matched fields. Suggested shape: rewrite `ts_highlights(tableoid)` at bind time into per-matched-column `ts_highlight` calls assembled into a MAP, skipping columns whose offsets list is empty.

## Acceptance

- One call returns snippets for exactly the matched columns (empty map for pure verbatim/code matches without offsets).
- Works under OR across columns and with per-clause boosts.
- sqllogic coverage alongside `tests/sqllogic/sdb/pg/index/headline.test`.

w3lld1:
Fixes #918

## Summary
- add `ts_highlights(tableoid[, options])` to return highlighted snippets keyed by each matched indexed text column
- rewrite the function against the inverted-index scan so each field uses its own offsets
- cover OR queries, per-clause boosts, custom tags, and empty maps when no positional offsets are available

## Validation
- `uvx --from pre-commit pre-commit run clang-format --files server/connector/functions/search.h server/connector/functions/ts_highlight.cpp server/connector/functions/ts_highlight.h server/connector/optimizer/iresearch_plan.cpp`
- `git diff --check`
- Full `ninja -C build serened` was attempted, but the local execution window was spent building the project toolchain dependencies and did not reach the changed target. Full compile and sqllogic coverage are therefore deferred to CI.

### DIFF
--- a/server/connector/functions/search.h
+++ b/server/connector/functions/search.h
@@ -106,6 +106,7 @@ inline constexpr std::string_view kHasAnyTokens = "has_any_tokens";
 
 // Highlighting + position projections.
 inline constexpr std::string_view kTsHighlight = "ts_highlight";
+inline constexpr std::string_view kTsHighlights = "ts_highlights";
 inline constexpr std::string_view kOffsets = "ts_offsets";
 
 // Term-dictionary access -- optimizer-claimed aggregate stubs.
--- a/server/connector/functions/ts_highlight.cpp
+++ b/server/connector/functions/ts_highlight.cpp
@@ -32,6 +32,7 @@
 #include <duckdb/common/constants.hpp>
 #include <duckdb/common/vector/flat_vector.hpp>
 #include <duckdb/common/vector/list_vector.hpp>
+#include <duckdb/common/vector/map_vector.hpp>
 #include <duckdb/common/vector/string_vector.hpp>
 #include <duckdb/execution/expression_executor.hpp>
 #include <duckdb/execution/expression_executor_state.hpp>
@@ -53,6 +54,16 @@
 #include "pg/sql_exception_macro.h"
 
 namespace sdb::connector {
+
+duckdb::unique_ptr<duckdb::FunctionData> TsHighlightsBindData::Copy() const {
+  return duckdb::make_uniq<TsHighlightsBindData>(*this);
+}
+
+bool TsHighlightsBindData::Equals(const duckdb::FunctionData& other) const {
+  const auto& rhs = other.Cast<TsHighlightsBindData>();
+  return options == rhs.options && field_names == rhs.field_names;
+}
+
 namespace {
 
 struct UTextDeleter {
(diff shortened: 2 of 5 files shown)


MBkkt (commented):
@w3lld1 Thanks for the PR -- multi-column highlights are a welcome addition and we'd like to merge this. A few things first:

1. **CLA (blocking):** you need sign https://serenedb.com/cla (contact form on the page, we will answer you with CLA to sign).
2. **Test separator:** in `headline.test`, the first new `query` block uses a literal `\t` instead of real TAB characters between columns -- that's why the check fails. Replace the `\t` with actual tabs (compare the multi-column blocks in `basic.test`). The single-column blocks are fine.
3. **Build & run locally before pushing.** The build was "deferred to CI" — please compile and run the test yourself first:
   ```bash
   cmake --preset lldb -DCMAKE_C_COMPILER=clang-21 -DCMAKE_CXX_COMPILER=clang++-21
   cd build && ninja serened && cd ..
   ./tests/sqllogic/run.sh --single-port 7890 --test 'tests/sqllogic/sdb/pg/index/headline.test'
   ```
4. **Add an EXPLAIN test** for the `ts_highlights` rewrite -- the `iresearch_plan.cpp` change has no plan coverage. Mirror the existing `EXPLAIN SELECT ... ts_highlight(body) ...` tests in the same file and paste the real plan output; it should show `IRESEARCH_SCAN`.

If anything in CONTRIBUTING.md doesn't work for you, open an issue and we'll help. I merged `origin/main` into your branch, so please `git pull` first. Thanks!

MBkkt:
@w3lld1 Hi, if you sent as CLA request, we answered on both of your emails

PR #936, opened 2026-07-20
feat: Partial inverted indexes (CREATE INDEX ... WHERE <predicate>)

pashandor789:
## Summary

Adds partial inverted indexes: `CREATE INDEX ... USING inverted(...) WHERE <predicate>`. Only rows satisfying the predicate are indexed; scans against the index see exactly those rows.

The predicate is persisted as `catalog::ExpressionData` on `catalog::InvertedIndex` (an empty serialized expression means a full index, so there is no cost or behavior change for indexes without a `WHERE`). The predicate's dependent columns are folded into the index's referenced-column set, so both our DML maintenance and DuckDB's update-as-`DELETE`+`INSERT` path (triggered when an update touches an index column) see them.

## Details

- **Backfill:** the predicate is spliced into the build plan as a `LogicalFilter`, which the optimizer pushes down into the scan as a `Column Filter` (verified via `EXPLAIN`). This works for both table- and view-backed indexes; for views the filter is placed below the column-narrowing projection so it binds against the view's natural bindings.
- **DML maintenance:** `AppendRows` and `ReplayAppend` filter incoming rows through the predicate via a shared `CalcFilter` helper. Rows migrating across the predicate boundary are handled naturally by the `DELETE`+`INSERT` update path; non-matching rows are simply never indexed, so concurrent deletes of them are no-ops.
- **Errors:** a non-boolean predicate is rejected with `argument of WHERE must be type boolean` (matching PostgreSQL); partial indexes are rejected on non-inverted access methods.

## Testing

- `tests/sqllogic/sdb/pg/index/inverted_index_partial.test` — completed-index semantics across tables and views (rows crossing the predicate boundary in both directions, `NULL` handling, `EXPLAIN` assertions) and the error matrix.
- `tests/sqllogic/recovery/online_create_index_partial.test` — online build racing concurrent DML that crosses the predicate boundary; the published index is verified to exactly match the table's live rows.
- Full `pg/index` suite (both wire engines) and the existing online-build recovery tests pass.


server/connector/duckdb_catalog.cpp
```diff
@@ -1029,6 +1038,14 @@ duckdb::unique_ptr<duckdb::LogicalOperator> SereneDBCatalog::BindCreateIndex(
     }
   }
 
+  if (create_index_info->where_clause) {
+    if (create_index_info->index_type != "inverted") {
```
MBkkt:
```suggestion
  if (create_index_info->where_clause && create_index_info->index_type != "inverted") {
```

third_party/duckdb
MBkkt, line 1: something wrong

tests/sqllogic/sdb/pg/site_docs/compatibility/core_sql_claims.test
```diff
@@ -188,7 +188,8 @@ Did you mean "<slt:ignore>"?
 statement error
 CREATE INDEX ddl_index_part ON ddl_index_base (a) WHERE a > 5;
 ----
-db error: ERROR: Creating partial indexes is not supported currently
+db error: ERROR: Table with name ddl_index_base does not exist!
+Did you mean "<slt:ignore>"?
```
MBkkt: looks wrong

MBkkt (approved):

PR #941, opened 2026-07-20
fix: concurrent CREATE OR REPLACE SECRET write-write conflict (RTA s3_import flake)

pashandor789:
## Problem

The `make release` RTA intermittently fails (arm this run) on:

```
sdb/pg/site_docs/cookbook/network_cloud_storage/s3_import.test .. [FAILED]
    statement failed: db error: ERROR: Catalog write-write conflict on create with "__default_s3"
```

Uncovered once #940 fixed the RTA auth failure and let the suite run to completion.

## Root cause

An unscoped `CREATE SECRET (TYPE s3, ...)` maps to the single cluster-global name `__default_s3` (`transform_create_secret.cpp`). The RTA runs `--jobs 32` against one shared server, and ~8 sibling `network_cloud_storage` doc tests all `CREATE OR REPLACE SECRET (TYPE s3, ...)`. When two run concurrently, both target `__default_s3`; the second hits a write-write conflict in `CatalogSet::CreateEntry` → `VerifyVacancy` (`catalog_set.cpp:120`) and fails. Non-deterministic → arch-dependent flake.

## Fix (in the duckdb fork)

`serenedb/duckdb@1cbaf4c1a8` — `CatalogSetSecretStorage::StoreSecret` now catches the `TransactionException` from `CreateEntry` and, for the **idempotent** conflict modes (`IGNORE_ON_CONFLICT` / `REPLACE_ON_CONFLICT`), defers to the racing writer's entry instead of surfacing the conflict. `ERROR_ON_CONFLICT` (plain `CREATE SECRET`) still propagates.

Safe because the conflict is thrown in `VerifyVacancy` **before** any catalog mutation or undo-buffer push (`CreateEntryInternal` ordering), so the transaction has made no partial change — the post-catch re-read runs on an intact snapshot, and falls back to re-throw if the winner isn't yet visible.

This PR bumps the superproject's duckdb pin to that commit.

## Validation

Single-TU syntax-check of the patched `secret_storage.cpp` passes with the release build flags. Correctness rests on the transaction-state analysis above (no mutation applied before the throw).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


pashandor789:
Superseded by #952 — same fix, recommitted cleanly on newest main over the pinned duckdb SHA (435f37c8f3) with a fresh single commit.

PR #950, opened 2026-07-21
fix: Adaptive centroids fixes

### LINKED ISSUE #903: perf: Adaptive centroids structure
IVF has adaptive centroids structures:
- `BruteForce` -- no centroids, each search is full-scan (`#vectors` < 64)
-  `Flat` -- one level of centroids (disabled)
-  `TwoLevel` -- two level of centroids
`Flat` is disabled, because `TwoLevel` struct outperforms it. We need to benchmark different amount of centroids and find the most effective bounds for these three structs.
### Research
Notably, we might use different centroids layouts, like:
- In-memory HNSW (~`O(sqrt(N))` RAM)
- sptag (https://github.com/microsoft/sptag)
- do reality-check with RCQ implementation from spann (https://arxiv.org/abs/2111.08566)

codeworse:
- some followups from adaptive centroids PR
- set SQ8 as a default quantizer
- fix reranking on streaming ann search (previously, there were no reranking with original vectors, so the results were slightly worse)
- more tests on vector search


tests/sqllogic/sdb/pg/index/inverted_index_ann_include.test
```diff
@@ -93,12 +88,7 @@ QUERY PLAN
 ╭─ IRESEARCH_SCAN ───┴───────────────────╮
 │ Index: vec_t_idx                       │
 │ Lookup: table                          │
-│ Index Filter:                          │
-│ ╭─ Vector KNN ─╮                       │
-│ │ Field: emb   │                       │
-│ │ Metric: l2   │                       │
-│ │ Dims: 3      │                       │
-│ ╰──────────────╯                       │
+│ Score: l2_distance(emb)                │
```
MBkkt:
```suggestion
│ Score: l2_distance(emb, FLOAT[3])                │
```

MBkkt (approved):

PR #954, opened 2026-07-22
feat: Row-Level Security

pashandor789:
PostgreSQL-compatible Row-Level Security, carved RLS-only out of the combined rbac-rls-wip branch and rebased onto current main. The RBAC infrastructure it builds on is already merged.

## Enforcement is a serened optimizer pass

`server/connector/optimizer/rls.cpp` (580 lines) is registered as an `OptimizerExtension` `pre_optimize_function`, so it runs before `RunBuiltInOptimizers()`. That placement is the whole design:

- Rewriting the plan **before** filter pushdown means the predicates it emits are still pushed into the scans, so row-group pruning is preserved. Enforcing after the optimizer would have cost that.
- It maps each `LogicalGet.table_index` back to a table and a definer through `StatementProperties::access_requirements`, which the binder already populates for RBAC. No new binder plumbing.

Two mechanisms:

- **Read (`USING`)** — every governed scan is wrapped in a `LogicalFilter` carrying the combined policy predicate.
- **Write (`WITH CHECK`)** — INSERT/UPDATE/MERGE get a `BoundCheckConstraint` with `is_rls` set, reusing the existing constraint verifier and reporting `new row violates row-level security policy for table "..."`.

Combination follows PostgreSQL: PERMISSIVE predicates OR together, RESTRICTIVE ones AND on top, and a table with RLS enabled but no applicable policy denies every row. Superuser, `BYPASSRLS` and the table owner bypass, the owner only when `FORCE ROW LEVEL SECURITY` is off. Inside a `SECURITY DEFINER` view the policies of the view owner apply, not the caller's.

`TRUNCATE` on a table with applicable policies is refused rather than silently emptying it — a deliberate, documented divergence from PostgreSQL, which has no `TRUNCATE` policy concept.

## Catalog

- `Policy` + `RowSecurity` side objects (`policy.{h,cpp}`, `persistence/policy.h`), modeled on `SecondaryIndex`; `ObjectType::Policy` / `RowSecurity`. Deliberately kept off `TableData` — adding a field to a `WriteTuple` struct bricks existing datadirs.
- `CreatePolicy` / `AlterPolicy` / `DropPolicy` / `SetRowSecurity`, owner-only, checked under `_mutex` on the mutation snapshot.
- `PolicyIds` / `GetRowSecurity` accessors; load and drop-cascade wiring; a policy's `TO` roles register `RoleDependency` edges so `DROP ROLE` is correctly refused.

## DDL and introspection

`CREATE`/`ALTER`/`DROP POLICY` and `ALTER TABLE ... {ENABLE,DISABLE,FORCE,NO FORCE} ROW LEVEL SECURITY` route through pragma handlers into the catalog commands. `pg_policy`, `pg_policies` and `pg_class.relrowsecurity`/`relforcerowsecurity` are populated; `polqual` / `polwithcheck` render as text through `pg_get_expr`.

## Fork dependency

`third_party/duckdb` is bumped to serenedb/duckdb#57, which is +279/−3: the POLICY grammar plus a 13-line `is_rls` flag so a `WITH CHECK` violation reports the RLS error instead of leaking the policy predicate through the generic CHECK message. The grammar was lifted out of the shared RBAC parser commit by serenedb/duckdb#58 so it travels with this feature — on `main` today `CREATE POLICY` parses and then fails with `serenedb_create_policy does not exist`.

No binder hooks. An earlier revision of this branch added `rls_wrap_scan` / `rls_append_check_constraints` config hooks and ~100 lines across five binder files; the optimizer-pass design removed all of it, and those five files are now byte-identical to upstream.

## Tests

23 files: 20 under `sdb/pg/rls/`, 2 oracle-verified under `any/pg/rls/` (byte-identical output against postgres:18.3), plus `sdb/pg/rbac/rls_enforcement.test`. 44/44 green across both `pg-wire-simple` and `pg-wire-extended`; 552/552 RBAC tests still pass.

Regression tests accompany each bug found during review, including: `WITH CHECK` on a non-boolean expression (was a crash), write-path default-deny, `DELETE` not being filtered at all when the facade lookup returned the store entry, `CREATE POLICY` on a view or sequence (was a `SIGABRT`), a NULL `WITH CHECK` result failing open, and `predicate_ordering_canary.test`, which pins the `pushdown_get` ordering behaviour that keeps RLS predicates from being evaluated after a user-supplied throwing filter.

## Known gaps

Deferred, tracked, and not blocking this branch: the pre-image filter is verb-blind (a `SELECT`-only policy also gates `UPDATE`'s read), `UPDATE` has no post-image default-deny, and `SELECT` does not fold the post-image. Cached plans are not invalidated on `SET ROLE`, so a prepared statement can outlive the role that planned it.


server/pg/pg_catalog/fwd.h
```diff
@@ -59,7 +59,10 @@ struct Name {
 };
 
 struct Empty {};
-using PgNodeTree = Empty;
+struct PgNodeTree {
+  std::string_view v;
+  bool is_null = true;
```
MBkkt: inspired by kaldb, we have nullmask!

PR #966, opened 2026-07-23
feat: derive pg_depend/pg_rewrite/pg_attrdef from the dependency graph

pashandor789:
## What

`pg_depend`, `pg_rewrite` and `pg_attrdef` were registered but empty stubs, so
the ~15 `information_schema` / `pg_catalog` views that join them silently
degraded, and there was no way to introspect the catalog's dependency graph.
This derives all three from the in-memory dependency graph and makes the DROP
cascade walk declarative.

## pg_depend edges (validated byte-for-byte against PostgreSQL 18)

- index -> table (per indexed column, `a`)
- object -> schema/namespace (`n`)
- table column -> user type (`n`)
- FK constraint -> referenced table (`n`)
- view -> table/view/sequence/function, via the view's `_RETURN` rewrite rule
  in `pg_rewrite` (`n`), plus rule -> view (`i`)
- column default -> sequence/function/type, via `pg_attrdef` (`n`), plus
  attrdef -> column (`a`)

`pg_rewrite` and `pg_attrdef` are now populated (one `_RETURN` rule per view;
one row per defaulted column) so those joins resolve.

## Cascade registry (refactor)

Each `*Dependency` node is now pure data: its reverse-edge buckets plus a
`static constexpr std::array kEdges` of `{bucket-pointer, CascadeVerb}`.
`Clone()` and the `kEdges`-driven `Emit()` are generated by a `DependencyMixin`
CRTP base, and every verb routes through one exhaustive `DropEmitter::Apply`
switch (no `default:`). This removes the per-struct boilerplate and makes the
"forgot to walk a bucket -> silently lost cascade delete" bug impossible: you
cannot add a bucket without a `kEdges` row, and a new verb will not compile
until `Apply` handles it.

The RESTRICT dependent-objects check (previously copy-pasted across five DROP
handlers) is unified into `Snapshot::ComputeDropPlanRestrict`.

## Fix

Dropped a spurious `pg_depend` namespace edge emitted for hidden owned (rowid)
sequences -- they live in `owned_sequences`, are excluded from `pg_class`, and
so must not get a namespace edge either.

## Tests

Four pg-validated sqllogic tests (pass against real PostgreSQL 18 and serenedb
on both wire protocols): `pg_depend.test`, `pg_depend_constraints.test`,
`pg_depend_view.test`, `drop_cascade_foreign_key.test`.

## Known gaps (serenedb feature work, not rendering)

`pg_depend` still omits composite rowtypes / array types, per-constraint
auto/internal rows (PK/UNIQUE/NOT-NULL/CHECK), FK enforcement triggers,
view-column -> type, and per-column view-rule subids. These need modeling
serenedb does not have (rowtypes, `pg_trigger`, NOT-NULL constraint objects,
column-level view dependency tracking); DROP cascade behavior is unaffected.


server/catalog/object_dependency.h
```diff
@@ -42,6 +43,28 @@ namespace sdb::catalog {
 
 struct Snapshot;
 
+enum class DependType : char { Normal = 'n', Auto = 'a', Internal = 'i' };
```
MBkkt, line 46:
```suggestion
enum class DependType : char { Normal = 'n', Auto = 'a', Internal = 'i', };
```

MBkkt (approved):

server/pg/pg_catalog/pg_rewrite.h
```diff
@@ -55,4 +55,13 @@ struct PgRewrite {
 };
 // NOLINTEND
 
+inline constexpr uint64_t kViewRuleBit = uint64_t{1} << 60;
```
MBkkt, line 58: needs to remove this

PR #974, opened 2026-07-25
ci: pre-commit check for third_party submodule gitlinks

pashandor789:
New `check-submodule-pointers` pre-commit hook (runs in the existing pre-commit PR check): a bumped `third_party/duckdb` gitlink must

1. **descend from origin/main's gitlink** — no rollbacks and no stale-lineage rebases (the original #936 bump sat on the v2026.07.03 lineage and silently reverted v2026.07.05 fixes; this rule catches exactly that), and
2. **be reachable from a fork version branch** (`v2026.MM.DD`) — i.e. the fork-side PR is merged, not a feature-branch head.

Validation prefers the local submodule clone (pure git, offline-friendly); in CI — where checkout runs with `submodules: false` / `fetch-depth: 1` — it falls back to `git ls-remote` + the GitHub compare API (fork is public, unauthenticated works; `GITHUB_TOKEN` honored when set). Fails closed when it cannot validate; `SKIP=check-submodule-pointers` is the escape hatch. `always_run` because pre-commit filters submodule gitlinks out of its file lists, and the script no-ops in two local git calls when the pointer equals main's.

Tested: current #936 pointer passes (on v2026.07.05, descends from main); the original broken pointer `18861c413c` fails rule 1; an unmerged feature-branch head fails rule 2; a dangling sha is reported as unknown-to-remote; both the local-git and API paths exercised.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -115,3 +115,13 @@ repos:
         language: python
         entry: python3 scripts/check_api_keys.py
         exclude_types: [binary]
+      - id: check-submodule-pointers
+        name: Check submodule gitlinks (on a version branch, not older than main)
+        language: python
+        entry: python3 scripts/check_submodule_pointers.py
+        # pre-commit filters submodule gitlinks out of its file lists, so a
+        # files: pattern on third_party/duckdb never matches; the script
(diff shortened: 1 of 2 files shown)


pashandor789:
Folding into #936 per review — everything lands in one PR.

Issue #796, opened 2026-06-12
Repository restructure: iresearch/ + src/ layout, naming rules, dependency boundaries


MBkkt:
## Top level

```
/
  iresearch/     standalone search/storage library (own README, LICENSE)
  src/           the database
  third_party/   all external deps, including patched duckdb -- shared
  scripts/       shared, single location for everything
  resources/     shared, single location for everything
  tests/         single tree; first segment names the component:
                 tests/iresearch/, tests/catalog/, tests/query/, tests/net/, ...
                 plus tests/sqllogic/, tests/drivers/
  bench/         same pattern: bench/iresearch/, bench/query/, ...
  examples/      examples/iresearch/ + database examples
  docs/ cmake/ packages/ serene-ui/ etc/
```

**Dependency rule (CI-enforced include check):** `iresearch -> third_party` only. Inside src: `utils <- catalog <- storage <- query <- net <- app`; every src dir may use `iresearch/` and `third_party/`. CI gates iresearch jobs on `iresearch/** + third_party/** + tests/iresearch/**` -- a pure src PR skips them, and the iresearch build is a cacheable unit keyed on those trees.

## iresearch/

```
iresearch/
  utils/       structured: containers/, geo/, simd/, encoding/, ...
  file/        directory/IO abstraction
  formats/     the on-disk format (.col/.idx/.doc): column/, index/, postings/
  index/       lifecycle over the format: write, read, compact/merge, refresh
  search/      filter/ (definitions) - optimize/ (filter optimizer) - prepare/ (index/column
               lookup) - execute/ (posting iteration, cost) - score/ (bm25, dfi, ...)
```

Analysis/tokenization is not part of the library -- it becomes query functions in src. Internal flattening (killing `include/iresearch/`) happens with this move; namespace is flat `irs`.

## src/

```
src/
  utils/       structured like iresearch/utils; zero deps on other src dirs:
               log, crash handler, message buffers, coro helpers, dtoa, file utils,
               pg/ (shared pg value/text/binary serde, if wire and COPY/functions both need it)
  catalog/
    (root)     object model, identifiers/ticks, options parsing, system framework
    store/     persistence serde + backends; rocksdb dies in place; engine-file next; clickeeper/ later
    bridge/    the engine's view of an immutable catalog snapshot: schema/table/index entries,
               transaction hooks, client state, entry cache, storage extension, logical-type mapping
    pg/        pg_catalog relations
    information_schema/
[...]

PR #968, opened 2026-07-23
feat: support build on mac os

mkornaukhov:

PR #976, opened 2026-07-25
build: add an embeddable IResearch-only target

ivan-digital:
## Summary

- add an opt-in `SDB_BUILD_IRESEARCH_ONLY` configuration
- expose `iresearch::embedded` without building the SereneDB server
- omit server adapters, geo, FastText, connectors, and unrelated third-party targets
- add a consumer smoke target covering in-memory BM25 and IVF centroid train/serialize/search

## Scope

The target leaves IResearch source behavior unchanged. It retains the current
DuckDB core, vendored ICU, and FAISS dependencies, but skips DuckDB's
`core_functions` and Parquet extensions. This PR does not extract IResearch
from DuckDB types or its current vector interfaces.

Linux uses SereneDB's existing project-built libc++ runtime. Platform-specific
macOS compatibility remains on the downstream
[`ivan/stenograf-iresearch-macos`](https://github.com/ivan-digital/serenedb/tree/ivan/stenograf-iresearch-macos)
branch because the pinned DuckDB fork deliberately rejects macOS.

## Validation

- built `iresearch-embedded-smoke` in the official ARM64 Linux build image with Clang 21, lld, and `SDB_BUILD_IRESEARCH_ONLY=ON`
- ran the Linux smoke executable: `bm25=pass ivf=pass`
- built and ran the same smoke target with Apple Clang 21 on the downstream macOS branch: `bm25=pass ivf=pass`
- repository security check passes

## Draft status

Ready for maintainer feedback on the target boundary, naming, and opt-in build
configuration. Windows remains unvalidated and is not claimed as supported by
this PR.

Related to #796.


MBkkt:
Hi @ivan-digital 

We have PR with macos support, so it will be fixed soon https://github.com/serenedb/serenedb/pull/968

Overall I think it's not bad idea, to have some build only iresearch flag.
I maybe want to restructure iresearch before this, as mentioned in the issue, but overall it sounds quite ok.

One thing that I noticed, why did you disabled geo and fasttext? Only because you don't need them and they're additional deps that easy to disable? Or some other reason? (e.g. some issues with build/etc)

ivan-digital:
Yes, exactly—this was only scope and dependency reduction, not a build issue. The target is a small embedded hybrid index: lexical search plus externally produced dense vectors. Geo and FastText analyzers are outside that scope, so I left their dependencies out. We can make them explicit options if the standalone target should preserve all features by default.

ivan-digital:
Windows is also an intended target for this embedded setup. IResearch already has Windows-specific build paths, so it looks feasible, but I have not validated the complete build and smoke test there yet.

MBkkt:
About macos/windows support, we plan to remove iresearch specific file utils and use duckdb/patched-duckdb file utils. Because  duckdb supports more fs types (s3, abfs) + better code coverage, as bonus it will supports windows and macos.

But I think it will takes some time, I think we planned to do this during August

PR #979, opened 2026-07-27
FDW follow-ups: lookup transport via in-out chunks, plus RBAC/catalog cleanups

pashandor789:
Follow-ups from the FDW review (#871): the remote-key lookup transport reworked onto duckdb's existing mechanisms, plus four independent catalog/RBAC cleanups.

Depends on two fork PRs — serenedb/duckdb#55 and serenedb/duckdb-postgres#5. Both are open, so the gitlinks here point at branch heads and `check-submodule-pointers` fails until they merge; the last commit exists to be re-pointed at the merged tips.

## The lookup transport

`TableFunctionInput` grew four serenedb fields plus ~20 lines of contract in #871:

```cpp
optional_ptr<Vector> lookup_keys;
idx_t lookup_count = 0;
bool (*lookup_gate)(void *state, int64_t value) = nullptr;
void *lookup_gate_state = nullptr;
```

Parquet — the mechanism this was meant to mirror — uses none of it. It has `pk_lookups` + `pk_survivors` (both predating #871) and writes the mapping itself: `data.pk_survivors[w] = pk_idx`. The remote path handed that same integer back out through a `void*` callback so the *caller* could write it. That gate was the **only per-row callback in the materializer layer** — `RowIdFetchIndexSource` passes `idx_t *out_survivor_idx`, the file sources set `in.pk_survivors`, nothing else uses one.

Keys now travel as the `in_out_function` **input chunk**, which is what that hook is for, and `OperatorResultType` supplies the continuation protocol the deleted comment spelled out in prose (`HAVE_MORE_OUTPUT` / `NEED_MORE_INPUT`). So:

- **all four fields are gone** — `TableFunctionInput` is back to its pre-#871 shape;
- `postgres_lookup` / `clickhouse_lookup` are real table functions like `MakeParquetLookupTableFunction()`, so `lookup := true` and the `if (bind_data.lookup)` branch in both production scanners disappear;
- each scan fills `pk_survivors` directly and dedups ordinals in its global state, so a duplicate remote row cannot claim a second slot;
- the `GetType().GetAlias() == "ctid"` type-alias sniff is gone — the caller knows the key shape (it built the `PkSpec`), so `BorrowKeyColumns` states it, and column count becomes the request arity, which `postgres_lookup` cross-checks against its bound parameter count.

`BorrowKeyColumns` follows `SortRows`: a reused member chunk whose columns are `Vector::Ref` of the pk's, so nothing is copied and only the first batch allocates.

## Catalog / RBAC cleanups

**`fix(clickhouse): declare the catalog as remote`** — `ClickHouseCatalog` never overrode `Catalog::Supports`, so it inherited `return false` and reported itself *local* while `PostgresCatalog` declares `IS_REMOTE`. Consumers treated it as a local database: `CheckAmbiguousCatalogOrSchema` did a pointless schema lookup and `DatabaseManager` left it out of `remote_catalog_count`. Declares `IS_REMOTE` only — `EXECUTE_QUERY_NODE` would hand plans to duckdb's `RemotePushdownOptimizer` when CH pushdown goes through the shared dbconnector optimizer, and `CONNECT` obliges a `RemoteExecute(string)` that doesn't exist. Every rewrite site in `RemotePushdownOptimizer` gates on `EXECUTE_QUERY_NODE`, so the newly-enabled pass cannot change CH plans.

**`refactor(rbac): split CollectAndEnforce up and skip local catalogs early`** — ~145 lines with five jobs inlined → `RequireForeignServerUsage`, `CollectRelations`, `MapStoreOrphansToFacades`, `CollectWriteTargets`; the verb→ACL switch became a table, dropping a function and its silent `default: NoRights`. Also: `GetIdentifierName()` returns `const string&`, so `const auto` was copying a `std::string` per relation (now a borrowed `string_view`), and the check consults `IS_REMOTE` before anything else because `GetForeignServer(name)` is the instance-wide overload that builds a vector of every database and probes each — a plain query used to run that scan per distinct alias only to conclude "not a foreign server".

**`fix(pg_catalog): drop the synthetic oid bits`** — `pg_rewrite` tagged its oid with bit 60 and `pg_attrdef` with bit 59, so a view's `_RETURN` rule reported oid `1152921504608847329`. PostgreSQL's oid is unsigned 32-bit (max 4294967295); these only fit because serenedb declares oid as BIGINT, so any client decoding oid as 32 bits or casting to int4 breaks on exactly these two catalogs. The tags were **write-only** — nothing masked or tested them — the ids already come from the one instance-wide `ObjectId` allocator, and `pg_depend` namespaces `objid` by `classid`, which is how PostgreSQL disambiguates. Removing them emptied the `switch` in `pg_depend` that existed only to apply them. Verified every `pg_depend` row with classid `pg_rewrite`/`pg_attrdef` still joins its catalog on oid with no cross-matches. One deliberate divergence: a rule's oid now equals its view's oid, where PostgreSQL allocates a distinct number.

**`refactor(catalog): drop the PgSql prefix from ObjectType enumerators`** — `PgSqlFunction`/`PgSqlType`/`PgSqlView` were the only prefixed values; `Table`, `Sequence`, `ForeignServer` already read as the bare kind. Enum **values are untouched**, so datadirs are unaffected, and no serialized or user-visible string carried the names (`ToPgObjectTypeName` still returns `"view"`/`"function"`/`"type"`). Classes keep their names. `sql_utils.h` pulls the enumerators in with `using enum` and names them bare, so this injects `Type`/`Function` into that scope — compiles clean.

## Tested

Full sqllogic suite **2264/0**; every lookup suite green on **both** wire engines — `lookup_composite` (multi-column keys), `lookup_nulls`, `lookup_edge`, `lookup_regression` (dedup + compaction), and the ctid path via `lookup_nulls_pgscan` / `inverted_index_pgscan`. Unit tests 569 basics + 489 connector; vendored extension suite 3550/0.

## Worth flagging

The `IS_REMOTE` short-circuit makes ClickHouse `GRANT USAGE ON FOREIGN SERVER` enforcement depend on the capability declaration, and the two live in different trees. During development I reverted one without the other and denied queries silently succeeded — the FDW tests caught it, but the gate fails *open*. A follow-up should key the short-circuit on "does this snapshot hold any foreign server at all", which is fail-closed and just as cheap.


server/connector/optimizer/rbac.cpp
```diff
@@ -138,53 +127,75 @@ ObjectId EffectiveRole(ObjectId caller, const duckdb::CatalogEntry* who) {
...
+// DML records the plan's scan table -- the store-side entry, not the user
+// facade. Map store entries back through the facades bound alongside them
+// (forward-composed store name; never parsed).
+void MapStoreOrphansToFacades(const AccessRequirements& reqs,
+                              std::vector<const catalog::Object*>& objects) {
+  containers::FlatHashMap<std::string, const catalog::Object*> facades;
+  for (size_t i = 0; i < reqs.size(); ++i) {
+    const auto* entry = reqs[i].table;
+    if (!objects[i] || !entry ||
+        !dynamic_cast<const connector::SereneDBTableEntry*>(entry)) {
```
gnusi: dynamic_cast - can remove?

server/connector/optimizer/rbac.cpp
```diff
@@ -138,53 +127,75 @@ ObjectId EffectiveRole(ObjectId caller, const duckdb::CatalogEntry* who) {
...
         ERR_CODE(ERRCODE_INSUFFICIENT_PRIVILEGE),
         ERR_MSG("permission denied for foreign server ", server->GetName()));
     }
   }
+}
+
+// DML records the plan's scan table -- the store-side entry, not the user
+// facade. Map store entries back through the facades bound alongside them
+// (forward-composed store name; never parsed).
+void MapStoreOrphansToFacades(const AccessRequirements& reqs,
```
gnusi: is it from kaldb?

MBkkt (approved):

PR #980, opened 2026-07-27
Drive the external lookup like the file lookups (in-out transport, no callback)

pashandor789:
Reworks the remote-key lookup added in #871 so it uses duckdb's existing mechanisms instead of its own. Depends on two fork PRs (see below).

### The problem

`TableFunctionInput` grew four serenedb fields plus ~20 lines of contract:

```cpp
optional_ptr<Vector> lookup_keys;
idx_t lookup_count = 0;
bool (*lookup_gate)(void *state, int64_t value) = nullptr;
void *lookup_gate_state = nullptr;
```

Parquet — the mechanism this was meant to mirror — uses none of it. It has `pk_lookups` + `pk_survivors` (both predating #871) and writes the survivor mapping itself: `data.pk_survivors[w] = pk_idx`. The remote path instead handed that same integer back out through a `void*` callback so the *caller* could write it. That gate was the **only per-row callback in the whole materializer layer** — `RowIdFetchIndexSource` passes `idx_t *out_survivor_idx`, the file sources set `in.pk_survivors`, nothing else uses a callback.

### The change

Keys travel as the `in_out_function` **input chunk**, which is what that hook is for, and `OperatorResultType` supplies the continuation protocol the deleted comment described in prose (`HAVE_MORE_OUTPUT` / `NEED_MORE_INPUT`). So:

- **all four fields are gone** — `TableFunctionInput` is back to its pre-#871 shape;
- `postgres_lookup` / `clickhouse_lookup` are real table functions, like `MakeParquetLookupTableFunction()`, so `lookup := true` and the `if (bind_data.lookup)` branch in both production scanners disappear;
- the scan fills `pk_survivors` directly and dedups ordinals in its global state;
- the `GetType().GetAlias() == "ctid"` type-alias sniff is gone — the caller knows the key shape (it built the `PkSpec`), so `BorrowKeyColumns` states it and the column count becomes the request arity, which `postgres_lookup` cross-checks against its bound parameter count.

`BorrowKeyColumns` follows `SortRows`: a reused member chunk whose columns are `Vector::Ref` of the pk's, so nothing is copied and only the first batch allocates.

### Depends on

- serenedb/duckdb#55 — removes the `lookup_keys` field
- serenedb/duckdb-postgres#5 — adds `postgres_lookup`

Both are open, so the gitlinks here point at branch heads and `check-submodule-pointers` will fail until they merge. The last commit exists to be re-pointed at the merged tips.

### Tested

Full sqllogic suite **2264/0**; every lookup suite green on **both** wire engines — `lookup_composite` (multi-column keys), `lookup_nulls`, `lookup_edge`, `lookup_regression` (dedup + compaction), and the ctid path via `lookup_nulls_pgscan` / `inverted_index_pgscan`. Vendored extension suite 3550/0.

### DIFF
--- a/server/connector/index_source_external_lookup.cpp
+++ b/server/connector/index_source_external_lookup.cpp
@@ -35,8 +35,10 @@
 #include <duckdb/common/vector/struct_vector.hpp>
 #include <duckdb/common/vector_operations/vector_operations.hpp>
 #include <duckdb/common/vector_size.hpp>
+#include <duckdb/execution/execution_context.hpp>
 #include <duckdb/main/client_context.hpp>
 #include <duckdb/main/connection.hpp>
+#include <duckdb/parallel/thread_context.hpp>
 #include <duckdb/parser/keyword_helper.hpp>
 #include <duckdb/parser/tableref/table_function_ref.hpp>
 #include <string>
@@ -120,7 +122,6 @@ ExternalLookupIndexSource::ExternalLookupIndexSource(
 
   BuildQuery(context, ref, select_names);
 
-  _filled.reserve(STANDARD_VECTOR_SIZE);
   _sort_perm.resize(STANDARD_VECTOR_SIZE);
   absl::c_iota(_sort_perm, duckdb::idx_t{0});
 }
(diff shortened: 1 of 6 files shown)


pashandor789:
Folded into #979 so the whole change set reviews as one PR — same seven commits, same 32 files.

PR #982, opened 2026-07-27
fix: Force split for similar vectors to end eventually

Dronplane:
Similar or same vectors may force index to run BuildAndSplit eternally.
Now if new group have same size as its parent - split is forced to happen by winning centroid


libs/iresearch/include/iresearch/formats/ivf/centroids.cpp
```diff
@@ -229,6 +229,24 @@ void Build(std::vector<CentroidsBuilder::Node>& nodes, std::span<float> data,
...
+    ForEachGroup(entry.ids, n_built, [&](size_t g, size_t, size_t count) {
+      if (count == sample_size) {
+        full_group = g;
+      }
+    });
+    if (full_group != n_built) {
+      const std::vector<float> winner(centroids.begin() + full_group * d,
+                                      centroids.begin() + (full_group + 1) * d);
+      for (size_t g = 0; g < n_built; ++g) {
+        std::copy_n(winner.begin(), d, centroids.begin() + g * d);
```
codeworse: `absl::c_copy_n(...)`?

codeworse (approved):

libs/iresearch/include/iresearch/formats/ivf/centroids.cpp
MBkkt: looks slow

Dronplane: At least this is for quite rare case - when we have not found acceptable split

Dronplane: even simpler - absl::c_copy fits here

PR #983, opened 2026-07-27
test: cover distinct aggregate input dedup and shared distinct radix tables

gnusi:
Bumps `third_party/duckdb` to pick up serenedb/duckdb#56, *"Scan a shared distinct radix table once and dedupe distinct aggregate inputs"*, and extends the pre-aggregation dedup test to cover the new behaviour.

## What changes in the engine

Distinct aggregates with equal inputs and equal filters already shared one distinct radix table, but the finalize scanned that table once **per aggregate** with a destroying scan, so the second reader found it drained. Because the pre-aggregation projection never shared a slot between distinct aggregates, that mapping was always 1:1 and the path was unreachable — distinct inputs were excluded from the input dedup for exactly that reason.

The submodule change makes the finalize scan a shared table once and feed every aggregate that shares it, then drops the exclusion. So `count(DISTINCT x), sum(DISTINCT x)` builds and scans a single distinct set, and their input collapses to one projection slot.

## Test changes

`tests/sqllogic/sdb/pg/explain/aggregate_input_dedup.test`:

- The header asserted *"Distinct aggregates keep their own input column"*, which is no longer true; it now describes the sharing.
- New result coverage: five distinct aggregates over one input; distinct aggregates over **different** inputs; distinct aggregates sharing an input **and** a filter; distinct aggregates over one input with **different** filters (deliberately not merged, since matching requires equal filters); the grouped variant; and `GROUPING SETS` over shared distinct inputs.
- Three new pinned plans:
  - distinct aggregates sharing a slot — `count(DISTINCT #0), sum_no_overflow(DISTINCT #0), min(#0)`, with the pre-aggregation projection elided entirely;
  - distinct aggregates over different inputs — a slot each (`#0`, `#1`);
  - same input with differing filters — input shared on `#0`, filters on separate slots `#1` / `#2`.

No existing expectations needed updating: the `ts_dict_*` plans and the rest of the explain subtree are unchanged by this bump.

## Validation

- Differential against stock DuckDB 1.5.5: **180/180 identical** across 30 query shapes × 6 datasets, run twice (before and after a tuning pass).
- `TASK_BLOCKED` resume path covered via fault injection: **13,136** forced resume cycles across both aggregate operators at 1–5 sharers, differential still 180/180.
- ThreadSanitizer clean; Debug assertions hold.
- `sdb/pg/explain/aggregate_input_dedup.test` passes without `--override` after pinning.

Aggregate conformance suite: **160/160 green, 0 failures** (`sdb/pg/any/sqlite/aggregates` 130 files, plus the `ts_dict_*` suite and the whole `explain` subtree) on the final build of this change.

## Performance

Release build, 50M rows, 32 cores, both servers live and alternating on every repetition, min of 15:

| query | before | after | delta |
| --- | --- | --- | --- |
| `count(DISTINCT x), sum(DISTINCT x)` | 25.72 ms | 21.16 ms | −17.7% |
| 4 distinct aggregates, same input | 26.19 ms | 21.57 ms | −17.7% |
| grouped, 2 distinct, same input | 394.27 ms | 221.13 ms | **−43.9%** |
| grouped, 4 distinct, same input | 414.75 ms | 244.89 ms | **−41.0%** |
| 1 distinct aggregate | 12.16 ms | 11.96 ms | −1.7% |
| 1 distinct aggregate, grouped | 198.49 ms | 198.33 ms | −0.1% |
| 2 distinct aggregates, different inputs | 26.06 ms | 25.45 ms | −2.3% |
| 3 distinct aggregates, different inputs | 40.23 ms | 38.83 ms | −3.5% |
| no distinct aggregates | 1.20 ms | 1.20 ms | +0.2% |
| no distinct aggregates, grouped | 18.92 ms | 19.04 ms | +0.6% |

Grouped shared cases nearly halve: two distinct radix tables were built and scanned where one now suffices. Queries with no shared distinct inputs are unaffected (worst case +0.6%, most slightly faster), including those that take the same new code path with a single reader per table.

`ts_dict` plans and timings are untouched by this bump: TsDict aggregate pushdown produces plain aggregates over already-distinct dictionary terms, so it never had shared distinct tables to begin with.

### DIFF
--- a/tests/sqllogic/sdb/pg/explain/aggregate_input_dedup.test
+++ b/tests/sqllogic/sdb/pg/explain/aggregate_input_dedup.test
@@ -1,6 +1,7 @@
 # Pre-aggregation projection dedupes equal non-volatile aggregate inputs and
-# filters into a single slot. Distinct aggregates keep their own input column,
-# and volatile inputs are never shared.
+# filters into a single slot, distinct aggregates included: distinct aggregates
+# that end up on the same slot share one distinct radix table, which the finalize
+# scans once for all of them. Volatile inputs are never shared.
 
 statement ok
 SET profiling_renderer_settings = {'deterministic': true};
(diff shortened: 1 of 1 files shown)


MBkkt (approved):
Add more test cases: same/differ-input x aggregate_func x distinct/no-distinct x filter/no-filter x same/differ-filter-input x order by/no order by x same/differ-order-by-input

PR #991, opened 2026-07-29
Improve `RaBitQ` with fast-scan

codeworse:
RaBitQ scoring was the slowest quantizer we ship, and `rabitq_bits = 1` — the
default — was its worst case: every candidate went through a per-vector
`faiss::RaBitQuantizer::get_distance_computer()` / `distance_to_code()` call, so a
kNN query paid a scalar distance computation per vector per probed cluster.

This replaces that with a single PQ4 fast-scan per cluster for **every** `nb_bits`,
so there is one code path instead of two:

- **Writer** buffers a whole cluster (`BeginCluster` / `FinishCluster`) and emits
  `[pq4_pack_codes(residual sign bits)][per-vector aux factors]`.
- **Reader** builds a 16-entry-per-group LUT from `compute_query_factors`, quantizes
  it to 8 bits (`quantize_LUT_and_bias` + `pq4_pack_LUT`), runs `accumulate_to_mem`
  over the cluster and finishes with `compute_1bit_adjusted_distance`.
- When `ex_bits = rabitq_bits - 1 > 0`, a second stage refines the top
  `kRaBitQRefinePool = 64` candidates per cluster with
  `compute_full_multibit_distance`.

No faiss changes; the on-disk payload layout changes, so indexes are not
cross-compatible with `main`.

## Benchmark

`dataset=dbpedia-100K dim=1536 metric=l2 k=10 nq=50 reps=3 quant=rabitq`

Both binaries were built from the same worktree with only
`libs/iresearch/include/iresearch/formats/ivf/quantizer.cpp` swapped between `main`
@ `8dec21d2e` and this branch, so faiss, duckdb, compiler and flags are identical.
Harness is checked in at `scripts/perf/rabitq_compare/`; full write-up in
`docs/RABITQ_MAIN_VS_BRANCH.md`.

The `scan` columns below are p50 of the server-side `Total Time` from
`EXPLAIN ANALYZE` **minus a measured 11.6 ms constant**: duckdb starts
`query.total_time` before the parser, so every raw measurement includes parsing the
~20 KB inlined 1536-float query literal. That floor was measured directly with the
same literals (`EXPLAIN ANALYZE SELECT <lit>::FLOAT[1536]`) and is the same on both
binaries (11.5 ms main / 11.6 ms branch). The literal cannot be avoided —
`iresearch_plan.cpp` requires a foldable constant, so a bound parameter drops the
query out of the ANN path.

recall@10 is micro-averaged against an exact `l2_distance` base-table scan computed
once and shared by all six variants. `rerank = 0` disables the raw-vector rerank so
the quantizer's own ranking is visible; `rerank = 4` is the shipping default.

### Build cost — unchanged

| bits | build_s main | build_s branch | index_MB main | index_MB branch |
|---:|---:|---:|---:|---:|
| 1 | 25.0 | 24.8 | 535 | 536 |
| 3 | 57.0 | 56.8 | 585 | 586 |
| 5 | 68.9 | 68.7 | 634 | 635 |

The 32-vector `pq4_pack_codes` padding plus the extra error float for `ex_bits > 0`
cost ~1 MB per variant at 100K rows.

### Headline: the 1-bit default

| bits | nprobe | rerank | scan main | scan branch | speedup | recall main | recall branch |
|---:|---:|---:|---:|---:|---:|---:|---:|
| 1 | 8 | 4 | 60.8 ms | 5.5 ms | **11.1x** | 0.884 | 0.878 |
| 1 | 32 | 4 | 226.5 ms | 8.2 ms | **27.6x** | 0.972 | 0.974 |
| 1 | 128 | 0 | 867.5 ms | 17.7 ms | **49.0x** | 0.846 | 0.862 |
| 3 | 32 | 4 | 92.5 ms | 26.0 ms | **3.6x** | 0.976 | 0.978 |
| 5 | 32 | 4 | 89.5 ms | 26.6 ms | **3.4x** | 0.966 | 0.968 |

<details>
<summary>Full matrix (3 bit widths × 5 nprobe × 2 rerank)</summary>

| bits | nprobe | rerank | scan main (ms) | scan branch (ms) | speedup | recall main | recall branch | Δrecall |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 1 | 1 | 0 | 11.6 | 4.5 | **2.6x** | 0.530 | 0.554 | +0.024 |
| 1 | 1 | 4 | 12.1 | 4.4 | **2.7x** | 0.584 | 0.576 | -0.008 |
| 1 | 4 | 0 | 32.6 | 4.2 | **7.8x** | 0.698 | 0.754 | +0.056 |
| 1 | 4 | 4 | 33.3 | 4.8 | **6.9x** | 0.792 | 0.818 | +0.026 |
| 1 | 8 | 0 | 60.5 | 4.6 | **13.2x** | 0.772 | 0.796 | +0.024 |
| 1 | 8 | 4 | 60.7 | 5.4 | **11.2x** | 0.884 | 0.878 | -0.006 |
| 1 | 32 | 0 | 225.4 | 7.3 | **30.9x** | 0.822 | 0.852 | +0.030 |
| 1 | 32 | 4 | 226.4 | 8.1 | **28.0x** | 0.972 | 0.974 | +0.002 |
| 1 | 128 | 0 | 867.4 | 17.6 | **49.3x** | 0.846 | 0.862 | +0.016 |
| 1 | 128 | 4 | 867.4 | 18.3 | **47.4x** | 1.000 | 1.000 | +0.000 |
| 3 | 1 | 0 | 7.3 | 4.7 | **1.6x** | 0.574 | 0.558 | -0.016 |
| 3 | 1 | 4 | 7.5 | 5.0 | **1.5x** | 0.578 | 0.558 | -0.020 |
| 3 | 4 | 0 | 15.0 | 6.3 | **2.4x** | 0.808 | 0.800 | -0.008 |
| 3 | 4 | 4 | 15.7 | 6.9 | **2.3x** | 0.816 | 0.812 | -0.004 |
| 3 | 8 | 0 | 26.5 | 8.8 | **3.0x** | 0.876 | 0.874 | -0.002 |
| 3 | 8 | 4 | 27.1 | 9.4 | **2.9x** | 0.894 | 0.890 | -0.004 |
| 3 | 32 | 0 | 91.4 | 25.3 | **3.6x** | 0.938 | 0.946 | +0.008 |
| 3 | 32 | 4 | 92.4 | 25.9 | **3.6x** | 0.976 | 0.978 | +0.002 |
| 3 | 128 | 0 | 344.4 | 93.4 | **3.7x** | 0.956 | 0.958 | +0.002 |
| 3 | 128 | 4 | 344.4 | 94.4 | **3.6x** | 1.000 | 0.998 | -0.002 |
| 5 | 1 | 0 | 7.3 | 5.4 | **1.4x** | 0.592 | 0.592 | +0.000 |
| 5 | 1 | 4 | 7.3 | 5.1 | **1.4x** | 0.592 | 0.596 | +0.004 |
| 5 | 4 | 0 | 15.2 | 6.3 | **2.4x** | 0.786 | 0.808 | +0.022 |
| 5 | 4 | 4 | 15.9 | 6.9 | **2.3x** | 0.788 | 0.812 | +0.024 |
| 5 | 8 | 0 | 25.9 | 9.0 | **2.9x** | 0.862 | 0.874 | +0.012 |
| 5 | 8 | 4 | 26.2 | 9.6 | **2.7x** | 0.864 | 0.880 | +0.016 |
| 5 | 32 | 0 | 89.4 | 26.0 | **3.4x** | 0.960 | 0.960 | +0.000 |
| 5 | 32 | 4 | 89.4 | 26.5 | **3.4x** | 0.966 | 0.968 | +0.002 |
| 5 | 128 | 0 | 340.4 | 95.4 | **3.6x** | 0.990 | 0.982 | -0.008 |
| 5 | 128 | 4 | 341.4 | 96.4 | **3.5x** | 1.000 | 1.000 | +0.000 |

</details>

### Findings

1. **The default configuration is the big win.** At `rabitq_bits = 1` the scan goes
   from 226 ms to 8.2 ms at nprobe=32 and from 868 ms to 18 ms at nprobe=128.
   `main`'s 1-bit path was also its *slowest* configuration — 868 ms vs 345 ms for
   its own 3-bit — so this removes a pathology, not just a constant factor.

2. **Recall does not regress; at 1 bit it improves.** With `rerank = 0`, where
   nothing masks a ranking error, branch 1-bit is ahead at all five nprobe points
   (+0.016 to +0.056). That is the estimator change, not noise in one cell:
   `compute_1bit_adjusted_distance` with an 8-bit quantized query is more accurate
   than `distance_to_code`, and it more than pays for quantizing the LUT. At 3 and 5
   bits the two agree within ±0.024, inside the ±0.026 binomial noise of 50 queries.

3. **One measurable cost.** The worst cell is `bits=5, nprobe=128, rerank=0`: 0.982
   vs 0.990 (−0.008) — the expected shape, since `main` refines every candidate's
   full multi-bit distance while this refines only the top 64 per cluster. It is
   gone at the default `rerank_factor = 4` (both 1.000).

4. **Multi-bit scales worse with nprobe than 1-bit.** Branch scan at nprobe=128 is
   18 ms at 1 bit but 94 ms at 3 bits (5.3x for two extra bits), because the refine
   stage is per probed cluster: 64 candidates × nprobe, independent of `k` and of
   `sdb_rerank_factor`. Worth noting that `kRaBitQRefinePool` is a fixed constant
   against a cluster size set by `sdb_ivf_posting_size` (default 1024), so the
   refine *fraction* — and the accuracy/latency trade-off with it — shifts if
   posting size changes.

### Caveats

- Query vectors are sampled from the dataset (same convention as
  `scripts/perf/vector_compare`), so each matches itself at distance 0 and all
  recall figures are inflated by a constant ~0.1. Identical on both sides, so
  Δrecall is unaffected.
- 50 queries ⇒ recall granularity 0.002, ~±0.026 noise at 95%. Only the 1-bit
  `rerank = 0` deltas clear that bar; 3/5-bit is parity.
- nprobe=128 is exhaustive at this scale (~100 leaves at
  `sdb_ivf_posting_size = 1024`; `nprobe` is clamped to the leaf count). Evidence:
  recall is exactly 1.000 at nprobe=128 / rerank=4 for all six variants. Read that
  row as pure scan throughput and as the quantizer's recall ceiling.
- Threads left at the machine default (128 cores).
- Cross-binary caveat for reviewers: `RaBitQStatsHeader` is `{nb_bits, d}` in both
  trees with no format discriminator, so a datadir built by one binary is silently
  mis-decoded by the other. The harness rebuilds every datadir from scratch.

### DIFF
--- a/libs/iresearch/include/iresearch/formats/ivf/quantizer.cpp
+++ b/libs/iresearch/include/iresearch/formats/ivf/quantizer.cpp
@@ -82,7 +82,9 @@ size_t FastScanNsq(size_t m) noexcept { return m + (m & 1); }
 
 constexpr int64_t kRaBitQRotationSeed = 0x5a17b17c5eed5eedULL;
 
-uint32_t RotatedDim(uint32_t d) noexcept { return std::bit_ceil(d); }
+uint32_t RotatedDim(uint32_t d) noexcept {
+  return std::max<uint32_t>(kFastScanBits, std::bit_ceil(d));
+}
 
 void GenerateSigns(uint32_t rotated_d, int64_t seed,
                    std::vector<float>& signs) {
@@ -627,7 +629,8 @@ class RaBitQuantizerWriter final : public QuantizerWriter {
       _metric{FaissMetric(metric)},
       _storage{
         faiss::rabitq_utils::compute_per_vector_storage_size(nb_bits, _rd)},
-      _ex_code_size{(static_cast<size_t>(_rd) * _ex_bits + 7) / 8} {
+      _ex_code_size{(static_cast<size_t>(_rd) * _ex_bits + 7) / 8},
+      _sign_stride{FastScanNsq(_rd / kFastScanBits) / 2} {
     GenerateSigns(_rd, kRaBitQRotationSeed, _signs);
     _stats.resize(sizeof(RaBitQStatsHeader));
     WritePodHeader(RaBitQStatsHeader{nb_bits, d}, _stats.data());
@@ -641,7 +644,7 @@ class RaBitQuantizerWriter final : public QuantizerWriter {
   }
 
   void BeginCluster(size_t total_docs) final {
-    _sign_codes.assign(total_docs * (_rd / 8), 0);
+    _sign_codes.assign(total_docs * _sign_stride, 0);
     _aux.assign(total_docs * _storage, 0);
     _filled = 0;
   }
@@ -652,7 +655,7 @@ class RaBitQuantizerWriter final : public QuantizerWriter {
       return;
     }
     SDB_ASSERT(_centroid.size() == _rd);
-    const size_t sign_stride = _rd / 8;
+    const size_t sign_stride = _sign_stride;
     std::vector<float> rotated(_rd);
     std::vector<float> residual(_rd);
     for (size_t i = 0; i < n; ++i) {
@@ -724,6 +727,7 @@ class RaBitQuantizerWriter final : public QuantizerWriter {
   faiss::MetricType _metric;
   size_t _storage;
   size_t _ex_code_size;
+  size_t _sign_stride;
   std::vector<float> _signs;
   std::vector<float> _centroid;
   std::vector<byte_type> _stats;
--- a/libs/iresearch/include/iresearch/formats/ivf/quantizer.cpp
+++ b/libs/iresearch/include/iresearch/formats/ivf/quantizer.cpp
@@ -21,7 +21,8 @@
 #include "iresearch/formats/ivf/quantizer.hpp"
 
 #include <faiss/impl/ProductQuantizer.h>
-#include <faiss/impl/RaBitQuantizer.h>
+#include <faiss/impl/RaBitQUtils.h>
+#include <faiss/impl/RaBitQuantizerMultiBit.h>
 #include <faiss/impl/ScalarQuantizer.h>
 #include <faiss/impl/fast_scan/fast_scan.h>
 #include <faiss/utils/AlignedTable.h>
@@ -34,6 +35,7 @@
 #include <cmath>
 #include <cstdint>
 #include <cstring>
+#include <functional>
 #include <limits>
 #include <vector>
 
@@ -69,6 +71,8 @@ std::span<const byte_type> FloatSpan(const std::vector<float>& v) noexcept {
 
 constexpr size_t kFastScanBbs = 32;
 constexpr uint32_t kPqNbits = 4;
+constexpr size_t kFastScanBits = 4;
+constexpr size_t kFastScanKsub = size_t{1} << kFastScanBits;
 
 size_t RoundUp(size_t n, size_t multiple) noexcept {
   return (n + multiple - 1) / multiple * multiple;
@@ -141,6 +145,10 @@ struct RaBitQStatsHeader {
   uint32_t nb_bits;
   uint32_t d;
 };
+
+constexpr uint8_t kRaBitQQueryBits = 8;
+constexpr bool kRaBitQCentered = false;
+constexpr size_t kRaBitQRefinePool = 64;
 
 class ScalarQuantizerWriter final : public QuantizerWriter {
  public:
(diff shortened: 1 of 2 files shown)


MBkkt (approved):
Please add more tests, it's not enough tests from you.
Tests should be with vedernikoff

PR #1001, opened 2026-07-31
feat: support azure storage

mkornaukhov:
feat: Azure blob storage support (az://) via vendored duckdb-azure

Adds the azure extension to the static build: COPY/read_csv/read_parquet/glob
over az:// URIs, CREATE SECRET (TYPE azure) with connection strings and scoping.

Vendoring (ssh serenedb forks, ClickHouse-style monolith wrappers):
- third_party/duckdb_azure @ v2026.07.30, one fork commit
- third_party/azure-sdk-for-cpp @ v2026.06.10 (12.18.0), unpatched upstream
- third_party/libxml2 @ v2026.07.30 (zlib via zlib-ng)
- third_party/azure-cmake, third_party/libxml2-cmake build the single static
  libs _azure_sdk / _libxml2

One HTTP stack for azure and httpfs:
- the SDK's curl transport is not compiled (BUILD_TRANSPORT_CUSTOM_ADAPTER);
  duckdb_azure injects a transport built on duckdb's HTTPUtil, so requests go
  through the httpfs curl client and its connection cache
- the SDK therefore needs no patches and spawns no threads on any reachable
  path: no connection-pool cleanup thread, no scheduler wiring in serened
- range reads stream on the calling thread; azure_read_transfer_*,
  azure_transport_option_type and azure_http_logging options are removed

duckdb fork: azure extension source shim; the PG single-column alias rename is
scoped to set-returning functions, so a synthesized replacement-scan alias no
longer renames a file's only column (`SELECT v FROM 'az://.../f.parquet'` used
to fail with "Referenced column v not found").

Tests: *_azure.test_slow auto-launch Azurite in run.sh (like minio for *_s3):
blob_azure (roundtrips, >8MiB streaming, unicode/space names, glob, error
cases), inverted_index_azure (inverted index over az:// parquet/csv views).
Plus single_column_file_alias, any/pg/simple/function_alias_columns (runs
against real postgres too) and the show_all snapshot.


tests/sqllogic/run.sh
```diff
@@ -659,6 +740,9 @@ launch_external() {
 	if [[ "$needs_s3" == "true" ]]; then
 		launch_s3
 	fi
+	if [[ "$needs_azurite" == "true" ]]; then
+		launch_azurite
```
MBkkt: why azurite?

tests/sqllogic/sdb/pg/simple/blob_azurite.test_slow
```diff
@@ -0,0 +1,363 @@
...
+  TYPE azure,
+  CONNECTION_STRING '${AZURITE_CONNECTION_STRING}',
+  SCOPE 'az://corner-ghost/'
+);
+
+
+statement error
+SELECT * FROM read_csv('az://corner-ghost/x.csv');
+----
+db error: ERROR: AzureBlobStorageFileSystem Read to az://corner-ghost/x.csv/**/*.csv failed with ContainerNotFound Reason Phrase: The specified container does not exist.
```
MBkkt: is it expected?

server/connector/azure_cleanup_scheduler.cpp
```diff
@@ -0,0 +1,89 @@
...
+yaclib::Future<> CleanupLoop(std::chrono::milliseconds interval,
+                             std::function<bool()> tick) {
+  auto& s = BackgroundScheduler::instance();
+  try {
+    for (;;) {
+      co_await s.Delay(interval);
+      if (lifecycle::IsStopping()) {
+        break;
+      }
+      const bool proceed = co_await s.Run([&tick] {
```
MBkkt: co_await On(s);

third_party/CMakeLists.txt
```diff
@@ -689,6 +689,16 @@ set(AWSSDK_FOUND ON CACHE BOOL "" FORCE)
...
+################################################################################
+## Azure SDK (monolithic build) + libxml2
+################################################################################
+
+sdb_update_module(${GIT_EXECUTABLE} "libxml2" ${CMAKE_CURRENT_SOURCE_DIR} "README.md")
+sdb_update_module(${GIT_EXECUTABLE} "azure-sdk-for-cpp" ${CMAKE_CURRENT_SOURCE_DIR} "README.md")
+
+add_subdirectory(libxml2-cmake)
+add_subdirectory(azure-cmake)
+
```
MBkkt, line 707: looks like 2 separate entries

third_party/duckdb
MBkkt, line 1: 1. make commits merge ready
2. Can you provide example? And sqllogic test exists, right?
3. Can we make it without additional bool member?
4. Why all other table function usages are safe?

third_party/CMakeLists.txt
```diff
@@ -763,6 +774,17 @@ target_include_directories(
...
+# duckdb_azure pins CMAKE_CXX_STANDARD to 17 in its own CMakeLists, but our
+# abseil (pulled in via duckdb fork headers) needs the project standard.
+foreach(_azure_ext_target azure_extension azure_loadable_extension)
+    if(TARGET ${_azure_ext_target})
+        set_target_properties(
+            ${_azure_ext_target}
+            PROPERTIES CXX_STANDARD ${CMAKE_CXX_STANDARD}
+        )
+    endif()
+endforeach()
```
MBkkt: just remove pin from their cmake and remove this

third_party/azure-sdk-for-cpp
MBkkt, line 1: patches doesn't look good, I think you need to investigate current difference between upstream azure and s3, answer on question what is better, and then make our fork for s3 and azure same.

tests/sqllogic/sdb/pg/settings/show_all.test
```diff
@@ -33,6 +33,22 @@ auto_fallback_to_full_download	on	Allows automatically falling back to full file
...
 autoinstall_known_extensions	off	Whether known extensions are allowed to be automatically installed when a query depends on them
 autoload_known_extensions	off	Whether known extensions are allowed to be automatically loaded when a query depends on them
+azure_account_name	NULL	Azure account name, when set, the extension will attempt to automatically detect credentials
+azure_context_caching	1	Enable/disable the caching of some context when performing queries. This cache is by default enable, and will for a given connection keep a local context when performing a query. If you suspect that the caching is causing some side effect you can try to disable it by setting this option to false.
+azure_credential_chain	NULL	Ordered list of Azure credential providers, in string format separated by ';'. E.g. 'cli;workload_identity;managed_identity;env'
+azure_endpoint	blob.core.windows.net	Override the azure endpoint for when the Azure credential providers are used.
+azure_http_logging	1	Enable Azure HTTP request logging into the DuckDB HTTP log. When true (default), logging follows DuckDB's HTTP log settings (see enable_logging('HTTP')). Set to false to disable Azure HTTP logging.
```
MBkkt: I think same setting for duckdb itself already removed?

tests/sqllogic/run.sh
mkornaukhov: it's kind of minio for azure stack
replaced with azure

tests/sqllogic/sdb/pg/simple/blob_azurite.test_slow
mkornaukhov: yes, it's absent container

third_party/CMakeLists.txt
mkornaukhov: fixed

third_party/CMakeLists.txt
mkornaukhov: fixed

server/connector/azure_cleanup_scheduler.cpp
mkornaukhov: removed

third_party/duckdb
mkornaukhov: Fixed.
Look at `tests/sqllogic/sdb/pg/simple/single_column_file_alias.test`.
The main problem is that we should differ auto generated aliases and user-written. Effectively, if `x.parquet` consist of single column, than
`select * from 'x.parquet'` ---> `select * from read_parquet('x.parquet') as x`, and for pg-compatibility we need to output it like 
```
x
-
1
```
But it breaks the following:
```
select v from 'x.parquet';
```
Expected:
```
v
-
1
```
Actual smth like `Error: v not found, did you mean x?`

mkornaukhov: Changed it a little, remove using table alias for single column table returning function, kept only set returning functions.

third_party/azure-sdk-for-cpp
mkornaukhov: Changed on clickhouse-like approach: reusing duckdb common and httpfs stuff

tests/sqllogic/sdb/pg/settings/show_all.test
mkornaukhov: They are kept on extension, removed in out fork

MBkkt (approved):

PR #1003, opened 2026-07-31
fix: mixed seek/advance case for IVF doc iterator

Dronplane:
If seek is called after advance it ignores buffered docs and may skip matches


libs/iresearch/include/iresearch/search/vector_similarity_query.cpp
```diff
@@ -208,6 +208,15 @@ class QVectorIterator : public VectorDistanceIterator {
     if (target <= _doc) {
       return _doc;
     }
+    if (_pos < _len && target <= _docs[_len - 1]) {
```
codeworse, line 211: don't we need the same fix for `RawVectorIterator`?

Dronplane: I don't see a buffering here. So it should be not affected

codeworse (approved):

PR #60, opened 2025-12-14
feat : Add default values

pashandor789:

PR #1005, opened 2026-08-02
test(dict_fsst): cover the cut paths that write past the block

gnusi:
## Problem

`CHECKPOINT` on a `DICT_FSST` column could write a segment past the end of its block. In debug that aborts on `D_ASSERT(layout.total <= block_size)`; in release the assert compiles out and `WriteNative` writes its regions **past the end of the block buffer**.

Root cause, in all cases: one row re-prices the entire segment layout, while the cut logic assumes size grows a row at a time. Four such defects are fixed in **serenedb/duckdb#60**, the last of them (a selection-width bump on the never-encoded path) found by the 564k-document corpus after the suite here was already green -- which is why this PR also adds a test distilling it.

## Resolution

1. Pin `third_party/duckdb` at the fix.
2. Add thirteen `recovery/dict_fsst*` tests to the pre-existing three. There was no coverage of this class before: no existing dict_fsst test crossed a uniqueness flip, which is why a 564k-document corpus kept dying at checkpoint while the suite stayed green.

| test (`tests/sqllogic/recovery/`) | covers | fails pre-fix |
| --- | --- | --- |
| `dict_fsst_unique_flip_block_overflow` | the flip on the cleaved path | yes |
| `dict_fsst_native_unique_flip_block_overflow` | the flip on the **native** path (`CutPlain`/`WriteNative`, where the original abort came from) | yes |
| `dict_fsst_plus_rewind_unique_flip` | a rewind flushing a candidate its `fit_rows` certificate never priced | yes |
| `dict_fsst_cut_flip_sweep` | the flip by **duplicate and by NULL**, swept across five distances from the block, both storage versions | yes |
| `dict_fsst_cut_enc_width` | the `bits(max_enc_len)` bump **in isolation** -- every shape stays all-unique, so only the width term can catch it | yes |
| `dict_fsst_cut_rewind_candidates` | the rewind under each commitment: `PLUS_ROW`, `PLUS_SORTED`, `PLAIN` | yes |
| `dict_fsst_cut_small_dict_width` | the selection-width bump on the never-encoded path: 127 distinct values riding a 7-bit width to ~292k rows in one large row group, then the 128th arrives | yes |
| `dict_fsst_cut_width_boundaries` | the same bump at the 4->5 bit boundary (~515k rows, the corpus trace's shape), with NULLs mixed through, a benign mid-segment bump that must NOT cut, and a default-row-group control | yes |
| `dict_fsst_cut_fuzz` | hash-driven near-block stress: cardinality ramps through every width regime, alternating unique/low-card blocks, 12% NULLs + length spikes, and all of it combined -- each table crossing several blocks, every row verified against its generator | yes |
| `dict_fsst_cut_prefix_budget` | the prefix-group budget at and around its doubling boundaries, and at the `entry_n/2` ceiling | no |
| `dict_fsst_cut_forced_modes` | all eight `force_dict_fsst_mode` values on the re-pricing shape | no |
| `dict_fsst_cut_degenerate` | one row, one NULL, all-NULL, all-empty, all-identical, one huge value, single group, interleaved holes, never-encoded dictionary, and the junk-prefix bait shape (sorted neighbours sharing exactly one byte, which must come out native rather than bloat a plus layout) | no |

The last three do not reproduce the original bug -- they guard the budget machinery and the edge paths the pre-fix build never reaches -- so they are listed honestly as coverage, not as regression tests.

## What the new tests pin

**`cut_flip_sweep`** removes the dependence on one hand-tuned constant. The existing flip test pins 33,000 unique rows; this sweeps 29k/31k/33k/35k/37k so that if the block size or the encoding shifts, some column still lands on a boundary. It also covers the NULL flip, which ends all-unique exactly as a duplicate does (both add no entry), including a single NULL mid-run that kills all-unique while the segment is still far from the block.

**`cut_enc_width`** isolates the other re-price. Every shape is all-unique start to finish, so the uniqueness term can never fire and the width term is the only thing that can catch it: one late step, a staircase across four bitpacking widths, steadily growing lengths, the wide value placed first as a control, and both re-prices on one segment.

**`cut_rewind_candidates`** drives the rewind under all three commitments by controlling *where* the shared prefixes are. Row-order neighbours sharing a prefix makes the row cleave win (`PLUS_ROW`, no selection buffer); prefixes only in sorted order with the rows shuffled leaves the row cleave nothing to group (`PLUS_SORTED`); a native attach keeps `allow_plus` false so every cut runs `CutPlain`. Each is pushed over by a duplicate and again by a NULL, and the duplicate is asserted to appear exactly twice and the NULL exactly once -- a rewind that dropped or double-counted the rows it moved shows up there even when the byte comparison passes.

**`cut_prefix_budget`** is the near-boundary research codified. Group size selects how many groups a dictionary wants, chosen so the count lands on and just under the budget's doublings and at the `entry_n/2` ceiling where it cannot grow. It also asserts grouped shapes pack more rows per segment than the ungrouped control, which fails if the budget ever starves the cleave.

## Verification discipline

Every test verifies **by value**: each row is compared against the expression that generated it, and each asserts the layout under test was actually reached (`DICT_FSST` present, or a plus mode present, or the exact per-table mode list under a pin). Both matter because in release the bug corrupts rather than aborts, and because a plan change must not silently retire the coverage.

Determinism comes from `ORDER BY` in every `CREATE TABLE AS`, which fixes the row order feeding the compressor and so the segment boundaries. (An earlier revision of this description claimed `threads=1`; no test sets it, and none needs to.)

`cut_forced_modes` checkpoints each table while its own mode is still in force, because compression is chosen when a segment is written, not when the rows are inserted -- a single checkpoint at the end silently compresses every table under whatever the setting was last. Note `force_dict_fsst_mode` is a **global**, not session, setting; it is safe here only because `run_recovery_tests.sh` gives each recovery file its own serened. Batching these files onto one shared server makes the mode assertions in this file, and in the pre-existing `plus_modes` / `plus_fuzz`, interfere.

## Results

- Full matrix, both directions, one harness run each: **15/15 pass on the fixed engine; 9 of 15 fail on unmodified `v2026.07.07`**, carrying the original `layout.total <= block_size` assert. The six that pass pre-fix are exactly the six documented as coverage rather than regression (the three pre-existing files plus `cut_prefix_budget` / `cut_forced_modes` / `cut_degenerate`).
- `cut_small_dict_width` was additionally validated against three binaries back to back: fails pre-fix, fails on the earlier commits of #60 (the defect predates them), passes with the guard.
- `sdb/pg/index/*`, `sdb/pg/dml/*`, `any/pg/simple/*`: **538/538**, both engines.
- The corpus that surfaced this, 564,316 documents of `<kind>:<nanoid>` keys, loads cleanly on **6 of 6 fresh datadirs** with its `force_compression='uncompressed'` workaround removed.

## Merge order

#60 must land on `v2026.07.07` first, then the submodule pointer here gets re-pointed off the PR branch commit.

### DIFF
--- a/tests/sqllogic/recovery/compression_block_boundaries.test
+++ b/tests/sqllogic/recovery/compression_block_boundaries.test
@@ -1,0 +1,259 @@
+# Segment sizing at block boundaries for EVERY forceable compression, not just dict_fsst.
+#
+# The dict_fsst work showed the failure class: a codec's segment must stop accepting rows before its
+# serialized layout exceeds one block, and the hard cases are single rows that re-price more than
+# themselves -- width steps, mode flips, exception bursts, incompressible spikes. Each codec here
+# gets the shape that stresses ITS version of that, at ROW_GROUP_SIZE 1048576 so segments actually
+# reach blocks (default row groups cap segments long before the boundary).
+#
+# force_compression is GLOBAL and compression is chosen at CHECKPOINT, so every table is
+# checkpointed while its own setting is in force. Everything is a pure function of ord: failures
+# replay exactly, and every row is verified against its generator -- in release an overflow corrupts
+# instead of aborting, so surviving the checkpoint proves nothing by itself.
+#
+#   bitpacking  BIGINT magnitude staircase stepping 18 bits per 100k rows, then a 2^62 spike late
+#   rle         long runs collapsing to run-length-1 near the boundary, NULLs threaded through
+#   dictionary  the dict_fsst defect-4 shape (127 distinct values, then the 128th at ~292k rows).
+#               Legacy Dictionary is subsumed by DICT_FSST in this fork: forcing it FALLS BACK to
+#               Uncompressed, and this file pins that fallback so a silent revival gets noticed.
+#   fsst        all-unique strings with late length spikes. Same subsumption, same pinned fallback.
+#   zstd        compressible prose flipping to incompressible digest soup near the boundary
+#   uncompressed strings crossing many blocks with a 60KB value dropped mid-stream
+#   alp         ALP-friendly decimals with a late burst of high-entropy exceptions
+#   alprd       high-entropy doubles end to end (the exceptions ARE the data)
+#   roaring     validity: dense NULL runs, sparse NULLs, and alternating flips across 1M rows
+
+control substitution on
+
+
+statement ok
+ATTACH '${__TEST_DIR__}/cbb.db' AS cbb (TYPE duckdb, STORAGE_VERSION 'serenedb_v1', ROW_GROUP_SIZE 1048576);
+
+
+statement ok
+SET force_compression='bitpacking';
+
+
+statement ok
+CREATE TABLE cbb.t_bitpacking AS
+  SELECT ord, CASE WHEN ord = 390000 THEN 4611686018427387904
+              ELSE (ord % 200) * (1::BIGINT << LEAST(54, (ord // 100000) * 18)::INT) END::BIGINT AS v
+  FROM range(400000) t(ord) ORDER BY ord;
+
+
+statement ok
+CHECKPOINT cbb;
+
+
+statement ok
+SET force_compression='rle';
+
+
+statement ok
+CREATE TABLE cbb.t_rle AS
+  SELECT ord, CASE WHEN ord % 13 = 7 THEN NULL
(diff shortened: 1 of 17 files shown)


Dronplane (approved):
LGTM

MBkkt (approved):

PR #1006, opened 2026-08-03
feat: Panorama scan for quant='none', radius via .pay, block-based quantizer protocol

codeworse:
# Faster vector search: Panorama scan for `quant='none'`, radius via `.pay`, block-based quantizer protocol

## Summary

1. **`quant='none'` becomes a real quantizer.** It writes its own `.pay` payload using faiss's Panorama layout — a PCA-rotated basis with per-level suffix norms — and prunes candidates during the posting scan via an exact Cauchy–Schwarz bound against the running k-th best. Previously `MakeQuantizerWriter(None)` returned `nullptr`, no `.pay` was written, and every query re-read raw vectors from the `.col` columnstore.
2. **Radius queries go through `.pay`.** `RangeVectorQuery::Execute` builds per-cluster quantized iterators with a constant pruning threshold instead of a raw reranker over the columnstore. Membership is gated on the payload; for a lossy quantizer a survivor's reported distance is then rescored exactly from the index's own vectors.
3. **The quantizer writer/reader protocol is now block-based.** A single `PayloadBlockSetting {group_size, record_size, pad_tail}` is the whole contract between the two sides. The writer has no cluster-lifecycle hooks and no tail buffer; the reader owns no `IndexInput` and no cluster offsets. This is what makes Panorama's batched layout expressible alongside PQ/RaBitQ FastScan groups and SQ's flat records.
4. **The two vector filters stop duplicating each other** — one shared `VectorFilterOptions` + `PrepareVectorState`.
5. **A DuckDB-side JSON fix** that (1) and (2) exposed.

## Panorama for `quant='none'`

**Layout.** The stats blob in `.idx` carries `PanoramaStatsHeader{n_levels, d}` followed by the `d×d` row-major PCA rotation. The payload is written in batches of `faiss::Panorama::kDefaultBatchSize` (128) rows: `count * (n_levels+1)` floats of cumulative suffix norms, then `count * d` floats of rotated codes in Panorama level layout. With `kPanoramaLevelWidth = 32`, `n_levels = ceil(d/32)`, so the logical record is `(d + n_levels + 1) * sizeof(float)` — for 768d that's 793 vs 768 floats, **+3.3% per vector**.

A partial trailing batch is written as its own smaller, self-describing Panorama (`MakePanorama(_d, count)`), and the reader reconstructs it the same way from the block length. There is no separate tail structure on either side.

**Gating.** The rotation is declined when `d < kPanoramaMinDim` (64), `metric == L1` (a rotation doesn't preserve L1), or there are too few rows to train a `d×d` basis (`rows >= max(1024, 8*d)`). On decline the stats blob is empty, the reader sees `_levels == 0`, and the payload is a verbatim copy of the input floats — plain contiguous `d` floats, scanned without pruning.

**Training is sampled, not exhaustive.** `TrainSamples(rows)` asks for `max(4096, 8*d)` rows and `IvfBuilder::Compute` fills that from a reservoir over the column, so the PCA cost does not scale with table size.

**Threshold plumbing.** `DisjointClusterUnion` owns a `ScoreThresholdAttr` and binds every child iterator to it, so the top-k collector's running boundary propagates into Panorama's per-batch bound; `QVectorIterator::BindThreshold` de-boosts by `_boost` before each block. The radius path uses `BindConstantThreshold` instead — its threshold is known up front, so pruning is effective from the first candidate rather than needing warm-up.

## Quantizer protocol rework

`QuantizerWriter` before → after:

| removed | added / changed |
|---|---|
| `BeginCluster(total_docs)` | `BlockSetting() -> PayloadBlockSetting` |
| `FinishCluster(out)` | `EncodeBlock(out, vecs, n)` (was `EncodeCluster`, and no longer `const`) |
| `CodeSize()` | `TrainSamples(rows)` |
| `StatsBytes() -> span` | `Serialize(DataOutput&)` |

- **No cluster lifecycle.** The payload is a sequence of self-contained groups of `group_size` records, so there is nothing to open or flush. `IvfTermReader::WriteTermPayload` derives its gather stride from `BlockSetting().group_size` (rounding `STANDARD_VECTOR_SIZE` down to a multiple of it) and hands whole groups to `EncodeBlock`; the writer handles its own partial tail — Panorama with a smaller batch, PQ/RaBitQ by zero-padding the group's codes, norms and side arrays. All four writers are now stateless across calls, and every `mutable` qualifier in the file is gone.
- **The reader stops doing I/O.** `ComputeBlock(std::span<const byte_type> block, score_t threshold, score_t* out)` replaces `StartCluster(pay_start, num_docs, centroid)` + `ComputeBlock(offset, length, out)`. `VectorBlockReader` moved out of its own header (`vector_block_reader.hpp` deleted) into the anon namespace of `vector_similarity_query.cpp`, where `QVectorIterator` does the group-aligned windowing and caches a partially-consumed group. `MakeQuantizerReader` and `QuantizerCodebook::MakeReader` no longer take an `IndexInput`.
- **`Serialize(DataOutput&)` writes through** rather than materialising a `bstring` — faiss's owned `std::vector` members make a zero-copy view design impossible, and this runs once per flush.
- **`TrainSamples` replaces the per-quantizer booleans** (`pq`, `sq_train`, `needs_centroid`) that `IvfBuilder::Compute` used to switch on. One reservoir path now serves both PQ residuals and Panorama's PCA rows; SQ returns the `kTrainStreaming` sentinel and is fed incrementally as before. The PQ residual cap dropped from 65536 to `ksub * 256 = 4096`, which is where faiss caps PQ k-means anyway, with no measured recall change.
- **Every quantization kind now has a writer**, so `IndexFeatures::Pay` is unconditional and the `qw != nullptr` branches through `ivf_writer` became asserts.
- `_accu` in the RaBitQ writer is a plain `std::array<uint16_t, kFastScanBbs>` on the stack instead of an aligned heap table.

## Filter deduplication

`ByVectorSimilarity::PrepareSegment` and `ByRadius::PrepareSegment` had drifted into two copies of the same ~88 lines of centroid-tree walking, codebook parsing, cookie collection and cost estimation — with the radius copy missing the codebook entirely, which is why it could only use the columnstore.

Both option structs now derive from a shared `VectorFilterOptions` (query, ids, metric, quant, inner filter, `operator==`), and both `PrepareSegment` bodies call one `PrepareVectorState` in [vector_filter_util.hpp](libs/iresearch/include/iresearch/search/vector_filter_util.hpp). The only difference is `nprobe`, which radius passes as `uint32_t max` (probe everything). `vector_similarity_filter.cpp` goes 176 → 47 lines and `vector_radius_filter.cpp` 87 → 49.

Server-side: `MakeVectorFilter` now forwards `quant` to the radius filter (it needs the payload kind), `ApplyScoreEmit` moved out of the individual producers into `AccountAndWriteVirtualColumns` so raw "larger = nearer" scores become user-facing at exactly one point, and the collector pool size is computed by one `CollectorPoolSize` helper instead of two copies.

## Benchmarks

Measured on **wikipedia-cohere-1M** (768d, metric=ip, k=10, clients=32, runs=3), 6 index builds, 34 points, no restarts. Panorama has no runtime switch, so the A/B was done with a temporary env gate on the early return in `PanoramaQuantizerWriter::Train` — with it set, no rotation is trained and the reader takes the `_levels == 0` plain-scan path, so both arms come from one binary and the "off" arm is production code, not a bespoke branch. **That gate is not part of this PR.** These numbers predate the protocol rework, which changed how the payload is chunked but not what the scan computes.

**QPS at matched recall, Panorama ÷ plain scan:**

| `sdb_ivf_posting_size` | r=0.80 | r=0.90 | r=0.95 |
|---|---|---|---|
| 1024 (current default) | 1.79× | 3.87× | 5.72× |
| 4096 | 2.02× | 4.30× | **7.16×** |
| 16384 | 1.94× | 2.85× | 3.33× |

**Cost:** +1.81% index size (5895 vs 5790 MB) and ~3% build time, constant across posting sizes.

The gain scales with how deep the scan goes, which is the mechanism behaving as designed — pruning does nothing until the k-th-best threshold tightens:

| posting=1024, nprobe | 8 | 32 | 64 | 128 | 256 |
|---|---|---|---|---|---|
| speedup | 1.11× | 2.73× | 4.32× | 5.88× | 6.30× |

At the lowest nprobe it is break-even: the +1.81% bytes are paid for pruning that can't yet bite.

**Posting size has an interior optimum around 4096**, which is a new shape — SQ has no interior optimum (monotone decreasing above the tree-collapse boundary) and PQ prefers large postings because its FastScan LUT amortizes. Panorama wants mid-size: long enough for the threshold to tighten mid-list, short enough that routing granularity still buys something. Going 1024 → 4096 is worth ~+23% QPS at fixed recall; 16384 gives most of it back (1518 → 928 qps at r=0.95, only ~61 leaves over 1M rows). **This argues for a per-quantizer posting-size default, which this PR does not change.**

Caveats worth recording:

- Recall differs slightly between arms and **the sign is not stable** (+0.9% at t=1024, −1.5% at t=16384). Two independent plain builds differ by ±0.4% on their own, so this is k-means build noise, not a Panorama effect — the tree is built on unrotated vectors in both arms (`CentroidsBuilder::Create`). All figures above are at matched recall, never matched nprobe.
- **r=0.99 was never reached** — the nprobe grids top out at 0.96–0.98. Since the advantage grows monotonically with nprobe, the high-recall regime is both uncovered and the one where Panorama should look best.
- Iso-recall interpolation brackets are wide (nprobe doubles between points). The per-point grid is the stronger evidence and agrees: at t=4096/nprobe=32 Panorama does 1591 qps at recall 0.9443 vs plain's 225 qps at recall 0.9462 — 7.1× with plain holding *slightly higher* recall, no interpolation involved.

## JSON lookup fix (DuckDB submodule)

`JSONLookupScan` materialized each looked-up row with `Slice(output, row, row+1)`, which is only correct for flat fixed-width columns — a nested type's slice gets its own child buffers, so every output LIST row aliased one row's child data. Fixed by transforming into a one-row scratch chunk and `VectorOperations::Copy`ing it into place.

This surfaced *because* of the `.pay`/rerank work: an IVF index on a view over `read_json()` fetched vectors back and saw one repeated vector, and stale child data reaching a `DOUBLE[] → FLOAT[8]` cast raised "value out of range for FLOAT".


tests/sqllogic/sdb/pg/index/inverted_index_matrix_vector_ivf.test
```diff
@@ -155,36 +155,40 @@ EXPLAIN SELECT pk FROM m_vec_ivf_st
...
+│ Lookup: table                  │
+│ Index Filter:                  │
+│ ╭─ Vector Range ─╮             │
+│ │ Field: emb     │             │
+│ │ Metric: l2     │             │
+│ │ Radius: < 1.5  │             │
+│ ╰────────────────╯             │
+│ Column Filter:                 │
+│ sdb_inverted_index_score < 1.5 │
+│ Projections: pk                │
```
MBkkt: looks wrong?

codeworse: Kinda yes. In the branch implementation SereneDB rerank exact distances from the range query and drop some of the vectors that didn't fit, but this will be removed, so the query plan will be the same

libs/iresearch/include/iresearch/formats/ivf/ivf_writer.cpp
```diff
@@ -111,6 +108,16 @@ BuiltIvf IvfBuilder::Compute(const ColumnReader& vector_column,
     std::vector<std::span<const float>> cents(
       needs_centroid ? STANDARD_VECTOR_SIZE : 0);
     size_t gathered = 0;
+    const auto reservoir = [&]() -> float* {
```
MBkkt:
```suggestion
    const auto reservoir = [&] -> float* {
```

libs/iresearch/include/iresearch/formats/ivf/ivf_writer.cpp
```diff
@@ -357,11 +359,12 @@ void IvfTermReader::WriteTermPayload(IndexOutput& out,
   if (n == 0) {
     return;
   }
-  _qw->BeginCluster(n);
+  const size_t group = std::max<size_t>(1, _qw->BlockSetting().group_size);
+  const size_t step = std::max(group, STANDARD_VECTOR_SIZE / group * group);
```
MBkkt: .

libs/iresearch/include/iresearch/search/vector_filter_util.hpp
```diff
@@ -20,16 +20,40 @@
...
+  field_id postings_id = field_limits::invalid();
+  VectorMetric metric = VectorMetric::L2Sqr;
+  VectorQuantization quant = VectorQuantization::None;
+  std::shared_ptr<const Filter> inner;
+
+  bool operator==(const VectorFilterOptions& rhs) const noexcept {
+    return query == rhs.query && centroids_id == rhs.centroids_id &&
+           postings_id == rhs.postings_id && metric == rhs.metric &&
+           quant == rhs.quant && inner == rhs.inner;
+  }
```
MBkkt: = default

libs/iresearch/include/iresearch/search/vector_radius_filter.hpp
```diff
@@ -20,32 +20,22 @@
...
 
-  std::vector<float> query;
-  field_id centroids_id = field_limits::invalid();
-  field_id postings_id = field_limits::invalid();
-  VectorMetric metric = VectorMetric::L2Sqr;
   float radius = 0.f;
   bool inclusive = false;
-  std::shared_ptr<const Filter> inner;
 
   bool operator==(const ByRadiusOptions& rhs) const noexcept {
```
MBkkt: default

libs/iresearch/include/iresearch/search/vector_similarity_filter.hpp
```diff
@@ -20,31 +20,22 @@
...
 
-  std::vector<float> query;
-  field_id centroids_id = field_limits::invalid();
-  field_id postings_id = field_limits::invalid();
-  VectorMetric metric = VectorMetric::L2Sqr;
-  VectorQuantization quant = VectorQuantization::None;
   uint32_t nprobe = 1;
-  std::shared_ptr<const Filter> inner;
 
   bool operator==(const ByVectorSimilarityOptions& rhs) const noexcept {
```
MBkkt: default

libs/iresearch/include/iresearch/CMakeLists.txt
```diff
@@ -252,6 +252,14 @@ include_directories(SYSTEM ${IRESEARCH_EXTERNAL_INCLUDES})
...
 add_library(iresearch-static STATIC ${IResearch_core_sources})
 
+# faiss::Panorama's prune_kernel asks for loop vectorization that aarch64 cannot
+# honour (a gather through active_indices plus a narrow store), and Release
+# promotes -Wpass-failed to an error.
+set_source_files_properties(
+    ./formats/ivf/quantizer.cpp
+    PROPERTIES COMPILE_OPTIONS "-Wno-pass-failed"
+)
+
```
MBkkt:
```suggestion
```

server/connector/duckdb_search_full_scan.cpp
```diff
@@ -894,9 +894,12 @@ void AccountAndWriteVirtualColumns(IResearchScanGlobalState& gstate,
   if (!gstate.ScanScore()) {
     return;
   }
-  // Scores arrive already mapped to the user-facing value (ApplyScoreEmit runs
-  // at the emit boundary) in the batcher's staged vector: reference it.
+  // The batcher stages raw "larger = nearer" scores, the space the pushed score
+  // filter was rewritten for; this is the one place they become user-facing, in
+  // place in the staged vector the output chunk then references.
   SDB_ASSERT(scores != nullptr);
+  ApplyScoreEmit(gstate, duckdb::FlatVector::GetDataMutable<float>(*scores),
```
MBkkt: ?

libs/iresearch/include/iresearch/formats/ivf/ivf_writer.cpp
codeworse: added assert for `group <= STANDART_VECTOR_SIZE`

server/connector/duckdb_search_full_scan.cpp
```diff
@@ -894,9 +894,12 @@ void AccountAndWriteVirtualColumns(IResearchScanGlobalState& gstate,
...
+                 num_rows);
```
MBkkt: it still doesn't fill right

libs/iresearch/include/iresearch/formats/posting/reader.hpp
```diff
@@ -211,6 +211,7 @@ inline size_t PostingsReaderBase::decode(const byte_type* in,
   }
   if (IndexFeatures::None != (features & IndexFeatures::Pay)) {
```
MBkkt: else if?

these features offs and pay not combinable, also maybe better to name `Vec`?

libs/iresearch/include/iresearch/formats/ivf/ivf_writer.hpp
```diff
@@ -111,6 +118,8 @@ class IvfTermReader final : public BasicTermReader, public TermPayloadWriter {
   bool _normalize;
   size_t _count;
   size_t _term_idx = 0;
+  bstring _carry;
+  size_t _carry_n = 0;
```
MBkkt: this feels meh

libs/iresearch/include/iresearch/formats/posting/reader.hpp
codeworse: added `else if`, renamed to `Vec`

libs/iresearch/include/iresearch/formats/ivf/ivf_writer.hpp
codeworse: what exactly? I think we still need some kind of a buffer for tail

libs/iresearch/include/iresearch/formats/posting/reader.hpp
```diff
@@ -123,7 +123,7 @@ inline void PostingsReaderBase::prepare(DataInput& in, const ReaderState& state,
 
   const bool needs_pay =
```
MBkkt, line 124: needs_vec

codeworse: But here we check for the payload file (if we need to write something into the `.pay`), not the renamed `Pay` enum

server/connector/duckdb_search_full_scan.cpp
codeworse: Now should be better. Actually, it is a fix for the streaming case for scorers, like:
```sql
SELECT id FROM t_idx
WHERE -(v <#> ARRAY[1.0, 0, 0, 0]::FLOAT[4]) <= 0.95
ORDER BY id;
```
So, the filter gets actual score and not the "larger is better" kind of scorer

libs/iresearch/include/iresearch/formats/posting/writer.hpp
```diff
@@ -349,9 +349,9 @@ inline void PostingsWriterBase::Prepare(IndexOutput& out,
   }
 
   // The ".pay" stream holds position-level offsets (IndexFeatures::Offs) and/or
-  // fixed-width per-document payloads (IndexFeatures::Pay, e.g. IVF codes).
+  // fixed-width per-document payloads (IndexFeatures::Vec, e.g. IVF codes).
   if (IndexFeatures::None !=
-      (state.index_features & (IndexFeatures::Offs | IndexFeatures::Pay))) {
+      (state.index_features & (IndexFeatures::Offs | IndexFeatures::Vec))) {
     if (IndexFeatures::None != (state.index_features & IndexFeatures::Offs)) {
       _pay.Reset();
```
MBkkt: .

codeworse: the if statement is kind of misleading. It means that we Reset _pay only on offsets, which was before. I've changed the if statements, so it would be more readily apparent

MBkkt (approved):

PR #1008, opened 2026-08-03
refactor: run the has_*_privilege functions through VariadicExecutor

gnusi:
## What

Sixteen `has_*_privilege` functions in `system.cpp` each hand-rolled the same loop. They now use `VariadicExecutor`.

**−545 / +298**, no behaviour change.

## The pattern being removed

Every one of them did this:

```cpp
duckdb::UnifiedVectorFormat tdata, pdata;
args.data[0].ToUnifiedFormat(args.size(), tdata);
args.data[1].ToUnifiedFormat(args.size(), pdata);
const auto* toid = duckdb::UnifiedVectorFormat::GetData<int64_t>(tdata);
const auto* priv = duckdb::UnifiedVectorFormat::GetData<duckdb::string_t>(pdata);
result.SetVectorType(duckdb::VectorType::FLAT_VECTOR);
auto* out = duckdb::FlatVector::GetDataMutable<bool>(result);
auto& validity = duckdb::FlatVector::ValidityMutable(result);
for (duckdb::idx_t i = 0; i < args.size(); i++) {
  auto ti = tdata.sel->get_index(i);
  auto pi = pdata.sel->get_index(i);
  if (!tdata.validity.RowIsValid(ti) || !pdata.validity.RowIsValid(pi) || !current) {
    validity.SetInvalid(i);
    continue;
  }
  bool is_null = false;
  bool r = HasTablePrivilegeByOidImpl(...);
  if (is_null) { validity.SetInvalid(i); } else { out[i] = r; }
}
```

becomes:

```cpp
duckdb::VariadicExecutor::Execute<bool, int64_t, duckdb::string_t>(
  args, result,
  [&](int64_t toid, duckdb::string_t priv) -> duckdb::optional<bool> {
    if (!current) {
      return duckdb::nullopt;
    }
    bool is_null = false;
    bool r = HasTablePrivilegeByOidImpl(...);
    return is_null ? duckdb::nullopt : duckdb::optional<bool>{r};
  });
```

The executor already does the format dispatch, selection vectors and validity, and its `optional<T>` overload turns the `is_null` out-parameter into a return value. It also brings the constant and dictionary fast-paths these loops skipped.

## What is deliberately left alone

**`CancelBackendsByPid`** keeps its loop. `pg_cancel_backend` / `pg_terminate_backend` have side effects, and collapsing a constant input to a single call — correct for a pure function — would cancel one backend and report that result for every row.

**Nothing else in the tree wants this.** I checked every SereneDB file with a `DataChunk` scalar function, and each remaining manual loop has a structural reason:

| file | why manual is right |
|---|---|
| `ts_lexize`, `ts_query`, `ts_highlight`, `ts_offsets`, `split_by_non_alpha`, `array.cpp` | build LIST/STRUCT results — child vectors and offsets written by hand |
| `vector.cpp` | ARRAY inputs, needs the child vector |
| `encode_key.cpp` | column count known only at runtime |
| `RegexpInstrFunction` | arity is 3 or 4 at runtime (`args.ColumnCount() == 4`) |
| `RegexpMatchFunction` | returns a LIST |

So `system.cpp` was the only real cluster.

## Testing

`any/pg` — **382/382 on both `pg-wire-simple` and `pg-wire-extended`**, zero failures.

The real gate is the 259-test `any/pg/rbac` subtree, run twice: once after converting a single function by hand to validate the shape, and again after all sixteen. Zero failures both times.

## Note on how this was done

Partly mechanically, and the first attempt was wrong in a way worth recording: a transformer that lifted only the code from `bool is_null` onward silently dropped a preceding `role_id = ResolveRoleOrPublic(...)` in `HasTablePrivilegeNameOid3Function`. It failed to compile, which is the lucky outcome — in RBAC code a dropped statement that still compiles is a security bug.

The version used here lifts the whole loop body verbatim and rewrites only four mechanical things (`ptr[sel_idx]` → `ptr`, `validity.SetInvalid(i); continue;` → `return nullopt;`, `out[i] = X;` → `return X;`, and drops the leading validity guard), then refuses to convert a function if anything it did not recognise is left behind. Three functions bailed out that way on the first run and were only converted once the transformer handled their shape properly.

Every converted function was still read individually, and the RBAC suite is what actually backs the claim.

### DIFF
--- a/server/connector/functions/system.cpp
+++ b/server/connector/functions/system.cpp
@@ -29,8 +29,8 @@
 #include <duckdb/catalog/catalog_entry/table_catalog_entry.hpp>
 #include <duckdb/catalog/catalog_search_path.hpp>
 #include <duckdb/catalog/entry_lookup_info.hpp>
-#include <duckdb/common/vector_operations/variadic_executor.hpp>
 #include <duckdb/common/vector_operations/generic_executor.hpp>
+#include <duckdb/common/vector_operations/variadic_executor.hpp>
 #include <duckdb/execution/operator/helper/physical_set.hpp>
 #include <duckdb/function/scalar_function.hpp>
 #include <duckdb/main/client_context.hpp>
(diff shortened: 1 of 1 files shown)


pashandor789 (approved):
Inspired by mega db

PR #1017, opened 2026-08-04
fix: give fuzzy expansion its own cap, sdb_levenshtein_max_terms

gnusi:
## Problem

`ts_levenshtein` took its `max_terms` from `sdb_scored_terms_limit`:

```cpp
FillByEditDistanceOptions(args, edit_opts);   // sets 64
edit_opts.max_terms = ctx.scored_terms_limit; // overwrites it with 1024
```

That setting is documented as *"the maximum number of terms to consider for
scoring in multi-term filters"*, and for the automaton and range filters that
is what it does: `SampledMultiTermVisitor` puts every visited term into the
query state and only samples which ones collect stats. The edit-distance
filter is different. With a non-zero limit `PrepareLevenshteinSegment` takes
the `TopTermsSelector` path, and only the selected terms are registered into
`query->State()`, so everything past the cut stops matching.

Three consequences:

- A knob sold as scoring cost changed result sets. On 2001 terms within
  distance 2 of `cat`, `SET sdb_scored_terms_limit = 50` returned 50 rows.
- The same predicate had two caps: 1024 standalone, 64 inside `ts_phrase`
  (which calls `FillByEditDistanceOptions` and never applied the override),
  and the phrase one ignored the session setting entirely.
- `ts_dict_agg` was truncated to the cap, so term enumeration returned a
  partial answer.

## Change

- New `sdb_levenshtein_max_terms`, default 64, is the fuzzy expansion cap.
  `0` removes the cap. It applies on both the standalone and phrase paths,
  since `FillByEditDistanceOptions` now takes the value from `FilterContext`
  instead of hardcoding one.
- `sdb_scored_terms_limit` is back to bounding scoring only and no longer
  touches which rows match.
- Term enumeration is exempt, decided per conjunct. `ClaimConjunct` zeroes the
  setting while claiming a conjunct that references an enumerated field (RAII,
  so a throwing claim still restores it),
  because the terms are the result of a ts_dict query and the automaton walk
  visits every accepted term either way, so capping would truncate the answer
  without saving traversal. The write goes to `ClientConfig::user_settings`
  directly rather than through a SET, so it does not trip the `set_callback`
  in `duckdb_client_state.cpp` that re-emits `ParameterStatus` and records
  transaction undo entries. A fuzzy predicate that only filters documents keeps
  the session cap, so the exemption follows the field being enumerated rather
  than the query kind:

```
SELECT unnest(ts_dict_agg(cat)) FROM ti
WHERE cat @@ ts_levenshtein('phon',2) AND body @@ ts_levenshtein('quikc',2)

│ ╭─ Levenshtein ───────╮  ╭─ Levenshtein ──────╮ │
│ │ Field: body(string) │  │ Field: cat(string) │ │
│ │ Max Terms: 64       │  │ Max Terms: 0       │ │
```

Kept the cap rather than removing it: it still bounds the postings unions and
per-term state on wide expansions, which is real work for document matching.
`LIMIT` pushdown was considered for the enumeration path and rejected for now,
since the selector orders by similarity with arbitrary tie-breaks, so
`ORDER BY similarity DESC, searches DESC LIMIT 1` (the spell-correction shape
in our docs) would pick an arbitrary member of a tied group.

## Verification

New `tests/sqllogic/sdb/pg/index/levenshtein_max_terms.test` pins the default,
the cap bounding rows, the closest-term survivor at a cap of 1, `0` removing
the cap, `sdb_scored_terms_limit` not moving the result set, the phrase path
honouring the setting and enumeration staying complete under a cap of 1.

Measured on 2001 distinct terms within distance 2 of `cat`, single segment:

| setting | matched rows | `ts_dict_agg` terms |
|---|---|---|
| default (64) | 64 | 2001 |
| 500 | 500 | 2001 |
| 0 | 2001 | 2001 |

Updated expectations: `show_all.test` (new setting), and the `Max Terms` line
in `filter_optimizer_lowering.test` (1024 -> 64) plus `ts_dict_aggs`,
`ts_dict_cartesian` and `ts_dict_where_deletes` (1024 -> 0, dict mode).

`serenedb-tests_connector` (489 tests) and the `sdb/pg/index` sqllogic subtree
pass. Those runs were done before rebasing onto current `main`; CI covers the
rebased tree.


MBkkt (commented):
i don't like that behavior of ts_dict will be different from normal execution

PR #1019, opened 2026-08-05
feat: Add `ts_phrase(..., slop := ...)` and `(...)::slop(...)`

aksel2904:
**Stacked on <https://github.com/serenedb/serenedb/pull/1014>** - engine-side sloppy phrase matcher.


server/connector/functions/search.h
```diff
@@ -33,9 +33,14 @@ inline constexpr std::string_view kTokenizerTypeName = "tokenize";
...
 inline constexpr std::string_view kModifierTSQueryTypeName = "TSQUERY_MODIFIER";
 
+// `EXPR::slop(N)` - composable phrase-slop modifier, parallel to
+// tokenize / boost. Only meaningful on a phrase; rejected elsewhere.
+inline constexpr std::string_view kSlopTypeName = "slop";
+
 // TSQUERY leaf constructors (unprefixed). Produce a TSQUERY value;
 // stubs throw at runtime -- the filter builder claims them at bind.
 inline constexpr std::string_view kTSQPhrase = "ts_phrase";
+inline constexpr std::string_view kTSQSloppyPhrase = "ts_sloppy_phrase";
```
MBkkt: maybe just argument (maybe named argument) for ts_phrase?
ts_phrase("a b", slop := 1)

server/connector/functions/search.h
MBkkt, line 1: needs to also extend lucene query parser. in lucene syntax, it's `~<some-number>`

tests/sqllogic/sdb/pg/index/sloppy_phrase.test
```diff
@@ -0,0 +1,331 @@
...
+# ----------------------------------------------------------------------
+
+query
+SELECT a FROM sloppy_idx WHERE b @@ ts_sloppy_phrase('quick', 0, 1, 'fox') ORDER BY a;
+----
+a
+1
+
+# ----------------------------------------------------------------------
+# Error paths.
```
Dronplane, line 178: Please add expected error text (in multiline mode) to the error cases like 
```
query error
SELECT id FROM docs_idx WHERE body @@ ts_phrase('fox') AND BM25(docs_idx.tableoid, 1.2, 0.75, 0.1) > 0;
----
db error: ERROR: No function matches the given name and argument types 'bm25(BIGINT, DECIMAL(2,1), DECIMAL(3,2), DECIMAL(2,1))'. You might need to add explicit type casts.
	Candidate functions:
	bm25(col0 BIGINT) -> FLOAT
	bm25(col0 BIGINT, col1 DOUBLE, col2 DOUBLE) -> FLOAT
```

That checks error generation + makes error comments unnecessary

server/connector/functions/search.h
Dronplane: +1 
FromSloppyPhrase ended up being almost full copy of   FromPhrase.
Let's drop it and stick to additional argument

tests/sqllogic/sdb/pg/index/sloppy_phrase.test
aksel2904: Done.

server/connector/functions/search.h
aksel2904: The grammar already parsed "..."~N into AddPhrase(value, slop) - the budget was parsed and silently dropped. Wired it to ByPhraseOptions::set_slop; the existing PhraseWithSlop tests now assert the budget, plus pins for bare ~, ~N^K, and a to_tsquery('"quick fox"~1') sqllogic case.

server/connector/functions/search.h
aksel2904: Done. Dropped ts_sloppy_phrase entirely and folded it into FromPhrase - slop is a named argument now: ts_phrase('a b', slop := 1), composable with gaps (ts_phrase('a', 1, 'b', slop := 2)). The binder appends the named vararg after the positionals and keeps the name as the child's alias; FromPhrase reads the aliased constant, the codec renders it back as slop := N and the structured-text whitelist admits the named shape (unknown names still fail to bind and fall back, so the ts_like(x := 'k') pin holds). ::slop(N) stays as the composable cast twin (tokenize/boost parallel); using both at once is rejected.

MBkkt (approved):

PR #1025, opened 2026-08-10
fix(docker): make the quick-start stack connectable

pashandor789:
## What

`curl -fsSL https://install.serenedb.com | sh` brings the stack up and then leaves you with no way in. Two independent blockers, both reproduced against the published `serenedb/serenedb:latest` (26.07.5):

**1. No route to a psql prompt.** A passwordless role is trusted only over loopback *inside* the container, no password is ever provisioned, and no `pg_hba.conf` is shipped. So `psql -h localhost` fails, and so does the compose `psql` service that the installer itself prints as the next step — it is a separate container, i.e. a network connection:

```
$ docker compose -f /tmp/serenedb-quick/docker-compose.yml run --rm psql
psql: error: ... FATAL:  password authentication failed for user "postgres"   (exit 2)
```

**2. No CA store.** `/etc/ssl/certs` is empty in the image, so attaching a TLS remote fails:

```
could not connect foreign server "play": Failed to verify SSL connection,
X509_v error: 20 unable to get local issuer certificate
```

Installing `ca-certificates` alone is not enough: the bundled OpenSSL resolves its default verify paths at build time and they do not match Ubuntu's, so `SSL_CTX_set_default_verify_paths()` looks in the wrong place. Copying the bundle to `/etc/ssl/certs` changed nothing; `SSL_CERT_FILE` fixed it.

## How

`SERENEDB_HOST_AUTH_METHOD`, mirroring `POSTGRES_HOST_AUTH_METHOD` from the postgres image. The entrypoint renders a `pg_hba.conf` from it and passes `--hba_config` (the flag already existed; nothing in the server changes). `trust` warns loudly on stderr and in the generated file, as the postgres image does.

### Accepted values

`trust`, `scram-sha-256`, `md5`, `password`, `reject` — the five methods serened evaluates
itself (`MethodExecutability` → `Native` in `server/network/pg/hba.cpp`). The hba parser
accepts nine more (`ident`, `peer`, `gss`, `sspi`, `pam`, `bsd`, `ldap`, `cert`, `oauth`)
but they are either unenforced or refuse the connection outright, so the entrypoint rejects
anything outside the five and says what is allowed.

That guard matters because the failure was silent otherwise: a typo makes serened log
`hba config file ... failed to parse ... -- keeping default` and fall back to requiring a
password, which surfaces to the user only as `password authentication failed`.

`install_docker.sh` defaults it to `trust` for a throwaway local stack; callers override it:

```sh
SERENEDB_HOST_AUTH_METHOD=scram-sha-256 curl -fsSL https://install.serenedb.com | sh
```

Plus `ca-certificates` and `ENV SSL_CERT_FILE` / `SSL_CERT_DIR` in the Dockerfile.

## Verified

Against the published image with the patched entrypoint mounted in:

| mode | result |
|---|---|
| `SERENEDB_HOST_AUTH_METHOD=trust` | connects from the host with no password; warning printed; hba generated |
| `SERENEDB_HOST_AUTH_METHOD=scram-sha-256` | passwordless connect refused; no warning |
| unset | no hba written, behaviour unchanged (backwards compatible) |

End to end through the fixed container — attach ClickHouse's public playground over TLS, index it, search it:

```sql
CREATE SERVER play FOREIGN DATA WRAPPER clickhouse_fdw
  OPTIONS (host 'play.clickhouse.com', port '9440', database 'default', user 'play', secure 'true');
CREATE VIEW hn AS SELECT id, title, "by", score, url FROM play.default.hackernews
  WHERE type = 'story' AND score > 500 LIMIT 1024;
CREATE INDEX hn_fts ON hn USING inverted(id, title en) INCLUDE ("by");
SELECT title, "by", score FROM hn_fts WHERE title @@ ts_phrase('rust') ORDER BY score DESC LIMIT 4;
```

```
 A half-hour to learn Rust                      | selvan       | 1444
 Tauri: An Electron alternative written in Rust | metalwhale   | 1113
 Async-await on stable Rust                     | pietroalbini | 1102
 Rust for Windows                               | dsr12        |  785
```

The Dockerfile change was validated by effect (mounted bundle + env), **not** by building the image — that needs the release tarball.

## Worth arguing about

Defaulting the quick-start to `trust` publishes a passwordless port, which in Docker's default networking is any other container on the same host. The postgres image made the opposite call: strict by default, `trust` opt-in. The quick-start is explicitly a throwaway demo stack, so trading that for zero friction seems right — but say so if you disagree and I will flip it to a generated password printed at the end.

Unrelated and untouched: the compose publishes `7890-7895:7890`, so the host port is not stable across runs (I saw 7890 through 7895 over several runs).

### DIFF
--- a/packages/docker/Dockerfile
+++ b/packages/docker/Dockerfile
@@ -40,11 +40,6 @@ RUN \
          /usr/share/doc/* /usr/share/man/* /usr/share/info/* \
          /usr/share/locale/* /usr/share/i18n/* /root/.cache /tmp/*
 
-# The bundled OpenSSL resolves its default verify paths at build time, which do
-# not match Ubuntu's store, so point it at the installed bundle explicitly.
-ENV SSL_CERT_FILE=/etc/ssl/certs/ca-certificates.crt \
-    SSL_CERT_DIR=/etc/ssl/certs
-
 VOLUME ["/var/lib/serenedb"]
 EXPOSE 7890
 
--- a/packages/docker/Dockerfile
+++ b/packages/docker/Dockerfile
@@ -16,8 +16,10 @@ RUN \
   rm /tmp/install.tar.gz && \
   # numactl: NUMA optimization for multi-socket systems
   # postgresql-client: psql for debugging, pg_isready for health checks
+  # ca-certificates: root store for outbound TLS (attaching a remote
+  #   ClickHouse/Postgres over TLS fails without it)
   apt-get update && \
-  apt-get install -y --no-install-recommends numactl postgresql-client && \
+  apt-get install -y --no-install-recommends numactl postgresql-client ca-certificates && \
   # Create non-root user for running serened
   groupadd -r serenedb && \
   useradd -r -g serenedb -d /var/lib/serenedb -s /usr/sbin/nologin serenedb && \
@@ -37,6 +39,11 @@ RUN \
   rm -rf /var/lib/apt/lists/* /var/cache/apt/* /var/log/* \
          /usr/share/doc/* /usr/share/man/* /usr/share/info/* \
          /usr/share/locale/* /usr/share/i18n/* /root/.cache /tmp/*
+
+# The bundled OpenSSL resolves its default verify paths at build time, which do
+# not match Ubuntu's store, so point it at the installed bundle explicitly.
+ENV SSL_CERT_FILE=/etc/ssl/certs/ca-certificates.crt \
+    SSL_CERT_DIR=/etc/ssl/certs
 
 VOLUME ["/var/lib/serenedb"]
 EXPOSE 7890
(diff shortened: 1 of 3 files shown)


packages/docker/Dockerfile
```diff
@@ -38,6 +40,11 @@ RUN \
          /usr/share/doc/* /usr/share/man/* /usr/share/info/* \
          /usr/share/locale/* /usr/share/i18n/* /root/.cache /tmp/*
 
+# The bundled OpenSSL resolves its default verify paths at build time, which do
+# not match Ubuntu's store, so point it at the installed bundle explicitly.
+ENV SSL_CERT_FILE=/etc/ssl/certs/ca-certificates.crt \
+    SSL_CERT_DIR=/etc/ssl/certs
+
```
MBkkt: this doesn't feel correct, how does clickhouse with same bundled ssl handles this?

MBkkt (approved):

PR #1030, opened 2026-08-12
fix: use unique secret name for parallel tests

Dronplane:
Fix secrets naming to avoid flaky failures in CI
Fix rounding in tests
Fix CI stalled containers cleanup sweep


MBkkt (approved):

tests/sqllogic/sdb/pg/explain/aggregate_input_dedup.test
```diff
@@ -205,7 +205,7 @@ g	cp	cr	c
 
 # Swapped arguments are a different input list, so each gets its own table.
 query
-SELECT covar_pop(DISTINCT x, y) AS xy, covar_pop(DISTINCT y, x) AS yx FROM agg_pair;
+SELECT round(covar_pop(DISTINCT x, y), 12) AS xy, round(covar_pop(DISTINCT y, x), 12) AS yx FROM agg_pair;
```
MBkkt, line 208: why?

Dronplane: It is not stable here due to order of records read and floating point rounding. It can get 16.250000000000004 in rare case. Same as in other tests in this file.

MBkkt (approved):

PR #1032, opened 2026-08-13
docs tests: reindex blocks for the view-index page

pashandor789:
Doc-example blocks for the new "Refreshing the index" section of the site's sql/indexes/inverted/views page (site PR follows; it renders these blocks by id, so this PR merges first):

- reindex_setup — parquet-glob view + index
- reindex_manual — add a file, SELECT * FROM serenedb_reindex(...) → delta row with counters, then a count through the refreshed index
- reindex_up_to_date — the no-op pass
- reindex_interval — CREATE ... WITH (reindex_interval), live ALTER INDEX ... SET retune, 0 = off

Also: 'control substitution on' (the file gains its first ${__TEST_DIR__} uses) and the second blank line that closes the multiline-error expectation above the new blocks.

Validated: the file passes on both engines against a fresh server.

### DIFF
--- a/tests/sqllogic/sdb/pg/site_docs/sql/indexes/inverted/views.test
+++ b/tests/sqllogic/sdb/pg/site_docs/sql/indexes/inverted/views.test
@@ -1,3 +1,5 @@
+control substitution on
+
 # DOCS_TEST: setup
 
 # DOCS_TEST_BODY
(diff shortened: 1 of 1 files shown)


pashandor789:
Folded into #1031.

PR #1037, opened 2026-08-16
fix: Don't drop zero-score hits from top-k

gnusi:
## Problem

`ORDER BY BM25(idx.tableoid) DESC LIMIT k` silently returned fewer rows than the same `WHERE` without the top-k — in the worst case, none at all.

The in-scan top-k collector seeded its threshold with `std::numeric_limits<float>::min()` — the smallest *positive* float (`1.18e-38`), not `lowest()` — while `LoserScoreCollector::TryPush` admits a hit only when `score > threshold`. A hit scoring exactly `0` could never enter the heap, not even while the heap was empty. The same value is the WAND skip bound (`SetScoreThreshold` rebinds the collector's pointer to the iterator's `ScoreThresholdAttr`), so `MaxScoreIterator::Split()` also demoted every zero-max-score branch to non-essential and skipped whole windows.

Zero is a legal score, and three ordinary setups produce it:

| case | before | after |
| --- | --- | --- |
| keyword column, no text dictionary, `LIMIT 10` | 0 rows | all matches |
| `sdb_scored_terms_limit = 0`, `LIMIT 3` | 0 rows | 3 rows, score 0 |
| `sdb_scored_terms_limit = 1`, `LIMIT 3` over 5 matches | 1 row | 3 rows |

The first needs no unusual settings at all: `BM25::PrepareScorer` returns `ScoreFunction::Default()` when the field carries no frequency feature, and that scores every document `0`.

## Fix

Seed with `lowest()`, which is what the value is supposed to mean — the k-th best score so far, or −∞ while fewer than k hits are held. As soon as `Push` fills the heap, `threshold = max(threshold, _root.score)` installs the real k-th score and both consumers behave exactly as before: pruning engages, and zero-score hits are rejected again because `0 > kth` fails once `kth >= 0`.

`CurrentPruneThreshold` carried the same "text scores are strictly positive" assumption in the streaming path, including a `nextafter(boundary, min())` that steps *up* rather than down for a `>=` bound. The two `ExecuteTopK` helpers in `doc_collector.hpp` are updated so the library's own top-k mirrors the engine.

## Verification

- New sqllogic coverage in `inverted_index_score.test`: zero-score top-k on a keyword column, `sdb_scored_terms_limit = 0` plain and top-k, the partial-cap case, and a k narrower than the match set.
- Green: full `sdb/pg/index` suite on both wire engines, `inverted_index_topk.test_slow`, `inverted_index_score_prune.test{,_slow}`, `inverted_index_lazy_seek.test_slow`, site_docs + es + tsquery suites, 3/3 score-prune recovery tests, 213/213 iresearch scoring gtests.
- No ranking change where the old code was already correct: on a 2M-doc `wiki_small` index, all 962 `TOP_100_DEBUG` result hashes are identical before and after.
- No measurable cost, hot = min of runs 2-3 over 962 queries × 20 reps:

| | before | after |
| --- | --- | --- |
| TOP_10 | 3.82 s | 3.77 s |
| TOP_100 | 4.70 s | 4.70 s |
| TOP_1000 | 6.87 s | 6.86 s |

### DIFF
--- a/libs/iresearch/include/iresearch/search/max_score_iterator.hpp
+++ b/libs/iresearch/include/iresearch/search/max_score_iterator.hpp
@@ -573,7 +573,7 @@ class MaxScoreIterator : public DocIterator {
   uint32_t _num_outer_windows = 0;
   doc_id_t _min_window_size = 1;
   ColumnArgsFetcher _fetcher;
-  score_t _next_threshold = std::numeric_limits<score_t>::min();
+  score_t _next_threshold = std::numeric_limits<score_t>::lowest();
   Attributes _attrs;
 };
 
--- a/libs/iresearch/include/iresearch/search/max_score_iterator.hpp
+++ b/libs/iresearch/include/iresearch/search/max_score_iterator.hpp
@@ -424,6 +424,7 @@ class MaxScoreIterator : public DocIterator {
                                          doc_id_t max) {
     _num_candidates += cand_docs.size();
     const score_t threshold = std::get<ScoreThresholdAttr>(_attrs).value;
+    SDB_ASSERT(threshold > 0);
 
     const auto first_required = _first_required;
     for (auto it = _first_essential; it != _itrs_sorted.begin();) {
--- a/tests/libs/iresearch/search/load_test.cpp
+++ b/tests/libs/iresearch/search/load_test.cpp
@@ -673,6 +673,23 @@ class LoadTest : public TestBase {
     }
   }
 
+  void SetUp() override {
+    TestBase::SetUp();
+    _null_sink = std::fopen("/dev/null", "w");
+    ASSERT_NE(nullptr, _null_sink);
+  }
+
+  void TearDown() override {
+    if (gExecutor) {
+      gExecutor->SetPrintSink(stderr);
+    }
+    if (_null_sink) {
+      std::fclose(_null_sink);
+      _null_sink = nullptr;
+    }
+    TestBase::TearDown();
+  }
+
   void RunAndValidate(std::string_view name) {
     const auto res_dir = resource("iresearch-load") / name;
     const auto queries_path = res_dir / "queries.json";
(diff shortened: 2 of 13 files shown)


MBkkt (changes requested):
It also was a few places in max_score_iterator and maybe other that relied on min instead of lowest.

I initially tried to make lowest everywhere, but it wasn't fully correct unfortunately. So I decided to make it later

libs/iresearch/include/iresearch/search/max_score_iterator.hpp
```diff
@@ -573,7 +573,7 @@ class MaxScoreIterator : public DocIterator {
   uint32_t _num_outer_windows = 0;
   doc_id_t _min_window_size = 1;
   ColumnArgsFetcher _fetcher;
-  score_t _next_threshold = std::numeric_limits<score_t>::min();
+  score_t _next_threshold = std::numeric_limits<score_t>::lowest();
```
MBkkt, line 577: it has sums/etc somewhere in the code, they will be broken?

libs/iresearch/include/iresearch/search/max_score_iterator.hpp
MBkkt, line 1: ProcessNonEssentialFromCandidates

tests/libs/iresearch/search/load_test.cpp
```diff
@@ -1120,6 +1124,12 @@ TEST_F(LoadTest, DisjunctionScoreAccuracy) {
 TEST_F(LoadTest, ReportIsOnlyAWayOfLooking) {
   ASSERT_NE(nullptr, gExecutor);
 
+  // Every hit printed, over this corpus, is tens of millions of lines in a CI
+  // log. These assertions are about the documents, never the text.
+  static std::FILE* const kNullSink = std::fopen("/dev/null", "w");
+  ASSERT_NE(nullptr, kNullSink);
+  gExecutor->SetPrintSink(kNullSink);
```
MBkkt: setup laod test + teardown for free fd

MBkkt (approved):

PR #1041, opened 2026-08-17
feat: Let a filter node override how its subtree is scored

gnusi:
`(<predicate>)::score(<scorer-expr>)` scores the wrapped subtree with a scorer of its own while the rest of the query stays on the ORDER BY scorer, so one scan can mix signals:

```sql
SELECT id, BM25(idx.tableoid) FROM idx
WHERE body @@ 'fox' OR (body @@ 'cat')::score('raw_boost()');
```

The `fox` branch stays on BM25; the `cat` branch scores a flat 1.

`::score(NULL)` goes further and excludes a subtree from scoring: it still selects rows but contributes nothing, and asks for no index features, so its postings are read without frequency or norms -- the shape Elasticsearch spells as a `filter` clause.

```sql
SELECT id, BM25(idx.tableoid) FROM idx
WHERE body @@ 'fox' OR (body @@ 'cat')::score(NULL);
```

Documents matching only through the unscored branch come back with 0, and the rest score exactly as if that branch were not there.

**This makes #1037 a prerequisite.** An unscored branch turns 0 into a legitimate score, and on `main` the top-k collector seeds its threshold above zero (`duckdb_search_full_scan.cpp:1649`), so `ORDER BY score ... LIMIT k` silently drops rows that match only through such a branch -- verified on this branch, where doc 2 vanishes from the top-k. #1037 fixes that seed; the sqllogic here deliberately covers the streaming path only, with a note, until it lands.

## More examples

All values below are asserted in `inverted_index_score.test` against three documents -- `1: 'fox fox fox dog'`, `2: 'cat cat'`, `3: 'fox cat'` -- indexed with frequency, position and norm.

### Every scoring leaf honours the override, not just terms

A phrase leaf resolves its scorer from the node it was collected for, so `raw_tf()` gives the phrase frequency:

```sql
SELECT id, BM25(idx.tableoid) FROM idx
WHERE (body @@ ts_phrase('fox'))::score('raw_tf()');
--  1 | 3
--  3 | 1
```

n-gram similarity publishes its match ratio as a per-document boost, and `constant(K)` multiplies by it -- so a document matching one n-gram of two scores half:

```sql
SELECT id, BM25(idx.tableoid) FROM idx
WHERE (body @@ ts_ngram('fox dog', 0.5))::score('constant(5)');
--  1 | 5      -- both n-grams
--  3 | 2.5    -- 'fox' only
```

Geo exposes no frequency, so the overrides that mean anything there are the frequency-independent ones. From `geo_search.test`:

```sql
SELECT id, BM25(geo_idx.tableoid) FROM geo_idx
WHERE (ST_Intersects(geo, '<polygon>'))::score('constant(9)');
--  1,2,3,6,7 | 9
```

### Branches keep their own scorers side by side

```sql
SELECT id, BM25(idx.tableoid) FROM idx
WHERE (body @@ ts_phrase('fox'))::score('raw_tf()')
   OR (body @@ 'cat')::score('constant(10)');
--  1 | 3     -- phrase only
--  2 | 10    -- cat only
--  3 | 11    -- 1 + 10
```

### An override on a group reaches the leaves under it

This is the case that previously passed validation -- the wrapped node is an `Or` -- and was then silently ignored by the phrase leaf:

```sql
SELECT id, BM25(idx.tableoid) FROM idx
WHERE (body @@ ts_phrase('fox') OR body @@ 'cat')::score(NULL);
--  1 | 0
--  2 | 0
--  3 | 0
```

### `::merge` picks how a boolean node combines its branches

Default is sum; `max` takes the best branch instead of adding them, which is what you want when branches are alternative spellings of one intent rather than independent evidence:

```sql
SELECT id, BM25(idx.tableoid) FROM idx
WHERE (body @@ 'fox' OR body @@ 'cat')::merge('max');
--  1 | 0.6671
--  2 | 0.6951
--  3 | 0.5235   -- max(fox, cat), not 1.0471
```

It binds to its own node, so nesting is meaningful -- inner `max` still lets the outer node add `dog` on top, while `max` on the outer node takes the best of all three:

```sql
WHERE ((body @@ 'fox' OR body @@ 'cat')::merge('max') OR body @@ 'dog')
--  1 | 1.4814   -- max(fox, cat) + dog

WHERE ((body @@ 'fox' OR body @@ 'cat') OR body @@ 'dog')::merge('max')
--  1 | 0.8143   -- dog alone wins
```

### The two compose

A per-branch scorer under a group policy -- 10 beats cat's 0.5235, so the constant survives the max:

```sql
WHERE ((body @@ 'fox')::score('constant(10)') OR body @@ 'cat')::merge('max')
--  1 | 10
--  2 | 0.6951
--  3 | 10
```

An unscored branch contributes nothing to compare, so the other branch wins by default:

```sql
WHERE ((body @@ 'fox')::score(NULL) OR body @@ 'cat')::merge('max')
--  1 | 0
--  2 | 0.6951
--  3 | 0.5235
```

And a boost multiplies what the winning branch scored, not what the sum would have been:

```sql
WHERE (body @@ 'fox' OR body @@ 'cat')::boost(2.0)::merge('max')
--  3 | 1.0471   -- 2 x 0.5235, not 2 x 1.0471
```

## What was already there

The score tree is per-node shaped today: `PrepareCollector::GetScorer()` is virtual, `CompoundCollector::Finish` builds a `StatsBuffer` **tree** with a scorer per node, `boolean_query.cpp ExecuteChild` threads `stats.Child(i)` per child, and `TermQuery::Execute` already read its own node's scorer to choose index features. Missing were a place to keep the override that survives the pipeline, optimizer support, and a surface.

## Where the override lives

On the `Filter` base, next to the boost. `Filter::MakeCollector` is non-virtual and resolves `_scorer ? _scorer : scorer` before delegating to a protected `MakeCollectorImpl`, so neither a filter kind nor a caller can forget to substitute -- the virtual only ever sees the winner. `FilterWithBoost` is folded into `Filter`, which had no other purpose, and its `BoostImpl` virtual goes with it.

Two placements were tried and rejected on the way, both worth recording:

- **`PostingCookie`** is per *term*, so a per-node property got duplicated across every term of a multiterm node.
- **`ScoreAdapter`** is transient: `operator DocIterator::ptr&&()` unwraps it when a composite collapses to one child, silently dropping the override. `(body @@ 'cat')::score('raw_tf()') OR body @@ 'nothingmatches'` scored **0** under that placement, because the collector had already built the node for `raw_tf` while scoring fell back to BM25 reading a RawTF stats blob.

An earlier revision wrapped the subtree in a transparent `ScorerFilter` node instead. That node taxed every optimizer rule with an unwrap, and it is gone -- along with 23 unwrap sites, which also un-blocked the prefix/levenshtein and complementary-range fusions a transparent node had silently prevented.

Excluding a subtree needs no third pointer state. `::score(NULL)` points the override at `irs::Unscored`, whose `PrepareScorer` returns a default (zero) score function and which declares `IndexFeatures::None`. A null scorer already means unscored downstream -- `Exclusion` leaves its excluded branch that way (`boolean_filter.cpp:128`) -- so this reuses the existing meaning instead of overloading `nullptr`, which stays "inherit from the parent".

## Optimizer

Rules rebuild nodes and re-apply only the state they know about, so the override travels three different ways:

- **Carried** alongside `boost` at the 17 single-node rebuild sites across `boolean_rules`, `lowering_rules`, `negation_rules`, `range_rules`, `terms_rules`.
- **Folded** when a parent collapses into its child (`FoldScorer`): the child's own override wins for its subtree, so it only inherits the parent's when it has none.
- **Refused** where siblings fuse into one node. `ByTermsRule` fuses sibling `ByTerm`s into a single `ByTerms`, and boost survives only because `ByTermsOptions` has a per-term boost slot -- nothing can hold two scorers, so differing overrides now block the fusion. `OrAcceptorFusionRule` already refused to fuse differing boosts when scored, so the scorer check sits beside it. `AndAcceptorFusionRule` is disabled for scored queries and needed nothing.

The preservation tests were written before the rule edits; three of five failed on the unmodified code, including the fusion case that dropped **both** overrides and produced wrong scores rather than an error.

## Surface

The modifier is a **BLOB**, so its value type alone tells it apart from tokenize's VARCHAR. A named `properties` entry would not work: `ExtensionTypeInfo::Equals` compares only keys present in *both* types and only `min(len)` modifiers, so a properties-only `::score` would have compared **equal** to `::tokenize` and risked cast elision on a chain.

The expression is parsed by the existing `catalog::ParseScorerExpression`, so anything `optimize_top_k` accepts works here. Parsed scorers are owned beside the filter they belong to, since nodes point into them non-owningly and the filter is shared into the offsets bind data.

Every leaf that scores through a `Scorer` now carries the override, phrase, ngram, geo and match-all included -- see the examples above. A leaf resolves it from the `ScoreSource` (`search/scorer.hpp`) its `Execute` was handed: the node's scorer paired with the stats collected for it, produced only by `StatsBuffer` and returned by value, so the two halves cannot be mismatched or drawn from different buffers. The postings path receives the scorer once per `Iterator()` call through `IteratorFieldOptions::scorer` and pairs it with each term's `PostingCookie::stats`. `PrepareScoreContext` loses its own `scorer` field, so a leaf has exactly one way to resolve its scorer.

What stays rejected is what genuinely cannot honour an override, each with its own message rather than one generic one: a vector-distance predicate (the distance *is* the score, and it passes no stats), a bare negation (`Not` collects its branch with no scorer at all), same-position matching (produces no score under any scorer), and a wildcard n-gram match (`Constant(kNoBoost)` by construction). `ByEditDistance`, `Empty` and `MixedBooleanFilter` were missing from the old whitelist and now work.

## Score-bound pruning

Per-block score bounds are written with the index's `optimize_top_k` scorer, so
they are not upper bounds for an overridden subtree -- pruning against them could
drop matching documents. `ExecutionContext::prune_scorer` replaces the old
`score_prune` flag and names the scorer the bounds were validated against, so
`MayScorePrune` lets a node prune only when `stats.GetScorer()` is that scorer.

Separately, the gate on when pruning may run at all was too strict: it required
the query's scorer to *equal* the index's, so `bm25(1.2, 0.75)` bounds went unused
by a `bm25(2.0, 0.5)` query. A bound is a max-freq / min-norm summary that each
scorer scores for itself, so what has to match is the **bound type**, not the
options:

| bound type | index `optimize_top_k` | queries it can serve |
| --- | --- | --- |
| `MaxFreq` | `bm25(k1, b=0)` | `bm25(k1', b=0)`, `tfidf(with_norms=false)` |
| `DivNorm` | `bm25(k1, b=1)` | `bm25(k1', b=1)`, `tfidf(with_norms=true)` |
| `MinNorm` | `bm25(k1, b)` | `bm25(k1', b')` -- any parameters |
| `None` | `bm25(k1=0)` (BM1) | never prunes; BM1 writes no bounds |

Only these two scorers implement the bound API at all; every other scorer inherits
the null default. `Scorer::compatible` reports looser cross-type degrees, but they
are unusable: the stored bytes carry no layout tag and nothing in the formats layer
records which writer produced them, so the query scorer's bound source must decode
the layout the index wrote. The gate is therefore bound-type equality, and
`compatible` stays unused.

`ScorerOptions::BoundType()` derives the type from the options via the `Owner`
static each scorer opts into, so the check needs no scorer instance; `EXPLAIN` grows
a `Bounds:` row naming the writing scorer when it differs from the query's.

### Posting stream alignment

For postings of `1 < docs_count < kBlockSize` the persisted bounds sit inline in the doc
stream, and the read cursor only lands on the postings once they are skipped
(`SkipScoreBounds`). The ordinary posting iterator never reads them, so it must always
skip; only the dedicated pruning iterator consumes them. `PostingIteratorImpl::Prepare`
therefore takes the node's scorer *before* its `score_prune` flag, and that flag stays
defaulted at the non-pruning call sites. Suppressing the skip there misaligns the stream
and bound bytes decode as doc deltas -- which surfaces as a `FormatTraits128::ReadTail`
assertion in the load tests, and as a top-k query admitting a document that matched only
one of two conjuncts while carrying another document's score.

## Tests

- `per_node_scorer_test`: per-branch override on term, phrase, ngram and match-all
  leaves; an override inherited from an ancestor group reaching a phrase leaf; and
  an `Unscored` oracle -- with `irs::Unscored` as the override every honouring leaf
  must score exactly `0.0f`, which is a value-free assertion, so adding a leaf kind
  is one table row.
- `optimizer_scorer_test`: collapse folds the parent's override; collapse keeps the
  child's own; fusion refused on differing overrides; fusion allowed and carried
  when they agree; fusion unaffected when no override exists.
- `geo_filter_test.per_node_scorer_override`: `raw_boost` scores the points inside
  the polygon 1, an `Unscored` override scores the same points 0 -- so the override
  reaches the geo leaf and the match set is untouched.
- `inverted_index_score.test`: hand-checkable values for every supported leaf
  (`constant(42)` on a phrase, `constant(7)` on a sloppy phrase, `constant(5)` on
  `ts_ngram` where the match ratio halves it to 2.5), per-branch scorers side by
  side, an override on a group, `::merge` alone and composed with `::score`,
  `::boost` and an unscored branch, the two EXPLAIN plans, and a pruning oracle that
  reruns each ranking with `sdb_disable_top_k_optimization` forced on.
- `inverted_index_score_prune.test`: cross-parameter pruning -- the EXPLAIN showing
  `optimized` plus the `Bounds:` row, and two 50k-row oracles (single term and
  `ByTerms` disjunction) proving a `bm25(1.2, 0.75)` bound read by a
  `bm25(2.0, 0.5)` query drops no documents.
- `tsquery_params.test`: modifiers on bound parameters, per-element list modifiers,
  and the rejection timing -- a malformed `::score` is caught at `EXECUTE` rather
  than `PREPARE`, because the scorer text is parsed exactly once where the filter is
  built. Validating at bind as well would parse the same string twice on every
  query; the test records that trade so it is not silently undone.
- `inverted_index_wildcard_score.test` and `geo_search.test` cover the leaves whose
  overrides are rejected or degenerate.
- `score_bound_dominance_test`: folds a block through `FreqNormProducer` and asserts the
  stored (freq, norm) pair outscores every member, over a block whose minimum length
  falls below its maximum frequency.
- `ts_query_codec_test`: a round trip over all six tsquery fields, plus the two render
  paths agreeing on `::merge` -- the pgwire row serializer read five of six struct
  children while the `TSQUERY -> VARCHAR` cast kept all six.


server/connector/duckdb_table_function.h
```diff
@@ -166,6 +165,7 @@ struct SereneDBScanBindData : public duckdb::FunctionData {
   // filter, scorer, offsets, ts-dict requests). Every scan bound through this
   // table function is a search scan, so `snapshot` is always set.
   std::shared_ptr<irs::Filter> stored_filter;
+  std::shared_ptr<std::vector<std::unique_ptr<irs::Scorer>>> filter_scorers;
```
MBkkt: .

server/connector/duckdb_table_function.cpp
```diff
@@ -368,12 +369,16 @@ static duckdb::vector<duckdb::column_t> SereneDBScanGetRowIdColumns(
...
   }
-  // TODO(mbkkt) use compatibility instead
   const auto& topk = index->GetTopKScorer();
-  return topk && topk == scorer;
+  if (!topk) {
+    return false;
+  }
+  const auto bound = topk->BoundKey();
+  return bound.first != irs::Scorer::ScoreBoundType::None &&
+         bound == scorer->BoundKey();
```
MBkkt: compatible call?

server/connector/duckdb_search_full_scan.hpp
```diff
@@ -178,6 +178,8 @@ struct IResearchScanGlobalState : public duckdb::GlobalTableFunctionState {
   float score_static_floor = std::numeric_limits<float>::lowest();
   bool score_prune_streaming = false;
 
+  bool score_prune = false;
```
MBkkt: .

server/connector/functions/ts_offsets.h
```diff
@@ -46,6 +46,7 @@ struct OffsetsBindData final : duckdb::FunctionData {
 
   size_t limit = 0;
   std::shared_ptr<irs::Filter> stored_filter;
+  std::shared_ptr<std::vector<std::unique_ptr<irs::Scorer>>> filter_scorers;
```
MBkkt: .

server/connector/functions/ts_common.hpp
```diff
@@ -61,6 +62,7 @@ struct FilterContext {
   duckdb::ClientContext& client_context;
   uint32_t scored_terms_limit = 1024;
   uint32_t levenshtein_max_terms = 64;
+  FilterScorers* scorer_sink = nullptr;
```
MBkkt, line 65: .

server/catalog/persistence/scorer_options.h
```diff
@@ -67,7 +101,6 @@ struct ScorerOptions {
   }
 
   // EXPLAIN-friendly spelling, e.g. `bm25(k1=1.2, b=0.75)`.
```
MBkkt:
```suggestion
```

MBkkt (approved):

libs/iresearch/include/iresearch/search/tfidf.cpp
```diff
@@ -208,11 +210,12 @@ ScoreBoundSource::ptr TFIDF::PrepareScoreBoundSource() const {
...
 }
 
-Scorer::ScoreBoundType TFIDF::GetScoreBoundType() const noexcept {
-  if (_normalize) {
-    return ScoreBoundType::DivNorm;
-  }
-  return ScoreBoundType::MaxFreq;
+bool TFIDF::Compatible(const ScorerOptions& persisted) const noexcept {
+  return irs::BoundTypeOf(persisted) == BoundTypeOf({.with_norms = _normalize});
+}
```
MBkkt, line 215: this scorer is persisted, options are user input.

PR #1047, opened 2026-08-21
feat: SQL/PGQ property graph queries

gnusi:
Vendors cwida/duckpgq-extension as the third_party/duckpgq submodule (serenedb fork, branch 'serenedb') and statically links it like the other DuckDB extensions. SQL/PGQ grammar and parse nodes live in the duckdb fork's PEG parser; duckpgq's parser_override (always on via allow_parser_override_extension=STRICT) parses with the core parser and wraps PGQ statements for its binder. Property-graph metadata persists in <session db>.public.__duckpgq_internal.

duckpgq's internal metadata connections are blessed through the new DBConfig::internal_connection_factory: CreateBlessedInternalConnection derives a ConnectionContext from the initiating session (same user and database, fresh catalog snapshot) so extension-side helper connections can reach the SereneDB catalog.

The DDL and pattern grammar follows PostgreSQL 19's SQL/PGQ surface: NODE/RELATIONSHIP as synonyms for VERTEX/EDGE, an element KEY clause (validated against the table's primary key at bind), REFERENCES without a column list (resolves to the referenced vertex table's primary key), LABEL and PROPERTIES in either order, DEFAULT LABEL, and label-only / anonymous pattern elements (`(IS person)`, `-[IS knows]->`) with deterministic generated bindings. Beyond PG19's fixed-length patterns, quantifiers, ANY SHORTEST + path_length, and search/vector/scorer integration work over both wire protocols.

Table-backed inverted-index scans project the store rowid, which makes index-relation property graphs path-finding capable: @@ predicates run natively inside MATCH patterns as IRESEARCH_SCANs, including under ANY SHORTEST and quantifiers, and BM25 binds through pattern variables. The fork also fixes ErrorData adopting an unrelated in-flight exception (which masked real binder errors inside PGQ statements), duckpgq's cross-connection registry and CSR races, and CSR sizing under deleted rows.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -221,3 +221,7 @@
 	path = third_party/libxml2
 	url = https://github.com/serenedb/libxml2.git
 	shallow = true
+[submodule "third_party/duckpgq"]
+	path = third_party/duckpgq
+	url = https://github.com/serenedb/duckpgq.git
+	shallow = true
--- a/scripts/check_submodule_pointers.py
+++ b/scripts/check_submodule_pointers.py
@@ -29,6 +29,7 @@ VERSION_BRANCH = re.compile(r"^v20\d{2}\.\d{2}\.\d{2}$")
 CHECKED = {
     "third_party/duckdb": VERSION_BRANCH,
     "third_party/duckdb_iceberg": VERSION_BRANCH,
+    "third_party/duckpgq": re.compile(r"^serenedb$"),
 }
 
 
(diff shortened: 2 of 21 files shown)


gnusi:
because CEO asked me to

MBkkt (commented):
Worth to check:
- About grammar approach and grammar itself: https://github.com/rahul-iyer/duckdb-gql
- About graph algorithms: https://github.com/CogitatorTech/onager , https://github.com/alitrack/duckdb_petgraph , https://github.com/DataZooDE/anofox-similarity
- Something strange: https://github.com/DataZooDE/erpl-web

Issue #198, opened 2026-02-03
Optimization of search delete mask storage and access


MBkkt:
### Background

In **SereneDB**, we manage search data in immutable segments. To handle deletions without rewriting entire segments, we use a `deleted_mask`. Currently, this is implemented using `absl::flat_hash_set<doc_id_t>` in memory and a simple list of `varint` values on disk.
These structs are immutable right now. Instead new deletes accumulated in bitset for new segments (this is probably best solution) and in hashset for old segments (this is probably not best option).

While functional, this approach is sub-optimal across different data densities and access patterns. 
We are looking to transform this into a polymorphic, highly optimized subsystem.

### The Challenge

The goal is to design a flexible interface for "liveness tracking" and implement a suite of storage strategies that adapt to the density of deletions. We need to answer the question: *Which data structure wins at 0.1%, 1%, and 10% deletion rates?*

#### 1. Interface Abstraction

Replace the hardcoded `flat_hash_set` with an abstract interface (e.g., probably `AliveDocs` but maybe `DeletedDocs`).

* start with `Get` and `Count`
* maybe iteration interface
* maybe special treatment in negation/intersection/etc

#### 2. Benchmarking Framework (Core Requirement)

We cannot optimize what we cannot measure. You will build a benchmark suite to evaluate:

* **Write Throughput:** Performance of marking docs as deleted.
* **Read Latency:** Point lookups during query execution.
* **Scan Performance:** Overhead during sequential scans (e.g., bitwise negation performance).
* **Memory/Disk Footprint:** Space-efficiency across different cardinalities.

#### 3. In-Memory Implementations

* **Dense Bitset:** For high-density deletion scenarios, probably better to have alive bitset in such cases.
* **Optimized HashSet:** For very sparse deletions (re-evaluating `absl::flat_hash_set`).
* **Sparse Bitset:** Utilizing our existing `SparseBitmap` implementation (it's roaring like), also you can try roaring itself. Or just simple sparse bitset implementation, e.g. see lucene (2^14 docs per block, always bitset blocks) https://github.com/apache/lucene/blob/7ebdb9316e5ccd6940c195f60df9e08f0d3a178a/lucene/core/src/java/org/apache/lucene/codecs/lucene90/Lucene90LiveDocsFormat.java#L141

#### 4. On-Disk Formats

* **Raw Bitsets:** For dense segments (high deletion rate, e.g. 50%) where 1 bit per document is the best.
[...]

PR #1056, opened 2026-08-24
perf: replace DocIterator with concrete root and clause APIs

MBkkt:
`DocIterator` is gone. The read side is a set of small, purpose-built APIs instead of one interface asked to be everything at once, the boolean filter layer is one node type instead of six, and scoring matches Lucene where it used to differ. On the wiki benchmark every command is faster than main on aggregate with identical documents and counts, and main already led Lucene 10.5 and tantivy there; the gap widens, measured without LTO or PGO. A few dozen individual queries are still slower than main and are being worked down.

## Why

The old read side had one interface, `DocIterator`, with 11 virtuals and ~35 implementations, and it was asked to be five unrelated things: a doc-at-a-time cursor, a run producer, a bitset accumulator, a top-k drain, and a counter. Each drain had its own contract written in comments (`FillBlock` needs priming, `EmitDocs` must not be primed, `count()` destroys state), illegal combinations were expressible and caught by runtime asserts if at all (a per-term pruning iterator under a summing disjunction shipped), the same block walk was copied seven times, and the shape was re-resolved per call. Block-max pruning was reachable from one construction site inside the postings reader, so it could never span fields.

A cursor is also the wrong primitive for most of the work. A disjunction wants to or documents into a window, a conjunction wants to probe a candidate, a phrase wants positions, a count wants a number. Forcing all of them through `advance()`/`seek()` paid a virtual call per document where one per 4096 would do, and carried score state into paths that never score.

### Roots: what a consumer asks for

| API | entry point | contract |
|---|---|---|
| `count::Root` | `uint64_t Run()` | a number; no score state anywhere in the tree |
| `docs::Root` | `uint32_t Run(doc_id_t* out, uint32_t capacity)` | ascending doc ids, resumable batches |
| `scored::Root` | `uint32_t Run(doc_id_t*, score_t*, uint32_t capacity)` | docs and scores, resumable batches |
| `top::Root` | `void Run(LoserScoreCollector&)` | run to completion into one collector spanning every segment; may prune |
| `offsets::Root` | `uint32_t Run(doc_id_t doc, std::span<Range>)` | replays the positional match on an already-matched document and materialises its offsets (highlighting) |
| `count::TermCounts` | `Count(term)` / `Any(term)` | ts_dict: the WHERE and the removals are folded once per segment into a lazily filled bitset, then asked per dictionary term |
| `TermPostings : lead::Node` | `Advance()`, `GetFreq()`, `Positions()` | what flush produces and merge consumes; `PostingsStream` replaces the read-side posting iterators on the write path |

Roots are siblings, never layers: no root drives another root, and a clause is never driven by a root. Each has an empty root for a query that matches nothing, so a null plan means exactly one thing: this query kind has no plan for this consumer (kNN outside `top::`). Every `QueryBuilder` has `PlanCount` / `PlanDocs` / `PlanScored` / `PlanTop` plus the three clause forms below, so there is no generic fallback and a missing shape fails loudly instead of running slowly.

### Clauses: what a node asks of its children

| API | entry point | used by |
|---|---|---|
| `lead::Node` | `Value()`, `Advance()`, `Seek(target)` | the head of a sparse conjunction or exclusion, positional approximations, two-phase shapes |
| `probe::Node` | `Probe(target)`: the first document at or past `target` the clause holds | the tail of a sparse conjunction, the excludes of a sparse exclusion, optional buckets |
| `fill::Node` | `FillOr` / `FillAnd` / `FillAndNot(min, max, mask)`, `Fill(..., scores)`, `Folded()` | window shapes: disjunction, dense conjunction and exclusion, thresholds |

A probe answers with a document rather than a bool, so a miss is a jump. Three fill operations rather than one because `~(x ∪ y) = ~x ∩ ~y`: an exclude clears its own bits in place as it decodes and never needs scratch, while a conjunction clause needs its operand materialised. `Folded()` hands out a segment-wide bitset once a set of terms is dense enough that re-entering it per window costs more than building it once.

Scored and unscored are separate types end to end. A `count`/`docs` tree carries no scorer, stats, boost or score buffer: not in members, not in `sizeof`, not in work. On the scored side a clause is one of three implementations chosen at plan time: unscored, constant (the docs node plus one number), or per-document. None of them wraps another.

### Execution model

Two modes, chosen at plan time, never at run time:

- **Document at a time** where the shape needs a current document: sparse conjunction and exclusion (the rarest clause leads, the rest are probed rarest first), probed disjunctions (asked densest first), phrases, n-grams, two-phase shapes (geo, wildcard n-gram, nested).
- **Window at a time** otherwise: 4096 documents as 64 words. Leaves write into the window, parents combine with or / and / and-not, the root turns it into an answer: a popcount, a left-packed run, a scored drain. Clearing rides on the pass that already holds the word. A threshold (`min_should_match`) tallies with bitplanes at k=2 and a counter array above.

Leaves stay concrete wherever a container's members share one layout, which covers a disjunction of terms, a conjunction of terms, and every phrase, so most trees over terms have no virtual call per document at all. Erasure appears only where members genuinely differ in type, and then it is crossed once per window or once per probe. Small arities (a tail of one, phrase slots of two or three) are compiled out; wider ones are dynamic. Containers are fixed arrays built in place, because a leaf holds buffers that point into itself and is not movable.

### The plan layer decides three things

Erasure or not, which container (and at what arity), and bitset vs window vs sparse. Everything else is settled before a plan is built:

- **Filter layer** (`irs::Filter`, `Optimize`): meaning-level rewrites with no index access.
- **Query layer** (`PrepareSegment` -> `QueryBuilder`): everything that needs the term dictionary. An absent term is `Empty`, a term covering the segment is `All`, estimates are real, clauses are sorted, nested booleans of the same operation are flattened, postings are deduplicated by identity, and `All` / `Empty` are absorbed and propagated upward. `QueryKind` is the real type, never a semantic tag.
- **Plan layer** (`Make*`): mechanical. It never resolves a field, walks children to discover a fact, or returns null for "no results".

`count/` additionally owns the counting algebra, because it is the only consumer that can answer with a subtraction: an exclusion whose include is everything is `|segment| - |disjunction of the excludes|`; `|a ∧ b| = |a| + |b| - |a ∨ b|` when both clauses are wide; `|a ∨ b| = |a| + |b| - |a ∧ b|` when they are skewed. A term's count is read off its metadata and opens no posting list.

## Filter and query layers

**One `BooleanFilter`** with `Must` / `Should` / `MustNot` buckets, `min_should_match` and a merge type replaces `And`, `Or`, `Not`, `Exclusion`, `MixedBooleanFilter` and `ByTerms`. Terms are inline `TermClause`s in a sorted vector, one allocation per bucket instead of a `std::set` node per term, deduplicated on insert with boosts merged; a `ByTerm` handed to `Add` is absorbed into the bucket, so the optimizer never sees one inside a node. Negation is a bucket, so De Morgan and double negation are bucket splices and there is no negation rule. The twenty boolean rules over six node types become eight over one (`boolean_flatten`, `_dedup`, `_absorb`, `_single_clause`, `_min_should_match`, `_normalize_terms`, `_null_marker`, `or_acceptor_fusion`); the lowering, range and Levenshtein-prefix rules stay. `IRESEARCH_DISABLE_RULES=a,b` turns rules off by name, which is how two shapes answering the same query get compared against each other.

**`PrepareSegment` builds one `BooleanQuery`** whose buckets are sorted arrays of prepared postings, absorbed `all` clauses and child queries. `must` is ascending by estimate so the rarest leads; `should` and `must_not` are descending so the densest answers first. Each bucket carries a uniformity (`Mixed` / `Constant` / `Scored` / `Bounded`), the only aggregate the plan reads: whether a scored container can be devirtualised and whether block-max bounds are available. Simplification happens once, while the node is built, over children that are already prepared.

**Statistics are a pointer into an arena.** A clause carries `{stats, scorer}` directly; the coordinate-based navigation and the per-thread collector trees with their recursive merges are gone. `MakeCollector` runs once per query with the thread count, each thread writes its own counters, and `Finish` reduces them once per slot.

**The Lucene parser** builds a `BooleanFilter` directly and follows Lucene's semantics: a query with no positive clause (`-hello`) matches nothing, and a threshold beside a required clause (`(+b a c)@2`) is a matching constraint, `b AND 2-of{a, c}`, as `setMinimumNumberShouldMatch` documents. SQL keeps its own reading of `NOT` (the complement) in the connector, where it belongs: a filter-layer rule cannot tell the two surfaces apart.

## Pruning

- **Block-max conjunction (WAND) for a conjunction of terms**, which main did not have. Score-first, as Lucene's `BlockMaxConjunctionBulkScorer`: the window is the lead's block, the other clauses are asked largest-max-score first so the suffix-summed allowance falls fastest, documents that cannot reach `threshold - remaining` are dropped before the next clause seeks, and the window is narrowed to the tightest clause block when there are four or more clauses. Score-first scores every lead document, so it is planned only where the clauses are estimated to agree on far more than k documents (a geometric-mean estimate against `k * 30`, `k * 75` for pairs) and falls back to the unpruned sparse conjunction otherwise. On the benchmark's 300 `TOP_100` conjunctions the pruned plan scores 25% of the matching documents where main scored all of them, at 0.66x main's time.
- **MAXSCORE disjunction, pruned posting, pruned fixed and interval phrase** are separate roots with their own leaf that reads the persisted score bounds, so `count`, `docs`, `scored` and unpruned top-k never carry the bound-reading machinery. The pruning containers no longer live inside the postings reader, so nothing structural prevents pruning across fields any more.
- **Table filters are part of the plan.** A `.col` predicate and the score filter reach the root as `search::TableFilter`. `count::` and `top::` narrow inside the root (for top-k this is correctness, not speed: a rejected document that reached the collector would raise the k-th score), while `docs::` / `scored::` only skip dead runs through `search::DeadRuns`, because `HitBatcher`'s filter pass is also its projection pass. Main wrapped the iterator and lost pruning entirely whenever a `.col` filter was pushed. A filtered root posting carries its own skip list and seeks to the block a dead run ends in.

## Scoring, aligned with Lucene

| change | before | now |
|---|---|---|
| BM25 numerator | `(k1 + 1) * idf * ...` | `idf * ...`, as Lucene. Every score is `1/2.2` of the old value and ranks are unchanged; hybrid `bm25 + vector` and `constant(N)` comparisons now sit on the same scale as Lucene 10.5, which agrees to six decimals on every reachable (freq, norm) pair |
| DOCS-only field | the field total was 0, so `avg_dl` collapsed to 1 | reports `sumDocFreq` as its total, Lucene's `sumTotalTermFreq == sumDocFreq` |
| missing norm | `dfi`, `lm_dirichlet`, `lm_jelinek_mercer`, `indri_dirichlet`, `raw_dl` scored 0 | a missing norm reads as 1 (`irs::DefaultNorms()`) and a missing frequency as 1, as Lucene's driver does |
| non-text fields (numeric, bool, `IS NULL`), geo, wildcard n-gram | went through the text scorer | `constant(1)` by default; an explicit `::score` still overrides, and inheritance is resolved where the scorer is picked |
| field statistics | collected only in segments where a term matched, so a rare term shrank `N` and `avg_dl` | collected wherever the field exists (`ByTerm`, prefix, range, Levenshtein) |
| constant scorers (`constant`, `idf`, `raw_boost`, `bm25(k1=0)`) | decoded a frequency block per document for a function that ignored it | `Scorer::ScoresPerDoc()` is false, so the plan builds the docs form and attaches one number |
| variadic phrase whose slots all resolved to one term | a per-document boost array carrying a constant | folds to a fixed phrase; the mean term weight becomes the query boost |

`must_not` never scores, in both layers. A sum over three or more clauses of a disjunction can differ from main in the last ulp, because the window sums in leaf order and the old heap summed in pop order; documents and counts are identical.

`score_constant_test.cpp` drives every scorer over every field shape (freq and norm, freq only, neither) through the lead, scored and top roots and asserts exactly when the score folds to one number. The sqllogic expectations follow the scale change and the new `EXPLAIN` shape: `Bool` with `Must` / `Should` / `Must Not` children and `Terms` nodes.

## Everything that used to hold a `DocIterator`

- **Connector scan.** Count -> `count::Root`, or the live count when no predicate survives the zonemap. Streaming -> `scored::Root` or `docs::Root`, one pointer, drained in resumable batches with the batcher's dead-run skipper. Top-k -> `top::Root` with the `.col` filter inside, into a collector seeded with the best k-th any worker has published. ts_dict -> `count::TermCounts` over one lazily filled bitset per segment. `ProxyFilter` is deleted: the stored WHERE conjunct is planned directly, and folding a dense clause into a bitset is what the plan layer does anyway. `SearchRemoveFilter` is a `lead::Node`. Highlighting drives `offsets::Root` per field over the documents already matched.
- **Index.** Flush (`field_data`) and merge (`merge_writer`) exchange `TermPostings`; `PostingsStream` is the one front-to-back list reader and the read-side posting iterators are deleted. `SubReader::mask()` is gone: removals are a per-consumer `MakeMasked` at the root.
- **Vector.** kNN is `top::Root` only; a radius query plans through all four roots. The IVF cluster merge is a sink-top heap and the radius gate moved into the cluster, so the merge sees survivors instead of every document (radius 0.65x main; kNN unchanged).
- **Nested.** One sparse two-phase lead: the parent leads, the child count verifies.
- **Geo.** `GeoQuery<Parser, Acceptor>` with the covering as a prepared `MultiTermQuery` and a concrete check, one translation unit per acceptor.
- **N-gram similarity and variadic phrase** were rewritten: flat position-indexed candidate arrays, the early exit inside the position walk, a sink-top heap for the per-slot term merge. Neither is in the default benchmark set; their numbers come with the wider-set run.
- **`same_position_filter`** is dropped, not ported. No SQL surface reached it.

## Size

`search/` goes from 139 to 585 files and from 53 to 299 translation units. That is deliberate: every shape is one file per consumer (`count/make_sparse_conjunction.cpp`, `top/make_wand_conjunction.cpp`, ...), because the leaves are concrete types instantiated on input type, field layout and arity, and a per-case unit keeps each instantiation out of its neighbours and lets the build run wide. Net the PR is about +15k lines, and roughly 9k of those are the 20-line license header on the added files. The on-disk format is unchanged; the benchmark runs both engines against one index directory.

## Benchmark

search-benchmark-game, wiki corpus, the 962 queries of `queries.txt`, both engines built with the same `perf` preset on the same index, min of 10 iterations, engine-side microseconds. Ratios are patch / main over the sum of per-query minimums.

| command | patch / main | median per query | faster / slower (of 962) | slower by more than 10us |
|---|---|---|---|---|
| `COUNT` | 0.865 | 0.858 | 889 / 47 | 13 |
| `DOCS` | 0.859 | 0.846 | 896 / 40 | 11 |
| `SCORED` | 0.927 | 0.918 | 926 / 23 | 2 |
| `TOP_100` | 0.883 | 0.941 | 876 / 69 | 22 |
| `TOP_100_COUNT` | 0.938 | 0.928 | 863 / 71 | 15 |

By shape, the largest moves: disjunction `COUNT` 0.57 and `DOCS` 0.73; conjunction `TOP_100` 0.66 and `DOCS` 0.77; required-plus-optional `DOCS` 0.42. Phrases are 0.97 to 0.98 across commands: they are bound by position decoding and skip-list parsing, which this PR leaves alone.

**Against Lucene 10.5.0 and tantivy**, same 962 queries, all four engines in one run on 2026-08-31 (the branch a week older than the table above, no LTO or PGO; a refreshed run with those builds will follow):

| command | patch / main | patch / lucene | patch / tantivy | main / lucene | main / tantivy |
|---|---|---|---|---|---|
| `COUNT` | 0.863 | 0.523 | 0.552 | 0.606 | 0.640 |
| `DOCS` | 0.854 | 0.445 | 0.507 | 0.520 | 0.594 |
| `SCORED` | 0.928 | 0.358 | 0.539 | 0.386 | 0.581 |
| `TOP_100` | 0.883 | 0.613 | 0.585 | 0.694 | 0.663 |
| `TOP_100_COUNT` | 0.937 | 0.327 | 0.560 | 0.350 | 0.598 |

Main already led both engines in every multi-query category but one, `TOP_100` conjunctions (1.14x Lucene, 1.18x tantivy); with the block-max conjunction that category is 0.77x and 0.80x, and the patch leads in every multi-query category of every command.

The wider `queries_all` set (variadic phrases, n-gram similarity, thresholds, slop and interval phrases, fuzzy, the boolean shapes) will be measured in a fresh run and added here.

Documents and counts agree with main on every query of every command. The queries still slower by more than 10us are a handful of two- and three-token phrases (up to 14%, the largest +123us), several conjunctions under `TOP_100` / `TOP_100_COUNT` (up to 1.45x on `+the +preakness`, +178us on `+city +in +iran`), `the incredibles` under `TOP_100` (1.36x, max-score disjunction) and `+bass -fish -guitar` under `TOP_100` (1.16x). These are being worked down before merge.

## Follow-ups

None of these hold the PR back. As it stands it beats main, which already led Lucene and tantivy on this benchmark, and it does so without LTO or PGO; those builds will be measured and added here.

1. **Scoring rework** (#1068, #1069). Every multi-term expansion becomes constant-scored by default, with scoring every term as an explicit opt-in; fuzzy keeps per-term scoring through one index-wide heap with a lower default term limit; variadic slop phrases regain their per-term boosts (#1083) and sloppy phrases score match tightness the way Lucene feeds `sloppyWeight` into the frequency (#1082).
2. **Cheaper constant scoring** (#1081). A constant-scored shape should not collect and score every candidate in prune mode; the pruned phrases and the positional `lead/` `probe/` `fill/` makers still lack the constant fast path.
3. **Deletions** (#198). The mask becomes another table-filter-style axis resolved into the root's type, so a segment with removals keeps its pruning plans instead of falling back to a masked walk.
4. **Threshold (`min_should_match`) shapes.** Rewritten and faster than main, but not tuned: window-level abandon, the k-th-smallest advance, the pigeonhole plan.
5. **More pruning.** Exclusion (a negated top-k query scores every match today where main pruned its include side), boosted optional clauses, tighter bounds inside disjunctions and conjunctions, the mixed required-plus-optional shapes, non-term clauses, and streaming block-max pruning on `scored::Root` (the connector still computes the threshold and nothing reads it).
6. **Bitset work.** A folded bitset beside long terms beats the window in more cases than the current threshold admits.
7. **AVX-512 popcount** for the bitset and window counts.
8. The load-test reference is being regenerated for the new scores.


pashandor789 (approved):
thank you for a tiny contribution !

PR #1061, opened 2026-08-30
fix: deadlock between catalog commits and object id allocation

gnusi:
Two concurrent transactions that each commit several catalog entries could deadlock the server for good. The process stayed alive and stopped answering any query.

## The cycle

The cluster catalog log is guarded by one process-wide mutex. It is taken on the first record a transaction writes and held until that transaction ends (`gHoldsClusterWal`, released in `EndClusterCatalogWal`). The per-catalog write lock and the per-`CatalogSet` lock, by contrast, are taken and released **once per undo entry** in `CommitState::CommitEntry`.

The log lock was reached from under those, from two different directions:

1. **Commit** — `WriteCatalogChange` runs under both entry locks and takes the log lock there. Thread A commits entry 1, takes the log lock, keeps it, releases the entry locks, then re-takes them for entry 2 while still holding it; thread B holds those entry locks and waits for the log lock.
2. **Object id allocation** — an id is handed out wherever a catalog entry is built, which is also under the catalog locks. Raising its durable horizon writes to the same log, taking the same mutex. Same inversion, reached the other way round.

## Reproducer

`tests/scripts/catalog_commit_deadlock.sh` — eight workers, each committing multi-entry DDL (`CREATE TABLE` with a primary key and a check constraint, then `DROP`) against the same `CatalogSet`, with a bounded probe throughout and after:

```
$ tests/scripts/catalog_commit_deadlock.sh --port 7978 --workers 8 --seconds 40
FAIL: server stopped answering while the load was running
```

Before this PR it wedges in about 25 seconds and stays wedged after the load stops. The load is deliberately plain DDL.

It runs in CI as `scripts/ci/steps/050-ci-in-docker-run-catalog-deadlock-test.bash`, called from `run_serened_core`. It brings up **its own serened** rather than sharing the sqllogic stack: the failure it looks for takes the whole process down, so a regression must not drag the other suites with it.

## The fix

**Commit path.** `PrepareCatalogChange` (added in serenedb/duckdb#78) is announced in `CommitState::CommitEntry` before the per-entry locks are taken, and `SereneDBCatalog` overrides it to acquire the log lock there. The log lock is now unambiguously outermost on that path.

**Object id horizon.** The horizon is raised at transaction start instead, in `SereneDBTransactionManager::StartTransaction`, where no catalog lock is held, with `kOidHeadroom = 4096` ids of headroom covering what the transaction goes on to allocate (`DatabaseManager::EnsureOidHeadroom`, also in serenedb/duckdb#78).

An allocation that still runs out *while under a catalog lock* refuses rather than waits, and reports a retryable `ERRCODE_T_R_SERIALIZATION_FAILURE`. That asymmetry is the point: the horizon has to be durable before an id is handed out, so waiting is the only alternative, and waiting under the catalog locks is exactly what deadlocks. Waiting is permitted in one place only, marked by `OidHorizonWaitScope` — the transaction-start top-up itself.

## Results

| | lock-order inversions | reproducer |
|---|---|---|
| before | 2738 | wedges in ~25s |
| commit path only | 10 | still wedges |
| both | **0** | **passes**, also at 16 workers / 60s |

No retryable conflict fired in any of those runs, so the refuse path stays theoretical in practice.

Regression run on `sdb/pg/ddl`, `sdb/pg/system` and `sdb/pg/rbac`: **102 passing, 0 failing**.

An earlier attempt regressed 22 of those with `object id horizon is busy`: the transaction-start top-up was itself taking the refuse path, in the one place where waiting is safe. That is what `OidHorizonWaitScope` exists to separate, and why the distinction is explicit rather than implied.

## Depends on

serenedb/duckdb#78 — `PrepareCatalogChange` and `EnsureOidHeadroom`. The submodule bump in this PR points at that branch.

### DIFF
--- a/scripts/ci/steps/050-ci-in-docker-run-catalog-deadlock-test.bash
+++ b/scripts/ci/steps/050-ci-in-docker-run-catalog-deadlock-test.bash
@@ -1,0 +1,64 @@
+#!/bin/bash
+
+# Concurrent multi-entry catalog commits must not deadlock the server.
+#
+# Runs against a serened of its own rather than the shared sqllogic stack: the
+# failure this looks for takes the whole process down, so a regression here must
+# not take every other suite with it.
+
+set -o pipefail
+
+cd "${WORKSPACE}" || exit 1
+
+PORT="${CATALOG_DEADLOCK_PORT:-7899}"
+DATA_DIR="$(mktemp -d)"
+LOG_DIR="./out/logs"
+mkdir -p "${LOG_DIR}"
+
+"${BUILD_DIR}/bin/serened" "${DATA_DIR}" --listen="postgres://0.0.0.0:${PORT}" \
+	>"${LOG_DIR}/catalog-deadlock-serened.log" 2>&1 &
+server_pid=$!
+
+cleanup() {
+	kill -9 "${server_pid}" 2>/dev/null
+	wait "${server_pid}" 2>/dev/null
+	rm -rf "${DATA_DIR}"
+}
+trap cleanup EXIT
+
+ready=0
+for _ in $(seq 1 90); do
+	if ! kill -0 "${server_pid}" 2>/dev/null; then
+		break
+	fi
+	if timeout 5 psql -h 127.0.0.1 -p "${PORT}" -U postgres -d postgres -tAc "SELECT 1;" \
+		>/dev/null 2>&1; then
+		ready=1
+		break
+	fi
+	sleep 1
+done
+
+if [[ ${ready} -ne 1 ]]; then
+	echo "serened did not come up on port ${PORT}" >&2
+	tail -30 "${LOG_DIR}/catalog-deadlock-serened.log" >&2
+	echo "CATALOG_DEADLOCK_TEST=FAILED"
+	exit 123
+fi
+
+if ./tests/scripts/catalog_commit_deadlock.sh --port "${PORT}" \
+	--workers "${CATALOG_DEADLOCK_WORKERS:-8}" \
+	--seconds "${CATALOG_DEADLOCK_SECONDS:-40}" 2>&1 |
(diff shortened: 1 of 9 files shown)


gnusi:
Updated: the deadlock is now closed end to end, and the reproducer runs in CI.

The remaining half was `WriteOidHorizon`. An object id is allocated wherever a catalog entry is built — under the catalog and set locks — and raising its durable horizon writes to the cluster log, taking the log lock from under them. Same inversion as the commit path, reached the other way round.

Raising the horizon now happens at transaction start, where no catalog lock is held, with headroom (4096) covering what the transaction goes on to allocate. An allocation that still runs out while under a lock refuses rather than waits and reports a retryable serialization failure — the horizon has to be durable before an id is handed out, so waiting there is the only other option and that is what deadlocks. Waiting is permitted in exactly one place, marked by `OidHorizonWaitScope`: the top-up itself.

Results on the reproducer (`tests/scripts/catalog_commit_deadlock.sh`):

| | lock-order inversions | reproducer |
|---|---|---|
| before | 2738 | wedges in ~25s |
| commit-path fix only | 10 | still wedges |
| both | **0** | **passes** (16 workers / 60s too) |

Zero retryable conflicts were observed in those runs, so the refuse path does not fire in practice.

Regression run on `sdb/pg/ddl`, `sdb/pg/system` and `sdb/pg/rbac`: 102 passing, 0 failing. An earlier attempt regressed 22 of those — the top-up was refusing at transaction start, where waiting is safe — which is what `OidHorizonWaitScope` exists to separate.

CI: `scripts/ci/steps/050-ci-in-docker-run-catalog-deadlock-test.bash`, called from `run_serened_core`. It brings up its own serened rather than sharing the sqllogic stack, because the failure it looks for takes the whole process down.

DuckDB side updated in serenedb/duckdb#78 (adds `EnsureOidHeadroom`).

PR #1063, opened 2026-08-30
feat: move doc examples onto sdb_ names

gnusi:
Mostly sqllogic fixture adjustments, plus one alias in the engine that two of them need.

### Fixtures

`site_docs` fixtures that called a `duckdb_*` metadata function now call its `sdb_*` alias. These fixtures render as the examples on docs.serenedb.com through `<SqlLogicTest>`, so they change together with serenedb/serenedb-site#275.

Names without an alias are unchanged.

Two fixtures follow their renamed pages: `duckdb_table_functions.test` -> `metadata.test`, `duckdb_environment.test` -> `environment.test`.

### Engine

`AliasNameFor` swapped a prefix, which covers only names that start with `duckdb_`. It now swaps the substring wherever it occurs, so `truncate_duckdb_logs` can alias as `truncate_sdb_logs`.

### Checks

`pre-commit` passes on the touched files. All `SqlLogicTest` ids referenced from the docs tree resolve to a fixture and marker (2845 ids, 0 broken) with both branches applied.


Dronplane (approved):
LGTM

PR #1064, opened 2026-08-31
fix: bump duckdb_iceberg for the IcebergTransaction lock

gnusi:
Bumps `third_party/duckdb_iceberg` to pick up serenedb/duckdb-iceberg#13.

**Stacked on that PR** — `scripts/check_submodule_pointers.py` requires the gitlink to be reachable from a `vYYYY.MM.DD` version branch, so the submodule-pointer check is expected to be red here until #13 merges into `v2026.07.10`. I will repoint this to the merged SHA then; the local commit used `SKIP=check-submodule-pointers`.

### What it fixes

`IcebergTransaction` holds catalog state shared by every schema in the catalog — `tables`, `schemas`, `listed_schemas`, `looked_up_entries`, `current_table_data` — but the code mutating it runs under `IcebergTableSet::entry_lock`, a **per-table-set** mutex, one per schema. Two executor threads scanning different schemas take different locks and write the same container.

DuckDB reaches this on its own: `duckdb_columns()` fans out over every attached catalog, and a statement with several metadata scans gets several `PipelineInitializeTask`s in flight, each landing in `IcebergSchemaEntry::Scan` → `IcebergTableSet::Scan`.

### How it showed up here

An ASAN run on #1063 aborted `serened` outright and took 45 tests with it, all reporting `connection closed`:

```
Assertion `!cap.IsReentrance() && "Reentrant container access during
element construction/destruction is not allowed."' failed.
  #17 IcebergTransaction::ReferenceTable  iceberg_transaction.cpp:50
  #18 IcebergTableSet::Scan               iceberg_table_set.cpp:89
  #19 DuckDBColumnsInit                   duckdb_columns.cpp:103
```

Not reentrancy — one thread's in-construction sentinel in the capacity word reads as reentrancy to the other. ThreadSanitizer names the same defect directly, on a sibling container, with the two threads holding **different** mutexes (M0/M1) on one address allocated in `IcebergTransactionManager::StartTransaction`.

### Verification

Reproduced locally against a REST catalog of 6 schemas with a `UNION ALL` of 24 `duckdb_columns()` scans: **21 races → 4** after the fix, with every `raw_hash_set`/`LoadEntries` race gone. The 4 survivors are an unrelated bug in `PhysicalIcebergCreateTable::MakeCreateTableRequest` (`std::call_once` on a per-instance flag guarding writes to a shared plan operator), documented in #13 and not addressed.

Upstream `duckdb-iceberg` still carries two of the three races and does not have the third at all, so a rebase would not have covered this — details in #13.


MBkkt:
Done in #1066

PR #1065, opened 2026-08-31
test: stop the definer-view test comparing a hijackable view to a pinned one

gnusi:
`enf_internal_view_definer.test` has been failing intermittently in CI:

```
substitutable  sqlite_wrapper
-   t   t
+   f   t
    at sdb/pg/rbac/enf_internal_view_definer.test:169
```

### It is not a concurrency flake

That was my first theory and it is wrong. Measured: **4305 executions** of the two-count statement while six sessions churned `CREATE`/`DROP TABLE` across six databases, two churned `CREATE`/`DROP DATABASE`, and four readers ran in parallel — the counts moved between 6 and 13 with **zero mismatches**. 1500 further samples through the definer views: zero. Intra-statement catalog reads are frozen per statement by `MetaTransaction::GetStatementDatabases` (`meta_transaction.cpp:300-306`, cleared only in `SetActiveQuery`), and every database is read through the meta transaction's pinned `DuckTransaction`.

### What it actually is

The two views being compared are not equivalent by construction:

| view | body | shadowable? |
|---|---|---|
| `w_source` | `SELECT * FROM duckdb_tables` — **unqualified** | yes |
| `w_alias` | `SELECT * FROM sdb_tables`, whose body is pinned to `system.main.duckdb_tables` | no |

A view body is bound with the **caller's** search path (`bind_basetableref.cpp:330-336`), and the caller's temp catalog wins the unqualified name. So any relation named `duckdb_tables` visible to the reader sends `w_source` somewhere else entirely while `w_alias` still reads the catalog.

Reproduced on a live server, matching the CI symptom exactly:

```
                        src | alias | substitutable
before                    0 |     0 | t
after CREATE TEMP TABLE
duckdb_tables (3 rows)    3 |     1 | f     <-- w_source is reading the temp table
```

And `sdb_function_aliases.test:155` plants precisely that relation — `CREATE TEMP TABLE duckdb_tables (pwned INTEGER)` — with a comment already recording the hazard: *"An unqualified body made all nine hijackable by any role able to create a TEMP table, and poisoned the view's cached column info process-wide."* #1062 fixed that for the `sdb_*` bodies, then wrote a control view keeping the hijackable shape.

### The fix

Qualify both sides so `w_source` is an actual control for `w_alias`. Verified: with `duckdb_tables` **and** `sdb_tables` both shadowed by temp tables, `substitutable` and `aliased` stay `t`, where the unqualified form gives `f`.

This is not a scoped-down assertion — it still compares full catalog counts through definer rights as the unprivileged role. It just compares two relations that are actually the same relation.

### DIFF
--- a/tests/sqllogic/sdb/pg/rbac/enf_internal_view_definer.test
+++ b/tests/sqllogic/sdb/pg/rbac/enf_internal_view_definer.test
@@ -80,7 +80,8 @@ table	zebrafin_vault
 # ride the same rule.
 
 query
-SELECT (SELECT count(*) FROM sdb_tables) = (SELECT count(*) FROM duckdb_tables)
+SELECT (SELECT count(*) FROM system.main.sdb_tables)
+     = (SELECT count(*) FROM system.main.duckdb_tables)
   AS aliased
 ----
 aliased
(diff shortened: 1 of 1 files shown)


Dronplane (approved):
LGTM

PR #930, opened 2026-07-17
feat: Split catalog and data

MBkkt:
well, attempt to make 
- sequences fast
- recovery of inverted indexes not memory bound and parallel
- dml for inverted indexes parallel
- catalog transactional?!
- i want to die
- catalog written as wal
- separate duckdb databases for separate serenedb databases
- many bug fixes

PR #1067, opened 2026-08-31
test: catalog stress suite for parallel DDL/DML churn

gnusi:
Randomized parallel DDL/DML churn against a `serened` of its own, with a model-based
consistency oracle, a hang detector and seeded repro. Plus the first regression test it
produced.

`#930` changed DDL concurrency control to optimistic MVCC on `CatalogSet` version chains:
no `ACCESS EXCLUSIVE` lock, and the second writer of an entry is refused with `40001` on
the spot. Our coverage of that is almost entirely sequential. The concurrent cases are a
handful of hand-unrolled sqllogic files, because the vendored `sqllogictest-rs` runner has
no loop, no randomization, no duration bound and no per-record timeout -- `recovery/wal_group_commit_checkpoint_race.test`
literally writes out `connection w1 ... w16` five times over.

This is the general form of `tests/scripts/catalog_commit_deadlock.sh`: same idioms
(bounded clients, a concurrent liveness probe on fresh connections, its own serened,
PASS/FAIL verdict), widened to the DDL surface with an oracle.

## What it asserts

1. **No crash** -- the server log is scanned for the sanitizer/assert/fatal regex shared with `tests/drivers/sqlsmith/run.sh`.
2. **No hang** -- a watchdog probes `SELECT 1` on a *fresh* connection, then samples `/proc/<pid>/task/*` twice to prove every thread is frozen, with a wchan histogram. No debugger, no core, no ptrace: `ptrace_scope=1` blocks attaching to a sibling. `statement_timeout` is deliberately unused -- the GUC says "accepted for compatibility but not currently enforced", so every deadline is client-side.
3. **Consistency** -- at each quiesce the harness's model is diffed against `pg_class` + `pg_description` + `sdb_catalog_sets()`, plus dependency-edge integrity, oid non-reuse, and an orphan-artifact walk of the datadir.
4. **Load held** -- committed ops, retries and a per-label error histogram. The budget metric is *committed* ops, never ops/s: a refusal is cheaper than a commit, so an ops/s floor reads a conflict storm as healthy.

The model is a **candidate set**, not a single expected state. An ambiguous outcome
(crash / cancel / timeout) leaves its key holding both the before and after state,
because the two catalog crash windows differ exactly there: before the log flush the
statement never happened, after it the commit is delayed rather than lost. After a restart
every ambiguous key must resolve to a member of its own set; a third state is the finding.
Identity is a `COMMENT ON` token read back through `pg_description`, so drop-then-recreate
is distinguishable from survival.

## What it found

| # | issue | state |
|---|---|---|
| 1 | Concurrent DDL wedges the server | known class, unfixed |
| 2 | A rolled-back `DROP SEQUENCE` durably erases the sequence counter | **new**, test included |
| 3 | Cross-database write: a permanent condition reported as retryable `40001` | new |
| 4 | Session `is_superuser` / `current_user` go stale when the role is dropped | new |
| 5 | Unimplemented DDL leaks pragma names, and one path leaks `XX000` | new |

**1.** `tables_only` at 8 workers wedged at 36.1 s after 93,576 ops: 125 threads, two
`/proc` samples 4 s apart identical for every thread, ~100 blocked on the *same* futex
word. Symbolized on a TSAN build with `resources/suppressions/tsan.txt` removed: 10
lock-order inversions, cycle `M0 => M1 => M0` with `sdb::catalog::ScopedCatalogWal()`
(`duckdb_global_catalog.cpp:317`) taken while holding a `lock_guard` from
`CommitEntry` (`commit_state.cpp:309`). This is the class `tsan.txt:33-51` calls
"Suppressed, NOT fixed"; `origin/fix/catalog-commit-deadlock` addresses it.

Reachable by any concurrent DDL given enough volume; sequences set the rate:

| workload | workers | time to wedge |
|---|---|---|
| `serial_churn` | 2 | 28 s |
| `tables_only` | 2 | ~124 s |
| `tables_only` | 4 | under 20 s |

With `tsan.txt` loaded the same run reports **nothing**, which is why this adds
`resources/suppressions/tsan-stress.txt` for the TSAN leg.

**2.** `DROP SEQUENCE` has two halves and only one is transactional. `WriteSequenceDropped`
appends and flushes under `gClusterWalMutex` with no transaction, so `ROLLBACK` restores
the entry but not the counter. Controlled A/B across a crash-restart: control resumes at
34 (the counter is logged 32 values ahead), treatment at 1. On a `SERIAL` table the rows
survive, so the next insert collides with its own primary key -- `Duplicate key "id: 1"`
on a table nobody dropped. Pinned by
`tests/sqllogic/recovery/sequence_counter_after_rollback.test` in the
`index_rollback_drop_gap.test` style, with the correct expectation written beside each
KNOWN GAP. It is self-controlling: an untouched sequence must still resume ahead across
the same crash, so it cannot be satisfied by "a crash loses sequence counters".

## Crash recovery is sound

Three injected catalog-window crashes with a restart and a full oracle pass after each:
`crash_after_catalog_before_data` and `crash_before_catalog_commit` both crashed and came
back (3.5 s, 5.0 s), and the oracle found **zero** inconsistencies -- every ambiguous key
resolved into its candidate set, no missing committed objects, no dangling edges, no orphan
artifacts. The catalog half of `#930` recovers correctly; what is broken is liveness, not
durability.

## False-positive baseline

The oracle ran to completion on **958,864 committed ops with zero findings** (1 and 2
worker `tables_only`, plus `ddl_churn` / `ddl_dml_race` / `dependency_churn` at 2 workers).
That is what makes a red run worth believing.

## Running it

```
tests/stress/run.sh --profile smoke
tests/stress/run.sh --scenario serial_churn --workers 2 --seconds 60 --seed 31337
```

`BUILD_DIR` / `SERENED` pick the binary, as in `tests/network/run.sh`. Every flag has an
`SDB_STRESS_*` env equivalent. Fault-dependent profiles need `-DSDB_FAULT_INJECTION=On`;
the suite prints `fault_injection=` in preflight and degrades cleanly when it is off.

Unit tests need no server: `python3 -m pytest tests/stress/` (110 tests covering the
collapse algebra, the classifier's closure, and that every fault constant resolves to a
real source literal).

## CI

New step `051`, run last and against its own serened, because the failure it looks for
wedges the whole process -- the same reason `run_stress` sits after `wait_bg` rather than
inside `run_serened_core`. Gated on `RUN_STRESS`, which defaults to false, so **this PR
does not enable it anywhere**. `STRESS_SOFT_FAIL` (default true) reports failures without
failing the job, for the burn-in period the suite has not yet served.

The `.github/workflows` inputs that pass `RUN_STRESS` / `SDB_STRESS_*` through are **not
in this PR** -- pushing workflow files needs a token scope I do not have here. They are a
149-line follow-up; without them the step is unreachable, which is the intended default
state anyway.

## Notes for review

- `tests/harness/python/{serened,procutil}.py` is the process lifecycle extracted from `tests/network/hba_mask_test.py`, which imports it back and passes unchanged.
- Four oracle facts were verified against a running server because each would otherwise be a false positive: `sdb_catalog_sets().visible` is hardcoded true (the doc comment above it says otherwise); `entry_type` is human-readable (`Dependency`, not the enum spelling); an index occupies **two** slots and its `Table` row must be dropped or every index reads as a ghost; implicit `*_pkey` indexes are in `pg_class` but not in `sdb_catalog_sets()`.
- `SET sdb_faults` does not validate names -- a typo arms nothing, silently -- so `test_fault_catalog.py` checks every constant against the real `SDB_IF_FAILURE` literals, and refuses any name listed in the checker's `KNOWN_UNTESTED_SOURCE_FAULTS` so declaring a constant cannot quietly retire a real gap.
- `scripts/check_fault_points.py` lists `crash_before_seq_counter_wipe`, which exists in no branch's `server/` or `libs/`. Left alone here; the sequence-counter crash window is currently uninstrumented.

---

## Update: second wave

The suite has since grown the chaos, coverage and quarantine machinery, and a
verification pass drove every remaining DDL family against a live server (70 object
families, 258 SQLSTATEs recorded). Findings are now **30**, including three crashes or
corruptions and several security-relevant items. Highlights:

- **The catalog-log rewrite crashes the server under concurrent DDL.** Reproducible in
  43 s: `tests/stress/run.sh --profile compaction-probe --scenario tables_only --workers 2
  --seconds 60 --compaction-windows 1 --seed 909` gives `verdict=dead` with
  `XX000 Attempted to dereference unique_ptr that is NULL!` then SIGABRT ->
  `std::terminate` in `SereneDBTransactionManager::CommitTransaction`. Both
  `compact_inside_ddl` and `compact_inside_drop` do it. This is the window
  `store.cpp:321-327` calls "not load-safe today".
- **Automatic catalog-log compaction is dead code**: `MaybeCompact()` gates on
  `_live_bytes`, which is never assigned anywhere in the tree. The log grows monotonically
  (~858 B per CREATE/DROP pair) and no SQL statement compacts it -- `CHECKPOINT` leaves it
  byte-identical.
- **A permission-denied or rolled-back `DROP DATABASE` leaves a zombie**: the catalog row
  returns, the attachment does not, and `DROP DATABASE IF EXISTS` then reports success
  while doing nothing.
- **Trigger durability is inverted**: a trigger survives `kill -9` but is silently lost
  across a graceful `SIGTERM` restart.
- **A prepared statement over a recreated table returns silently wrong data** for one
  execution (an integer decoded as a string against the stale RowDescription).
- **Secrets have no isolation**: a plain non-superuser can list and drop the superuser's.
- **`ALTER DEFAULT PRIVILEGES` and `ALTER ROLE ... SET` are stored and never applied**,
  yet the dead default-privilege entry still blocks `DROP ROLE`.

Also in this branch now: crash-window chaos with restart + oracle collapse (clean --
crash recovery is sound), graceful-SIGTERM restarts, index-build parks under load,
protocol-level cancellation (atomic -- clean), name-reuse/ABA, a shared-arena conflict
scenario (252k ops, 9.5k real 40001s, 0 findings), a row-content oracle, a rot-proofed
quarantine registry, and a coverage report with two gates.

Twelve harness bugs of my own were found and fixed on the way, nine of them by a check
rather than by inspection -- including a 2,318-false-positive run and one false *negative*
where a wedge during a chaos window was masked as healthy. Both are written up.

Unit tests: **157**, no server needed. The oracle now has **15 self-mutants** proving each
invariant actually fires; without them a green run only proved the oracle does not cry
wolf, not that it can see anything.

Known state unchanged: the suite still fails on `main` by design, and CI ships gated off
(`RUN_STRESS=false`) and soft-fail.


resources/suppressions/tsan-stress.txt
MBkkt, line 1: we shouldn't have separate suppression

.gitignore
```diff
@@ -45,6 +45,12 @@ package-lock.json
 /sqllogic-tests.log
 /recovery-tests.log
 
+# Catalog stress suite: a run's own artifacts go to /out/stress (covered above).
+# These two are what a human collects out of a run to hand around -- an evidence
+# bundle and a write-up. Kept local; the findings belong in the PR or the issue.
+/catalog_stress_evidence/
+/catalog_stress_findings.md
```
MBkkt: remove this?

tests/stress/known_wedges.yaml
MBkkt, line 1: remove this?

MBkkt (approved):

PR #1072, opened 2026-09-02
perf: date_trunc fast paths for TIMESTAMP and TIMESTAMPTZ, ES date_histogram without per-row strftime

gnusi:
First part of the `date_trunc`/GROUP BY work. Continued in serenedb/serenedb#1075, whose description explains the whole picture (scalar fast paths, the zone table, the GROUP BY bucket rewrite and the results).

Speeds up `date_trunc` on TIMESTAMP, DATE and TIMESTAMPTZ and fixes the SQL shape the ES `date_histogram` handler emits.

**duckdb submodule** -> serenedb/duckdb PR https://github.com/serenedb/duckdb/pull/83 (branch `gnusi/date-trunc-lut` on top of `v2026.08.25`; the pointer here is that branch head and has to be moved to the merged commit once the fork PR lands, so the submodule-pointer check fails until then):
- TIMESTAMP/DATE: the truncation arithmetic is shared via `duckdb/common/operator/date_trunc_operators.hpp`, runs as an inlined functor (the function-pointer executor never inlined, even under LTO), fixed units are one floor division, calendar units use DuckDB's cumulative tables inlined with a fast path inside the 1970-2369 cycle.
- TIMESTAMPTZ: a per-zone lookup table (one 16-byte entry per day for 1900-2299, built from ICU's transition list, cached by zone id) replaces per-row `icu::Calendar` work. Wall-clock results resolve back to instants with ICU's rules (later offset for duplicated wall times, earlier for skipped ones). Rows outside the window, on days with several transitions, non-Gregorian calendars, `era`, BC decades and the quarter/decade gap quirk fall back to ICU, so results are identical.

**ES `date_histogram`** (`server/network/http/es/handlers.cpp`): aggregate on the truncated timestamp and format `key_as_string` in an outer projection instead of running `strftime` on every row and grouping on a VARCHAR key.

**Test**: `tests/sqllogic/sdb/pg/simple/temporal/date_trunc_timestamptz.test` pins fixed-offset zones, DST transitions (Europe/Berlin, Europe/Moscow, Australia/Lord_Howe), date-line jumps (Pacific/Apia, Pacific/Kiritimati, Pacific/Kwajalein), the São Paulo/Havana gap quirks, Africa/Monrovia's seconds offset, out-of-window fallback, BC `decade` semantics and the TIMESTAMP path. Several cases come from ClickHouse's and DuckDB's own test suites.

Numbers (100M rows, 32 threads):

| query | before | after |
|---|---|---|
| TIMESTAMP `hour` | 34 ms | 22 ms |
| TIMESTAMP `month` | 82 ms | 35 ms |
| TIMESTAMPTZ Etc/UTC `hour` | ~1000 ms | 37 ms |
| TIMESTAMPTZ Europe/Berlin `hour` / `day` | ~1000 ms | 74 / 62 ms |
| ES date_histogram SQL shape | 300 ms | 29 ms |

Verified against the previous binary: every representable day for all calendar units (TIMESTAMP and DATE), and for TIMESTAMPTZ a 15-minute grid over 1899-2300 with ±1 µs neighbours plus 10M random instants in 16 DST/odd-offset zones and 4 fixed-offset zones, all units, constant and non-constant specifier: identical checksums. The ES SQL shapes are set-equal.

### Second commit

- duckdb pointer -> `4a5882a0dda`: `date_part`/`extract` (all parts and forms), `last_day`, `monthname`, `dayname`, the `TIMESTAMP <-> TIMESTAMPTZ` / `-> TIMETZ` casts and `timezone()` with a constant zone go through the same per-zone table; identical to ICU on the same 19-zone grid, except `last_day` no longer returns the first day of the next month for São Paulo January 1965 (an ICU-path bug: truncating division of a negative epoch after a midnight gap).
- `server/pg/serialize.cpp`: the TIMESTAMPTZ / TIMESTAMPTZ_NS text renderers take the session offset from the table instead of `icu::TimeZone::getOffset` per value.
- `server/pg/deserialize.cpp`: zone-less TIMESTAMPTZ text input resolves through the table (same duplicated/skipped wall-time rules as the casts), falling back to ICU outside the table.
- `tests/sqllogic/sdb/pg/simple/temporal/date_part_timestamptz.test`: parts across the Berlin fall-back, both cast directions through a gap and a duplicated hour, `TIMETZ`, `timezone()`, the 1965 `last_day`, Kolkata half-hour offsets.

100M rows, 32 threads, Europe/Berlin: `date_part('hour')` 409 -> 55 ms, `TIMESTAMPTZ -> TIMESTAMP` 532 -> 50 ms, `TIMESTAMP -> TIMESTAMPTZ` 52 ms, `last_day` 71 ms.

### Third and fourth commits

- duckdb pointer -> the fewer-divisions table path (Europe/Berlin `hour` 74 -> 55 ms, `day` 62 -> 50) and the dense-bucket `GROUP BY` rewrite: `GROUP BY date_trunc(<second|minute|hour|day|week>, ts)` with a finite range that fits the perfect-hash threshold becomes a bucket-index key with the timestamp restored above the aggregate, so the plan uses `PERFECT_HASH_GROUP_BY` (100M rows: 720 hourly groups 29 -> 21 ms, 30 daily groups 27 -> 20 ms). Larger ranges, non-integral co-groups, grouping sets, distinct aggregates and columns with infinities are left alone.
- `tests/sqllogic/sdb/pg/simple/temporal/date_trunc_bucket_groups.test`: pins the rewritten plan and the untouched plans (infinity, minute range too wide), results for hour/week/day/minute/month buckets, HAVING, two groups, ROLLUP, NULL timestamps, DATE and non-constant units, and the same results with `compressed_materialization` disabled.

### DIFF
--- a/server/network/http/es/handlers.cpp
+++ b/server/network/http/es/handlers.cpp
@@ -495,11 +495,11 @@ yaclib::Task<bool> RunAggregation(RequestContext& ctx, const Aggregation& agg,
     case Aggregation::Kind::kDateHistogram:
       // agg.interval comes from the calendar_interval whitelist.
       sql = absl::StrCat(
-        "SELECT epoch_ms(date_trunc('", agg.interval, "', ", field,
-        ")) AS k, strftime(date_trunc('", agg.interval, "', ", field,
-        "), '%Y-%m-%dT%H:%M:%S.000Z') AS ks, count(*) AS c FROM ", relation,
+        "SELECT epoch_ms(b) AS k, strftime(b, '%Y-%m-%dT%H:%M:%S.000Z') AS "
+        "ks, c FROM (SELECT date_trunc('",
+        agg.interval, "', ", field, ") AS b, count(*) AS c FROM ", relation,
         " WHERE (", filter, ") AND ", field,
-        " IS NOT NULL GROUP BY 1, 2 ORDER BY 1");
+        " IS NOT NULL GROUP BY 1) ORDER BY 1");
       break;
     default: {
       std::string_view fn;
--- a/server/pg/deserialize.cpp
+++ b/server/pg/deserialize.cpp
@@ -27,6 +27,7 @@
 #include <fast_float/fast_float.h>
 
 #include <cstdint>
+#include <cstring>
 #include <duckdb/common/operator/cast_operators.hpp>
 #include <duckdb/common/operator/decimal_cast_operators.hpp>
 #include <duckdb/common/types/bit.hpp>
@@ -43,6 +44,7 @@
 #include <duckdb/function/cast/default_casts.hpp>
 #include <duckdb/inet/inet_ipaddress.hpp>
 #include <duckdb/main/client_context.hpp>
+#include <icu-zone-lut.hpp>
 #include <limits>
 #include <string>
 #include <string_view>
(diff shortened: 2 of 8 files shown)


gnusi:
Closed: consolidated into #1075, which is based on `main` and carries these five commits as its first five, so nothing here is dropped. The fork side is likewise one PR now, serenedb/duckdb#85.

PR #1074, opened 2026-09-02
Fix every ThreadSanitizer finding for real and delete the serenedb catalog mutexes

pashandor789:
TSAN work only, one commit on top of main. The index scan-wrapper dissolve is **not** in here.

- The #930 suppressions go away; `Catalog::_mutex` and its statement-scoped `MutationScope` are deleted outright, so DDL serializes on duckdb's per-catalog write lock and MVCC conflicts the way upstream does. Catalog records are buffered per transaction and spliced into the log under duckdb's own StorageManager wal lock, and the log rewrite runs whole under it with an optimistic marker.
- Roles are catalog entries that own their record, so a contextless read no longer hands out an entry reference from under the set lock; permissions changes ride duckdb's own alter road.
- Cloud cookbook tests name their secrets per database, since no global DDL lock is left to serialize them.

Submodule pins carry the fork-side halves — merged as serenedb/duckdb#82 (on `v2026.08.25`) and serenedb/duckdb-iceberg#13. Note the duckdb pin moves the version branch forward, so it also brings in what was already merged there: the fork-side scan-wrapper hooks (#81) and gnusi's SharedObjectCache (#75). Verified that pre-dissolve serenedb code builds and tests clean against it.

Local: build clean, full sdb/pg suite **2202 passed** with only the 4 pre-existing failures (`create_index_typed_keys` x2, `encrypted` x2), pre-commit green including `check-submodule-pointers`.

Supersedes #1073 (which carried the dissolve as well).


gnusi (approved):

server/auth/role_closure.h
```diff
@@ -40,11 +40,7 @@ namespace duckdb {
 class ClientContext;
 
 }  // namespace duckdb
-namespace sdb::catalog {
-
-class Role;
-
-}  // namespace sdb::catalog
+namespace sdb::catalog {}  // namespace sdb::catalog
```
MBkkt, line 43:
```suggestion
```

PR #1075, opened 2026-09-02
perf: date_trunc fast paths, GROUP BY bucket rewrite and monotone predicate ranges

gnusi:
The whole `date_trunc`/GROUP BY work in one PR, based on `main`; it supersedes #1072. Three commits, with the fix kept out of the perf work: the `date_bin` infinity fix, then the engine side, then the sqllogic coverage. Pairs with the single fork PR serenedb/duckdb#95, five commits based on `v2026.07.10`, with fixes, features and the cleanup kept in separate commits. The submodule here pins its head. Their descriptions explain the two layers of the change: cheaper scalar functions (inlined truncation, the day-indexed calendar table, the per-zone `ZoneLUT`, the UTC fast path for ICU `time_bucket`, `date_part` statistics) and the GROUP BY bucket rewrite that gives the perfect-hash aggregate dense integer keys for `date_trunc`, `time_bucket` (with or without a zone), `CAST(tstz AS DATE)`, `strftime`, `monthname`, `dayname`, `last_day`, numeric bins such as `floor(x / 100) * 100`, compositions such as `epoch_ms(date_trunc('hour', ts))`, date labels such as `year(ts) || '-Q' || quarter(ts)`, epoch arithmetic such as `floor(epoch(ts) / 3600) * 3600`, `date_bin`, CASE keys, keys behind `AT TIME ZONE`, nested keys, aliases defined in subqueries and CTEs (also wrapped in an outer expression or defined per UNION ALL branch), coordinate sets such as `year(ts), month(ts)`, and `SELECT DISTINCT`, gated by a perfect-hash budget that follows the aggregate state size; plus zone-table fast paths for scalar `time_bucket` with a zone, `strftime` (microsecond and nanosecond, constant and non-constant formats), interval arithmetic, `tstz - tstz`, `age`, `date_sub` and `date_diff` on TIMESTAMPTZ, and a rewrite of comparisons over monotone functions into column ranges that the inverted index consumes as range filters. `date_bin` now returns infinite inputs unchanged as PostgreSQL does.

## What changes in this repository

Four things, none of them large, plus the submodule pin:

- **pgwire text conversion** (`server/pg/serialize.cpp`, `server/pg/deserialize.cpp`): rendering a TIMESTAMPTZ took the session offset from an ICU calendar operation per value and now reads it from the zone table, and text input carrying no zone resolves the wall clock through the table's local-day view, which settles a repeated or skipped hour the way the engine settles it everywhere else. Rendering needs the zone's offset, which no calendar affects, so it reads the table whenever the session zone is not UTC; reading text back has to resolve a wall clock, so it uses the table only when the session calendar is gregorian and leaves any other calendar on the ICU path. Either direction falls back per value for an instant the table does not cover.
- **`date_bin`** (`server/connector/functions/math.cpp`): ours rather than DuckDB's, so it carries its own `bucket_rewrite` callback to take part in the GROUP BY rewrite. It also returned a wrapped finite value for an infinite input, where PostgreSQL returns the input.
- **Elasticsearch `date_histogram`** (`server/network/http/es/handlers.cpp`): the handler emitted the truncation twice, as the epoch key and as the formatted label, and grouped by both. One truncation as the group key with both derived above it is the same result and a shape the bucket rewrite can claim.
- **The submodule pin**, which is where everything else comes from.

## Coverage

One file per topic under `tests/sqllogic/sdb/pg/simple/temporal/`:

- `date_trunc_bucket_groups.test`: fixed and calendar units through the rewrite, pinned plans, result equivalence with the optimizer disabled, the year-1 and BC guards, leap-day and year-boundary rows, a second group column, the untouched plan for non-eligible shapes.
- `date_trunc_calendar_table.test`: every day from 1002 BC to 2325 AD against `make_timestamp(year(), month(), …)`, the 1899/1900 and 2299/2300 table boundaries, bucket round-trips.
- `date_trunc_bucket_groups_tz.test`: TIMESTAMP WITH TIME ZONE, pinned plans for Europe/Berlin (rewritten) and Asia/Kolkata (hour declined, day rewritten), the Berlin and New York fall-back hours, the Berlin spring-forward gap, the 25-hour local day, Lord Howe minutes, Sao Paulo quarters, equivalence with the optimizer disabled, errors of the internal functions on misuse.
- `date_trunc_bucket_groups_time_bucket.test`: `time_bucket` on TIMESTAMP, DATE and TIMESTAMPTZ (default origin, explicit origin, offset, month widths), `CAST(tstz AS DATE)` including `TRY_CAST`, the zone-string variant staying on the plain plan, Berlin fall-back and Kolkata dates.
- `date_part_timestamptz_groups.test`: perfect-hash plans for `year(tstz), month(tstz)` and `date_part('hour', tstz)`, the plain plan when the data holds infinity, bounds in Pacific/Kiritimati (+14).
- `date_trunc_bucket_groups_strings.test`: `strftime` formats that rewrite and the ambiguous `%I`-without-`%p` format that does not, DATE inputs with time specifiers, `monthname`/`dayname` naive and zoned with NULL and infinity groups, the Berlin fall-back hour, equivalence with the optimizer disabled, misuse errors.
- `date_trunc_bucket_groups_precision.test`, `date_trunc_bucket_groups_time_bucket_zone.test`: TIMESTAMP_NS/MS/S inputs with nanosecond boundary rows, `%G-%V` and `%y` formats, cross-frame casts, and `time_bucket` with an explicit zone across Santiago's midnight transitions, Berlin's fall-back hour, Lord Howe, Kiritimati, St. John's and Kolkata.
- `date_trunc_bucket_groups_zone_matrix.test`, `date_trunc_bucket_groups_strftime_matrix.test`, `date_trunc_bucket_groups_precision_matrix.test`, `bucket_groups_numeric_matrix.test`, `bucket_groups_label_matrix.test`: matrix tests that compute every shape with the optimizer disabled into a reference table and enabled into a second one, and pin both set differences as empty (26 zones times 73 shapes, 42 strftime formats times three input types, the three extra precisions, 71 numeric and 57 composition shapes, 38 label shapes times three input types). Zoned keys are rendered to text inside the query because the extended protocol prints TIMESTAMPTZ in UTC and the simple one in the session zone.
- `time_bucket_zone_scalar.test`: the scalar zoned `time_bucket` pinned from the ICU implementation before the zone-table fast path existed: checksums over 20k pseudo-random instants spanning 1900 to 2299 in 26 zones times 10 widths, per-row outputs around the transition windows of 11 zones, infinity and NULL rows, the unknown-zone error and a non-constant zone.
- `bucket_groups_epoch_case_matrix.test`, `bucket_groups_at_time_zone_matrix.test`, `bucket_groups_nested_matrix.test`, `bucket_groups_alias_distinct_matrix.test`: the same reference-versus-rewritten matrices for the epoch arithmetic, inverse conversion, `date_bin` and CASE shapes (102), for `AT TIME ZONE` and `timezone()` keys in three session zones (200 shapes each), for two-key nested buckets (47 shapes), and for keys aliased in subqueries and CTE chains, multi-reference declines and `SELECT DISTINCT` (335 shapes), with pinned plans.
- `monotone_predicate_matrix.test`: 4 282 predicate shapes on TIMESTAMP, TIMESTAMPTZ, DATE, TIMESTAMP_MS, BIGINT and DOUBLE columns (every comparison operator, BETWEEN, IN and NOT IN, OR/NOT/CASE/COALESCE contexts, aligned, unaligned, out-of-range, NULL, NaN and infinite constants) computed with the expression rewriter disabled and enabled, both set differences pinned empty, plus seven zones around their transitions with Antarctica/Casey pinned as declined, and the pushed range filters pinned in plans.
- `bucket_groups_coordinate_matrix.test`: 82 coordinate-set shapes (year/month, year/month/day/hour/minute, isoyear/week, year/quarter, strftime and date_part forms, aliases, DISTINCT, two inputs, HAVING) on TIMESTAMP, TIMESTAMPTZ and DATE, pinned plans, and a rotation through 23 session zones exercising the bounded zone table cache.
- `timestamptz_scalar_arithmetic.test`: `strftime`, `tstz ± INTERVAL`, `date_sub` and `date_diff` on TIMESTAMPTZ pinned around the transitions of ten zones (Havana, Berlin, Kiritimati, Lord Howe, Amsterdam 1937, Monrovia 1972, Kolkata, UTC, Santiago, St. John's), the constant-format and constant-part results checked inside the test against the per-row ICU path, infinity and NULL rows, dates outside the zone table, month-end pinning, a non-Gregorian calendar and the out-of-range errors; also `tstz - tstz`, `age`, decade/century/millennium differences, nanosecond `strftime` and non-constant formats and parts.
- `date_trunc_bucket_groups_budget.test`: the perfect-hash budget, pinned plans for an 18 500-bucket hourly histogram with a count (perfect hash) and with five aggregates (hash table), the same aggregates on days (perfect hash), and `last_day` groups on all three input types.

- `strftime_nanoseconds.test`: the `%n` specifier followed by further specifiers, naive and zoned, at a 1940 and an 1850 instant outside the zone table, and 3 000 rows stored through a table with constant and non-constant formats compared row by row.

- `bucket_groups_interval_matrix.test`: 41 arbitrary-width bucket shapes (`time_bucket` from 150 ms to 4 months with interval offsets, negative offsets, TIMESTAMP and DATE origins, naive, zoned and with an explicit zone; `date_bin` with an origin; `date_trunc` second/minute/hour; the epoch forms; a non-constant width) computed with the rewrite disabled and enabled, both set differences pinned empty, plus plans for the accepted shapes and the three declines (budget, non-constant width, wall-clock hour in a half-hour zone) and the widths `time_bucket` and `date_bin` reject.

- `bucket_groups_time_of_day_matrix.test`: 44 shapes in six session zones (Berlin, Havana, Lord Howe, Chatham, Santiago, UTC) computed with the rewrite disabled and enabled, both set differences pinned empty: the date-free `strftime` formats and their declines, TIME casts over hour, minute, day and `time_bucket` leaves, TIME columns at widths that do and do not divide a day, millisecond and microsecond truncation, `time_bucket(width, tstz, zone, origin)` at micros, day and month widths, and the grids whose phase is shifted away from midnight (`date_bin` with a seconds origin, `time_bucket` with an offset), which decline because a slot key carries no phase; plans, scalar values and the rejected widths pinned.

- `bucket_groups_combinations_matrix.test`, `bucket_groups_edge_data.test`: the cross product (every `date_trunc` unit against every input type, every `strftime` specifier alone plus a third of the ordered pairs of sixteen, every `time_bucket` width in every argument form, `date_bin`, the TIME widths and casts, the epoch, composition, label and coordinate shapes) computed with the rewrite disabled and enabled, 876 shapes with both set differences pinned empty and the 408 zoned shapes repeated in Lord Howe and Havana; and the values themselves on the data that breaks things (infinity, empty and single-row tables, year 1 to 9999, BC centuries, the DST hours) under ROLLUP, GROUPING SETS, windows, joins, DISTINCT, UNION ALL, HAVING, aliases, two perfect-hash thresholds and a non-Gregorian calendar.

- `bucket_groups_rewrite_coverage.test`: which shapes the rewrite claims, read out of `json_serialize_plan(sql, optimize := true)`, which reflects the optimizer and honours `disabled_optimizers`. 77 shapes in three session zones, each pinned with whether the plan serialized, whether a bucket function reached it and how many logical operators the pass added: every family that must fire next to every guard that must decline (ambiguous formats, an offset in the format, microsecond TIME casts, over-budget ranges, a non-constant width, a foreign leaf zone, half-hour zones, DISTINCT aggregates, GROUPING SETS and CUBE).

- `bucket_groups_distinct_root.test`: 54 `SELECT DISTINCT` shapes in three zones, values with the rewrite disabled against enabled in both set directions, and no plan that fails to serialize. A distinct that is its own plan root is the shape where the pass used to leave its root pointer on the operator it had already moved away.

- `zoned_scalar_out_of_range.test`: 48 zoned expressions against their constant-folded form for 15 instants in three zones. Folding evaluates through plain ICU while a column goes through the vectorized fast paths, so a disagreement is exactly the class of bug where a fast path answers for a row it cannot serve; the pre-1900 values are pinned as well.

- `pgwire_zone_text.test`: the text conversion itself, per session zone, in five zones and both directions: the rendered offsets across the repeated and the skipped hour, sub-minute local mean time, instants on both sides of the zone table's range, and the read-back of a zone-less wall clock. It renders through a file rather than the wire so the expectation does not depend on which protocol the runner picks; the pinned renderings were checked against the system time zone database.

- `cast_comparison_infinity.test`: 128 predicates comparing a date expression cast to a timestamp against infinite and non-midnight constants, with the expression rewriter disabled and enabled, both set differences pinned empty. It guards a bug the predicate fuzzer found in upstream's comparison simplification, fixed in the pinned fork: the enabled side used to raise "Can't get TIME of infinite TIMESTAMP" on a query that answered fine without it.

- `aggregate_common_types.test`: twelve aggregate lists whose duplicates the common-subexpression pass erases, each asserted to still produce a readable logical plan, and eight of them compared with that pass disabled against enabled. It guards the second upstream fix in the pinned fork; before it, six of the twelve could not be read back at all.

## Merge order

The submodule pointer check stays red until the fork side lands, so:

1. Merge serenedb/duckdb#95 into `v2026.07.10`.
2. Point `third_party/duckdb` here at the merged commit on that branch and push; the `check-submodule-pointers` hook goes green.
3. Merge this PR into `main`.

Both branches are rebased: the fork sits on the `v2026.07.10` tip (`d36f48aa1ac`, which is what `main` already pins), and this branch sits on `main`'s tip, so neither carries a conflict.


server/pg/serialize.h
```diff
@@ -32,6 +33,11 @@
 #include "basics/message_buffer.h"
 #include "query/config.h"
 
+namespace duckdb {
+
+class ZoneLUT;
+}
+
```
MBkkt, line 39:
```suggestion
namespace duckdb {

class ZoneLUT;

}  // namespace duckdb
```

MBkkt (approved):
I will check duckdb part separately

PR #1076, opened 2026-09-02
feat: embedded docs

mkornaukhov:
You can now search the SereneDB docs with nothing but a running serened. Start it with an HTTP listener that has `api=mcp`, point any MCP client at `/_mcp`, and ask:

```bash
serened ./data --listen='postgres://0.0.0.0:5432,http://0.0.0.0:9200?api=mcp'
claude mcp add --transport http serenedb http://127.0.0.1:9200/_mcp --header 'Authorization: Basic cG9zdGdyZXM6'
```

The docs are compiled into the binary, indexed into a BM25 search table at startup, and served through three MCP tools. The Markdown extension this builds on is vendored, so `md_to_text` and friends are available in SQL too.

What changed:
- **Markdown extension**: vendored `duckdb_markdown` and `cmark-gfm` as submodules, built into serened, with the upstream suite in CI (`tests/duckdb/run.sh --suite markdown`) and a wire-level sqllogic test.
- **Embedded documentation**: `scripts/generate_docs.py` compiles the Docusaurus sources into a C++ table (`SDB_EMBEDDED_DOCS`, default ON); at startup the loader builds `sdb_docs.docs` as a search table with a BM25 inverted index and an English stopword dictionary, rebuilding only when the embedded hash or layout changes, and the schema is read-only for everyone but the loader.
- **MCP server**: stateless MCP streamable HTTP at `/_mcp` (`api=mcp`) with `search_docs`, `read_doc` and `list_docs`; router `:param` segments no longer match `_`-prefixed pieces so ES `/:index` routes cannot shadow it.
- **HTTP layer cleanup**: `HttpStatus` and `RpcCode` enums replace bare integers across the ES, MCP and test handlers, and the shared SQL-quoting and body helpers moved from ES into `network/http/common`.
- **Tests and tooling**: `test_mcp_api.py` covers the protocol and all three tools, `sdb_docs.test` checks the table and index invariants plus the read-only guarantees, gtests follow the `HttpStatus` enum, and pre-commit skips `docs/`.
- **Misc**: `ConnectionContext` fields reordered to drop padding (288 -> 280 bytes).


server/network/listen_spec.cpp
```diff
@@ -128,9 +128,9 @@ void ApplyParam(ListenSpec& spec, std::string_view key,
       SDB_FATAL(GENERAL, "'api' is only valid on an http endpoint '", url, "'");
     }
     for (std::string_view a : absl::StrSplit(value, ',', absl::SkipEmpty())) {
-      if (a != "es" && a != "test") {
+      if (a != "es" && a != "test" && a != "mcp") {
```
gnusi: There will be more, generalize?

server/docs/docs_loader.cpp
```diff
@@ -0,0 +1,247 @@
...
+           result->GetValue(1, 0).GetValue<int32_t>() == kLayout;
+  }
+
+  bool Rebuild() {
+    for (const auto& sql : {
+           absl::StrCat("DROP TABLE IF EXISTS ", kTable),
+           absl::StrCat("DROP TABLE IF EXISTS ", kMeta),
+           absl::StrCat("DROP TEXT SEARCH DICTIONARY IF EXISTS ", kDictionary),
+           absl::StrCat("CREATE TEXT SEARCH DICTIONARY ", kDictionary,
+                        " (template = 'text', locale = 'en_US.UTF-8', "
```
gnusi: segmentation is better

gnusi:
1. I think we can use basics::WriteObject/basics::ReadObject for serde
2. Would be nice to have SQL functions to interact with docs (e.g. sections/function/type reference, smth else?)
3. We need to think of how this would interact with future upgrades/backups etc

mkornaukhov:
>     3. We need to think of how this would interact with future upgrades/backups etc
We store hash, so if on startup hash differs -> rewrite the database

MBkkt (approved):

PR #1086, opened 2026-09-06
fix: search table expression indexes bind their columns by the wrong ordinal

gnusi:
An expression index on a search table (`storage='search'`) is accepted by `CREATE INDEX` and then either breaks every `INSERT` or, worse, silently indexes the wrong column.

### Root cause

`CREATE INDEX` binds the index expression against the projection its scan produces, so each column reference inside it is a *position in that projection*. `NormalizeBoundExpression` then maps those positions to stable catalog column ids through `col_index_to_id`. The transactional path built that vector from the same projection (`BuildCreateIndexProjection`); the search table path built it from the table's column order:

```cpp
const auto col_index_to_id = IsDuckDBTable()
    ? make_column_ids(BuildCreateIndexProjection(_pk_positions, _info->column_ids))
    : make_column_ids(std::views::iota(size_t{0}, columns.size()));
```

So a reference to the indexed column resolved to whatever table column sat at its projection slot. Both paths now map through the projection.

### What it looked like

| table layout | before | after |
|---|---|---|
| `ts TIMESTAMPTZ_NS, svc, sev, body` | `INSERT` fails: `Failed to bind column reference "" [...]: inequal types (VARCHAR != TIMESTAMPTZ_NS)` | ok |
| `svc VARCHAR, body, ts, sev` | `INSERT` succeeds, index is built over `svc`; `@@` finds nothing and nothing says why | ok |
| `body, ...` | worked by coincidence | ok |

The type error reported against `main` is the lucky case; a VARCHAR column in first position gives a silently wrong index.

### Verification

`tests/sqllogic/sdb/pg/index/search_table_expression_index.test` covers the three layouts plus a transactional control and reads the matches back through the index relation (10/10 each). The existing index suite (`tests/sqllogic/sdb/pg/index`, 456 checks) and the search table tests pass on the fixed build.

Not in scope, noted while here: `@@` against a search *table* rather than its index relation errors by design (`search_table_create_index.test:143`, and the docs prescribe `FROM index_name`), but the hint still says to create an index when one already exists.

### DIFF
--- a/server/connector/duckdb_physical_create_index.cpp
+++ b/server/connector/duckdb_physical_create_index.cpp
@@ -367,11 +367,8 @@ SereneDBPhysicalCreateIndex::GetGlobalSinkState(
            std::ranges::to<std::vector<catalog::ColumnId>>();
   };
 
-  const auto col_index_to_id =
-    IsDuckDBTable()
-      ? make_column_ids(
-          BuildCreateIndexProjection(_pk_positions, _info->column_ids))
-      : make_column_ids(std::views::iota(size_t{0}, columns.size()));
+  const auto col_index_to_id = make_column_ids(
+    BuildCreateIndexProjection(_pk_positions, _info->column_ids));
   const auto relation_id = catalog::IdOf(_relation);
 
   // Normalize + serialize a bound expression (index key or partial-index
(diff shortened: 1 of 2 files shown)


Dronplane (approved):
LGTM

PR #1087, opened 2026-09-06
fix: lower search-table predicates through the table's inverted indexes

gnusi:
## Problem

`col @@ q` on a search table's own relation failed with "@@ requires an inverted-indexed column on one side" even when `col` had an inverted index; only `FROM <index>` worked. The same root cause left `tag = 'red'` as a post-scan Column Filter on the table while the index relation answered it from the term dictionary. The table scan resolved WHERE columns through the shard config by column id, but an inverted index's term field is allocated separately, so an indexed column was never claimable.

## Changes

- The table scan resolves columns through the table's inverted indexes, with the shard's own term dicts (the PK columns) as fallback, so the table form and the index relation lower to the same iresearch filter. `WithSearchGetters` takes the set of indexes; the NOT NULL check that drops a column's null marker keys by column id instead of term field.
- EXPLAIN renders the Index Filter for search-table scans. It was gated on the index relation, so claimed filters, PK term filters included, were invisible and tests pinned "no Column Filter line" as the signal. The field resolvers pick the owning index per field and map a null marker back to its column instead of printing `col<id>(null)`.
- Refactor: `inverted_index` doubled as "this scan is an index relation" (an index relation over a search table was `SearchTable` plus a pointer). `ScanEntryKind::SearchTableIndex`, `IsIndexRelation()` and `ScannedIndex()` make that explicit, and the bind data holds one `indexes` vector: the scanned index, or the search table's indexes.
- The bare-table hint mentions the index relation. `@@` on transactional tables still errors.

## Tests

- `search_table_match_on_table.test` (new): table vs index parity for `@@` (counts, combined predicate, expression index, ordered rows), multi-index `And(svc, Phrase body)` plan, PK + indexed predicate composition, indexed + residual Column Filter, NOT NULL vs nullable negation plans, bare-table and transactional errors.
- `search_table_create_index.test`: `@@` on the table returns rows; the `tag = 'red'` plan moves from Column Filter to Index Filter.
- `search_table_pk_pushdown.test`: PK term filters are now visible in the plans.

Each commit builds and passes `tests/sqllogic/sdb/pg/index` on its own; the final state also passes `pg/simple` and every other sqllogic file touching search tables or inverted indexes.

### DIFF
--- a/server/connector/duckdb_table_function.cpp
+++ b/server/connector/duckdb_table_function.cpp
@@ -44,6 +44,7 @@
 #include <iresearch/search/all_filter.hpp>
 #include <iresearch/search/vector_radius_filter.hpp>
 #include <iresearch/search/vector_similarity_filter.hpp>
+#include <ranges>
 
 #include "catalog/ddl/catalog.h"
 #include "catalog/entry/duckdb_index_scan_entry.h"
@@ -67,6 +68,7 @@ void CopyCommon(const SereneDBScanBindData& src, SereneDBScanBindData& dst) {
   dst.table_entry = src.table_entry;
   dst.entry_kind = src.entry_kind;
   dst.inverted_index = src.inverted_index;
+  dst.search_indexes = src.search_indexes;
   dst.stored_filter = src.stored_filter;
   dst.filter_scorers = src.filter_scorers;
   dst.snapshot = src.snapshot;
@@ -396,6 +398,17 @@ irs::Filter::ptr MakeVectorFilter(const VectorScorerOptions& vs,
   o->max_search_fanout = vs.max_search_fanout;
   o->inner = std::move(inner);
   return f;
+}
+
+std::vector<const catalog::InvertedIndex*>
+SereneDBScanBindData::InvertedIndexes() const {
+  if (inverted_index) {
+    return {&catalog::InvertedInfo(*inverted_index)};
+  }
+  return search_indexes | std::views::transform([](const auto& index) {
+           return &catalog::InvertedInfo(*index);
+         }) |
+         std::ranges::to<std::vector>();
 }
 
 namespace {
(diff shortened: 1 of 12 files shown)


Dronplane (approved):
LGTM

server/connector/duckdb_table_function.cpp
```diff
@@ -398,24 +399,46 @@ irs::Filter::ptr MakeVectorFilter(const VectorScorerOptions& vs,
...
 
-auto MakeFieldNameResolver(const SereneDBScanBindData& bind_data,
-                           const catalog::InvertedIndex& index) {
-  return [&bind_data, &index](catalog::ColumnId col_id) -> std::string {
+const catalog::InvertedIndex* OwningIndex(
+  std::span<const catalog::InvertedIndex* const> indexes, irs::field_id fid) {
+  const auto it = absl::c_find_if(indexes, [&](const auto* index) -> bool {
+    return index->LookupField(fid).entry;
+  });
+  return it == indexes.end() ? nullptr : *it;
```
Dronplane: Can this check really fire? E.g. we have unknown field but still got here and not bailed out earlier. Maybe ASSERT? And get rid of code that checks nullptr at call sites

gnusi: It does fire: a search table's PK term filter is claimed through the shard config, not through any inverted index (and a search table may have no inverted index at all, so `indexes` is empty), so no index owns that field and the resolvers fall back to the scan's own column names. That is a legitimate path rather than an assert.

Reworked in the last push so the lookup itself decides ownership: `FindFieldOwner` returns the index together with its `FieldLookup`, and the call sites no longer test for null (`index` is only dereferenced under `lookup.entry`). It also fixes a slip the old shape had: `LookupField` registers the PK field with a null entry and a valid `entry_field_id`, so testing `.entry` skipped the `kPKFieldId` branch on index relations.

PR #1088, opened 2026-09-06
fix: view-backed index crashes when the view spells file columns in a different case

gnusi:
## Problem

`SELECT Body FROM ixr LIMIT 1` over an inverted index on `CREATE VIEW vr AS SELECT timestamp AS Ts, body AS Body FROM read_parquet(...)` crashed the server whenever the parquet file spells its columns `Timestamp`/`Body`, the usual OTel export layout. DuckDB binds file columns case-insensitively, so the view is valid, but the fast path keyed the file schema by exact name and dereferenced the miss: `SDB_ASSERT` in assert builds, SIGSEGV in release. The "not a recognised fast-path source" guard never applies here because a column projection over `read_parquet` is a recognised fast path.

## Fix

`InitProjection` takes the source's column names and resolves each projection name once: exact match first, case-insensitive fallback; a miss raises `undefined_column`. The three per-source lookup maps (file reader, attached catalog, serenedb table) are gone. Exact-first keeps serenedb tables, which bind case-sensitively today, right while file readers and attached catalogs get DuckDB's case-insensitive semantics.

## Tests

`inverted_index_view_column_case.test`: capitalised parquet columns referenced in lowercase by the view; real-column materialisation through the index, an `@@` lookup returning the `INCLUDE`d column, and a count. The report's four statements pass verbatim on such a file. `pg/index` and the site-docs inverted-index suites pass.

### DIFF
--- a/server/connector/index_source_view.h
+++ b/server/connector/index_source_view.h
@@ -21,10 +21,8 @@
 #pragma once
 
 #include <absl/functional/function_ref.h>
-#include <absl/hash/hash.h>
-#include <absl/strings/ascii.h>
-#include <absl/strings/match.h>
 
+#include <duckdb/common/case_insensitive_map.hpp>
 #include <duckdb/common/types.hpp>
 #include <duckdb/common/types/data_chunk.hpp>
 #include <duckdb/execution/expression_executor.hpp>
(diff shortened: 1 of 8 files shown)


server/connector/index_source_view.h
```diff
@@ -20,22 +20,63 @@
...
+  }
+};
+
+class SourceColumns {
+ public:
+  template<std::ranges::input_range Names, typename Proj = std::identity>
+  explicit SourceColumns(const Names& names, Proj proj = {}) {
+    duckdb::idx_t index = 0;
+    for (const auto& candidate : names) {
+      _by_name.try_emplace(std::string_view{std::invoke(proj, candidate)},
```
Dronplane: is it safe to store string_view? E.g. does we have guarantee that proj returns not temporary?

Dronplane (approved):
LGTM  with one remark

server/connector/index_source_view.h
gnusi: Fair catch. The `string_view` keys relied on every projection viewing storage that outlives scan init (the reader's bind data, the catalog columns), which held for the three call sites but nothing enforced it: a projection returning a temporary `std::string` would have dangled. Since f833eabfb the map is a `duckdb::case_insensitive_map_t<idx_t>`, so the names are copied into `std::string` keys and a temporary from `proj` is fine.

Issue #97, opened 2025-12-24
Move sqllogictest-rs in third_party


MBkkt:
Otherwise it's not convenient to exclude third_party code from search in vscode

PR #1089, opened 2026-09-06
fix: catalog-split stress-suite findings + keyword-fold regression

pashandor789:
Fixes for the bugs gnusi's catalog stress suite (#1067) surfaced, plus one regression Andrey root-caused. Stacked on #1067 (it modifies the suite's classifier and a recovery test, so it's based on that branch). Paired with the duckdb fork PR **serenedb/duckdb#97** (unmerged), which the repin here points at.

## Fixes

- **Sequence counter survives a rolled-back DROP.** `CatalogStore::DropSequence` (wipes the counter + appends a log record) fired immediately, so `ROLLBACK` didn't undo it. Deferred to commit via `DeferDropAction` at all three user drop sites.
- **Permanent `TransactionException`s no longer look like a retryable 40001.** Three of them can never succeed on retry — a cross-database write, a read-only write, `CHECKPOINT` with local changes — so a class-40 retry client loops forever. The fork (#97) now attaches a structured `error_subtype`; the facade reads it and maps them to `0A000` / `25006` / `25001`. The stress classifier accepts those as expected permanent rejections (marker-gated, so all of its existing unit tests still pass; +4 new).
- **Keyword-lexed identifiers fold.** A column name/alias that lexes as a keyword (`timestamp`, `time`, `interval`, `text`, ...) reached the case-sensitive catalog in its source case, while identifiers folded — so `SELECT Timestamp FROM (… AS Timestamp)` failed. Fixed in the fork (#97, at the single `KeywordParseResult` site); this side adds the test.

## Validation
- Repro for each fixed and re-checked; `column_name_case` + `sequence_counter_after_rollback` pass on both pg-wire engines; RBAC (~39) and DDL (~55) suites unchanged; classifier unit tests 283 pass; stress smoke/soak/scale/attach/iceberg-chaos/remote all green.

## Not addressed here (owner-level, documented)
The recovery-wedge + compaction-SIGSEGV cluster the suite also found (unbounded catalog-WAL replay on boot; `CheckpointEntriesOf` null-deref under concurrent DROP). Root-caused with perf + deterministic repros, but they're catalog-durability design work.


pashandor789:
Wrong shape — folding these fixes into #1067 (pushed onto gnusi/catalog-stress-suite) instead of a separate stacked PR.

PR #1091, opened 2026-09-07
fix: DML on an indexed search table keys off the wrong column

gnusi:
## Problem

Two defects on DML against search tables, found while writing metrics tests:

1. **Key taken off the wrong column once the table has an index.** A search table's DML scan appends its row-identity virtuals as `[pk columns..., indexed non-PK columns..., generated pk]` (`BuildRowIdColumns`). Since #1050 a search table can carry storage-less inverted indexes, so that tail is no longer just the key, but `PlanDelete` and `PlanUpdate` still took the PK slots off the end of the child's columns. On an explicit-PK table with an index on another column, `DELETE`, `UPDATE`, `DELETE ... RETURNING` and an in-transaction `TRUNCATE` (row-wise delete) either failed with `Expected unified vector format of type INT64, but found type VARCHAR` (TEXT index) or reported success and deleted by the indexed value instead of the key (BIGINT index), which removes the wrong row when values collide. Generated-PK tables, an index on the key itself, INSERT, autocommit TRUNCATE and MERGE (rejected up front) were unaffected.
2. **Virtual-id collision, older.** PK virtuals were encoded as `VIRTUAL_COLUMN_START + column position`; position 2 equals `MultiFileReader::COLUMN_IDENTIFIER_FILE_INDEX`, which the scan classifies before the generic virtual branch. Any search table whose key (or indexed column) is its third column failed DML with `column "file_index" is only served by glob-backed inverted indexes`, index or not.

## Fix

- `SearchPkSlots` derives the key's slots from the entry's own `GetRowIdColumns()` order; both planners use it, and `PlanUpdate` passes through exactly those slots.
- PK virtuals move to a private id range (`kColumnIdentifierPkVirtualStart`, `PKVirtualColumnId`), decoded by `VirtualToPKColumnIndex`; no overlap with DuckDB's multi-file identifiers.

## Tests

`index/search_table_dml_indexed.test`: explicit PK with TEXT and BIGINT indexes where the indexed value collides with another row's key (delete, update, RETURNING, in-transaction TRUNCATE, with the index relations checked for removed and re-added terms), the key as the third column without any index, a composite key with an index on a non-key column, and a generated key with an index. `pg/index`, `pg/dml`, `pg/system` and the search-table recovery tests pass.

## Also: TRUNCATE inside a transaction

Autocommit `TRUNCATE` on a search table clears the store, but inside a transaction it was planned as a row-wise delete over the scan, which only sees published rows: rows committed but not yet refreshed survived, and so did rows inserted earlier in the same transaction (`INSERT; TRUNCATE` in one transaction kept the row). The in-transaction form is now a remove-all query in the transaction's serial iresearch transaction, ordered among its other writes: earlier writes go, later ones stay, `ROLLBACK` discards it. The WAL keeps recording it as a truncate op in order, and replay implements it the same way rather than as a `Clear` at the record tick, which used to discard the same record's later inserts.

Tests: `index/search_table_truncate_tx.test` (unpublished rows, insert-truncate-insert in one transaction with the index relation checked, rollback, autocommit) and `recovery/search_table_truncate_tx_recovery.test` (crash after the WAL commit of an insert-truncate-insert transaction; only the post-truncate row survives replay). All search-table and TRUNCATE recovery tests pass.


MBkkt (approved):

Dronplane (approved):
LGTM

PR #1095, opened 2026-09-07
perf: adaptive MaxScore disjunction when block maxima cannot prune

gnusi:
## Problem

With `optimize_top_k`, a scored `ts_any` over two common terms was slower than the same query without the option: 34.9 ms vs 16.1 ms single-threaded on identical 20M-row tables (release build). Profile: 67% of the time was self time in `PostingPrunedDisj::ScoreCandidates`.

Causes:

1. `ScoreCandidates` matched candidates against a block with `std::find` from the block start and never advanced on a miss. Both sides are sorted, so a block whose documents mostly miss the candidates (the normal situation for a non-essential term) did quadratic work.
2. MaxScore partitions lists into essential and non-essential by block maxima. When the non-essential term's maximum keeps every essential document competitive, `FilterCompetitive` removes nothing and every candidate is probed doc-at-a-time, while the unpruned `WindowDisjunction` fills windows with SIMD. No bound can fix that: a document could legitimately contain both terms.
3. Once promotion routes dense windows through `Fill`, the pruned leaf's `Fill` scattered a block with a sequential loop, where `PostingLeaf::AddWhole` walks a full block through `VisitDocs<kBlockSize>` (eight interleaved chains, overlapping stores).

## Change

- `ScoreCandidates`: gallop then `lower_bound`, and advance past misses.
- `MaxScoreDisjunction`: count per outer window how many candidates reach `ScoreCandidates`. When at least three quarters of candidates × non-essential lists get probed, run the next windows with every list essential, which is the window fill path and still skips windows whose summed maxima cannot beat the threshold. Exponential backoff up to 64 windows, then one normal window to resample.
- `PostingPrunedDisj::Fill`: scatter a full block through `VisitDocs<kBlockSize>`, like `AddWhole`.

## Numbers

Release build, identical data, single thread, `LIMIT 10`, min of 5 interleaved runs:

| shape | with option, before | with option, after | without option |
|---|---|---|---|
| `ts_any` two 50% terms, 20M | 34.9 ms | 16.6 ms | 16.2 ms |
| `ts_any` 50% + 97% term, 20M | 5.9 ms | 6.1 ms | 25.0 ms |
| conjunction, 20M | 20.5 ms | 20.5 ms | 39.9 ms |
| single term, 20M | 1.6 ms | 1.6 ms | 4.1 ms |
| `ts_any` five terms, 2M | 4.2 ms | 4.2 ms | 4.5 ms |

Results are identical to the unpruned tables for k = 10, 100 and 1000 on every shape. The 2M-row tables keep the option's fixed per-segment setup cost of a fraction of a millisecond, unchanged by this PR.

## Test

`sdb/pg/index/inverted_index_score_prune_dense_disjunction.test`: dense two-term disjunction under the option, pruned plan, tied top 10 by id, parity with a full sort at k = 100 and 1000. All scorer and prune sqllogic suites and the recovery prune tests pass.

### DIFF
--- a/libs/iresearch/include/iresearch/search/top/max_score_disjunction.hpp
+++ b/libs/iresearch/include/iresearch/search/top/max_score_disjunction.hpp
@@ -65,6 +65,10 @@ class MaxScoreDisjunction : public Root {
     _num_candidates = 0;
     _num_outer_windows = 0;
     _min_window_size = 1;
+    _window_candidates = 0;
+    _window_probes = 0;
+    _dense_left = 0;
+    _dense_streak = 0;
 
   outer:
     while (window_min < max) {
@@ -98,6 +102,7 @@ class MaxScoreDisjunction : public Root {
 
       window_min = std::min(Top(), window_max);
       ++_num_outer_windows;
+      UpdateDensity();
     }
     _admit.Flush(collector);
   }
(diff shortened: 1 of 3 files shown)


MBkkt:
Closed because of this: https://github.com/serenedb/serenedb/pull/1098
test was adopted, query measured

PR #1097, opened 2026-09-08
fix: term-dictionary and offsets pushdowns on search tables

gnusi:
Three planner bugs on search tables, all with the same root cause: code that resolved the scanned index with `ScannedIndex()` or keyed a lookup off the column id, where a search table's plain columns live under an index-allocated term field and a bare table scan carries no index at bind time. One commit per bug.

### 1. `ts_dict_*` aggregates crash on a bare search table
`SELECT unnest(ts_dict_agg(col)), unnest(ts_dict_count(col)) FROM t` (a search table, not its index relation) asserted in `SereneDBScanBindData::ScannedIndex()`. The planner now resolves the table's inverted indexes, picks the one whose term dictionary covers the column and reads its allocated term field; filter pushdown routes `TsDict` mode on a bare table into the dictionary claim (faceted `WHERE` works, also across several indexes on one table); term-level comparisons on the enumerated column map to the allocated field; EXPLAIN prints the column name for `TsDict` instead of the raw field id (this also fixes the display on search-table index relations). A column no index covers gets the existing `column has no text term dictionary` error.

### 2. Dictionary `GROUP BY` / `count(DISTINCT)` / `min` / `max` rewrites skipped search tables
The facet pushdown required an index relation and checked `IsKeywordField` with the column id, so keyword facets on search tables ran as hash aggregates over document rows. It now adopts bare search-table scans, resolves each key column to its covering index and allocated term field, and resolves `WHERE` predicates against every index on the scan.

### 3. `ts_offsets` on search tables
The bare-table scan hit the same `ScannedIndex()` assertion, and the recomputed-offsets path (a dictionary without `offset = true`) bound the column id into `OffsetsBindData`, failing with `Field id ... not found in the index definition` on the index relation as well. Both paths now use the covering index's allocated term field.

### Tests
`search_table_ts_dict.test` +23 cases (bare-table aggregates, `unnest` shape, faceted `WHERE`, multi-index table, clean error, facet rewrites and `GROUPING SETS` facets -- keys in one or two indexes, one nullable key with its synthesized NULL group, two nullable keys declining -- with three EXPLAIN pins) and `search_table_ts_offsets.test` +6 cases (stored and recomputed offsets through the bare table). Both pass under pg-wire-simple and pg-wire-extended; transactional-table behaviour is unchanged (identity mapping when no term field is allocated). Related files worth including in the CI sweep: `ts_dict*.test`, `ts_dict_facets.test`, `inverted_index_*offsets*`.

### DIFF
--- a/server/connector/duckdb_table_function.cpp
+++ b/server/connector/duckdb_table_function.cpp
@@ -614,7 +614,7 @@ void SereneDBScanBindData::AppendSummary(
         ? std::string{"Index Filter"}
         : absl::StrCat(
             "Index Filter(",
-            display_field(static_cast<catalog::ColumnId>(req.field_id)), ")");
+            display_field(static_cast<catalog::ColumnId>(req.display_id)), ")");
     out.insert(std::move(key), duckdb::ExplainValue(irs::ToExplainNode(
                                  *req.having_filter, name_of, kind_of)));
   }
@@ -667,7 +667,7 @@ void SereneDBScanBindData::AppendSummary(
   if (TsDictMode()) {
     auto names =
       absl::StrJoin(ts_dicts | std::views::transform([&](const auto& req) {
-                      return display_field(catalog::ColumnId{req.field_id});
+                      return display_field(catalog::ColumnId{req.display_id});
                     }),
                     ", ");
     out.insert("TsDict", std::move(names));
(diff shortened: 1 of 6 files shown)


Dronplane (approved):
LGTM

PR #1098, opened 2026-09-08
perf: one boolean per strategy

MBkkt:
## Problem

Every boolean shape had its own container. A conjunction, a disjunction, an
exclusion, a boost, a min-match and a pure negation each carried a dedicated
class per read API, so `count`, `docs`, `scored` and `top` held four parallel
families of near-duplicates: sixteen threshold containers across six
directories, separate sparse and window conjunctions and disjunctions,
separate makers for each, and adapters wherever one API's node had to appear
inside another's.

That cost more than the duplication. A shape only ran well if someone had
written its container, so exclusions probed per document instead of filling a
window, a wide expansion in probe position re-probed every leaf per candidate,
min-match never pruned, and the strategy choice was spread across seven
dispatchers that could not be compared with each other.

## Change

Every boolean shape is now an instantiation of the read API's `BooleanWindow`,
`BooleanSparse` or `BooleanBitset` with group slots for the lead, the others,
the optional clauses, the excludes, the table filter and the score. Groups
(`OrGroup`, `ThresholdGroup`, `TallyGroup`, `probe::*Leaves`,
`fill::SetLeaves`) fill those slots, and one shared builder classifies the
query and picks the strategy. A dedicated container survives only where it is
measured to win, and the measurement is in the commit that keeps it.

The work is 39 commits, each gated on its own. The larger ones:

- **The three containers per API**, replacing the per-shape roots (`top`,
  `scored`, `docs`, `count`, and the `lead`/`probe`/`fill` nodes).
- **Min-match as a window group**, bitplanes at `k = 2` and counters above,
  deleting the sixteen threshold containers.
- **A window-form scored probe** for a wide disjunction, which was the 3x to 7x
  loss on a fuzzy or prefix expansion in probe position.
- **Exclusions as a cost model** in cycles rather than a hand rule: probes,
  lazy window, eager bitset, window lead or a count/docs fold, chosen per
  query, with a dense lead folding its excludes and skipping clean blocks.
- **Pruning coverage**: conjunctions of bounded disjunction children, probe
  groups as pruned clauses, constant multiterm clauses, excludes under a pruned
  root, dense-leaf promotion in the pruned disjunction, and a min-match
  disjunction over a stopword. The algorithm names are gone from the files and
  classes: these are `Pruned*`, because the shapes are a mix rather than any
  one paper.
- **Statistics as prepare-time fields** on `QueryBuilder` (`_postings`,
  `_leaves`, `_estimate_matches`) instead of virtuals, with flattening in
  `PrepareSegment` so no planner expands a child's terms.

Net: 121 files deleted, 33 added, about 13.3k lines removed against 9.2k added.

## Numbers

`search-benchmark-game`, 962 queries, five commands, patch over main on the sum
of per-query minima. Below 1.000 is faster.

| shape | queries | COUNT | DOCS | SCORED | TOP_100 | TOP_100_COUNT |
|---|---|---|---|---|---|---|
| term | 1 | 1.000 | 0.993 | 1.012 | 0.992 | 1.001 |
| phrase | 300 | 0.987 | 0.974 | 0.984 | 0.977 | 0.983 |
| intersection | 300 | 1.007 | 0.993 | 1.000 | 1.012 | 0.986 |
| union | 301 | 1.006 | 0.988 | 0.977 | 0.864 | 1.002 |
| intersection_union | 40 | 1.036 | 0.994 | 0.978 | 0.988 | 0.985 |
| negated | 19 | 0.520 | 0.513 | 0.571 | 0.612 | 0.621 |
| two-phase-critic | 1 | 1.005 | 0.993 | 0.993 | 0.999 | 1.004 |
| **all** | **962** | **0.991** | **0.979** | **0.981** | **0.950** | **0.991** |

Negation is where the exclusion model shows: `+bass -fish -guitar` goes from
528us to 84us on COUNT and from 549us to 125us on TOP_100. Pruned disjunctions
carry TOP_100: `george r brown convention center` 2804us to 1098us,
`los angeles daily news` 1715us to 729us, `graduate management admission test`
1219us to 657us.

Three cells sit above 1.000 by more than a percent and none of them is code:

- `intersection_union` COUNT 1.036 is forty queries of five to eight
  microseconds each, 251us against 260us in total. That is the harness's
  one-microsecond timing step, not a measurement.
- `intersection` COUNT 1.007 and TOP_100 1.012, and `union` COUNT 1.006, are
  driven by conjunctions and disjunctions over a stopword. Re-measured one
  query at a time under `perf stat -r 3` at 150 repetitions, all of them come
  back between 0.98 and 1.02 **at instruction counts identical to main within
  0.2%**, which is layout, not work. The harness runs a query next to its
  neighbours and a stopword list gets evicted between passes.
- `term` SCORED 1.012 is the single query `the`, 1626us against 1645us.

The table predates the last two performance commits, which were measured in
isolation instead.

### Count conjunctions with a dense probe

The one genuine regression the aggregates showed, up to 1.067 on
`+the +movement`. It was a register spill: the sparse count root took the probe
miss as an early `continue`, clang ran out of registers and spilled the
accumulator to the stack, so every matched document paid a read-modify-write.
Making the match the taken branch, which is the shape main has, returns the
register: `+the +movement` 1.067 to 1.005, `+the +progressive` 1.036 to 0.974,
`+the +english +restoration` 1.033 to 1.002, `+walk +the +line` 1.030 to 0.987.
Instructions were already within 0.2% of main on all of them, so the whole
difference was the spill.

### Dense disjunctions under `optimize_top_k`, the shape in #1095

A pruned disjunction sizes its outer window to the end of the current block of
every essential leaf, because a block bound is what the max-score split may
trust. On two dense clauses a block spans about 256 documents and
`_min_window_size` only grows while candidates are sparse, so the window stayed
one block wide: the root paid `UpdateWindowScores`, a `Split` sort, a
`make_heap` and a 64-word drain scan for every 256 documents, sixteen times
over what the exhaustive root pays for the same span. The widening to a full
window already existed but sat inside the branch a successful `Promote` takes,
so it never ran when `Split` itself returned every clause essential, which is
precisely the dense case: nothing is non-essential, no bound is being trusted
for a skip, and the window need not stop at a block.

20M rows, one thread, `LIMIT 10`, minimum of five runs:

| | before | after | without the option |
|---|---|---|---|
| `ts_any` two 50% terms | 25.5 ms | **15.7 ms** | 14.9 ms |
| ratio to unpruned | 1.73x | **1.06x** | — |

Per query that is 592M instructions down toward the exhaustive root's 343M.
#1095 reports 34.9 ms before and 16.6 ms after against 16.2 ms unpruned for the
same shape. Across 36 `queries.txt` queries spanning unions, conjunctions,
phrases, negations and the scored and count commands, measured against the
previous commit at 120 repetitions, cycles come back at a median of 0.997 and a
maximum of 1.017 with instructions at a median of 1.000, so nothing on the
benchmark moved. #1095's own sampling criterion was ported here first and
measured: it gains nothing on this branch, because `Split` already returns
every clause essential on that shape so there is no probing left for it to
detect, and it costs `album^2 military` 9.4% more instructions.

## Correctness

- Parity against main on 1573 queries across the five commands: COUNT, the docs
  hash, the scored hash and the top-100-with-count hash are byte-identical.
- The top-100 hash differs on four queries. All four return the **same
  documents**, with scores apart by 1.3e-7 to 1.8e-7 relative, which changes the
  checksum; on one of them that reorders a pair of tied documents. Declared as
  a rounding change and verified by printing both engines' hundred hits and
  comparing sets, order and scores.
- The load test reference was regenerated: 306 of 1549 entries report a
  different number of documents walked, because the window a pruned shape uses
  changed. Every document and score comparison passed, apart from the one tied
  pair above, and the regenerated reference reproduces three runs in a row.
- Adds `inverted_index_score_prune_dense_disjunction.test` from #1095, which
  pins the dense two-term disjunction shape: the pruned plan, the tied top ten
  by id, and parity with a full sort at k of 100 and 1000.


mkornaukhov (approved):
OK

PR #1150, opened 2026-09-11
feat: a typed object catalog over the embedded docs

gnusi:
## Why

`sdb_docs.docs` is **page shaped**: rows are documentation pages and headings keyed by file path. Nothing in it says what *kind* of thing a name refers to, so a name lookup is ambiguous:

- `sdb_docs.reference('create')` returns a **LangChain Python method**, because the only link between a name and its docs is a fuzzy title match.
- There is no way to ask "what aggregate functions exist?" or "does `ts_phrase` exist, and what is its signature?" without searching prose.

## What

`sdb_docs.objects()` derives one row per documented object from the same markdown, and `sdb_docs.summary()` is the one line description it uses. **No new tables** — both are SQL functions over `sdb_docs.docs`, classifying by path prefix and heading shape.

| kind | count |
| --- | --- |
| function | 628 |
| setting | 166 |
| statement | 45 |
| type | 31 |
| tokenizer | 28 |
| index_type | 2 |

892 of the 900 carry a summary. Three cases needed care:

- a `split: headings` title row opens with its own `# H1`, which has to be skipped **without** letting a container heading borrow its child's prose (`Scorers` correctly has no summary);
- aggregates document themselves with a `| **Description** |` table rather than a lead paragraph, which recovers 71 of 77 otherwise empty function summaries;
- index types spell their identifiers only in the **link target**, not the label, so `inverted` and `art` come from `./inverted/index.md` and `./art.md` rather than "Inverted Index".

The regex allows an optional schema prefix so the new page's own `sdb_docs.objects()` heading classifies, i.e. the feature appears in its own catalog.

## MCP

`list_objects` and `describe_object`. The three existing tools are all keyed on doc paths, so an agent cannot cheaply confirm a function exists before emitting SQL against it. `describe_object` reports **every** object carrying a name, since one name can be a function and a tokenizer at once (`minhash`). Asking for `to_tsvector` now returns a clear "no such object" rather than nothing useful.

Measured: the whole function vocabulary is ~20k tokens with summaries, ~1.7k as bare names.

## Also

- Documents the `sdb_docs` schema, which had **no page at all** — this covers the four pre-existing functions too.
- Folds the SQL in `docs_loader.cpp` into named constants so the statements read as SQL instead of `StrCat` fragments. `Rebuild()` goes from ~90 lines to 9.
- `kLayout` 12 -> 13, so existing datadirs rebuild.

## Testing

- `sdb/pg/system/sdb_docs.test`, `sdb_docs_objects.test` (new, structural invariants over every row) and `site_docs/sql/functions/docs.test` (new, nine runnable examples) pass on **both** pg-wire-simple and pg-wire-extended.
- `test_mcp_api.py`: 31 passed, including 8 new cases.
- Verified that the layout bump actually triggers a rebuild against an **existing** datadir, not just a fresh one.

## Depends on

serenedb/duckdb_markdown#1. `objects()` fans `md_extract_tables_json` across worker threads, which makes an existing abort in cmark-gfm reachable: `cmark_gfm_core_extensions_ensure_registered()` guards itself with a plain `static int`, and `cmark_register_node_flag()` calls `abort()` when a flag is registered twice. Measured at 1 abort in 10 cold starts before the fix, 0 in 35 after.

**This should land after that submodule fix and the pin bump**, or CI here may flake intermittently.

### DIFF
--- a/docs/sql/functions/docs.md
+++ b/docs/sql/functions/docs.md
@@ -1,0 +1,134 @@
+---
+title: Documentation Functions
+split: headings
+---
+
+import SqlLogicTest from "@site/src/components/SqlLogicTest";
+
+SereneDB ships its own documentation inside the server binary. At startup the server
+builds the `sdb_docs` schema from it, so the reference material always matches the
+running version and is available with no network access.
+
+The schema holds two tables and a set of functions over them:
+
+| Object | Description |
+| :--- | :--- |
+| `sdb_docs.docs` | One row per documentation page or section, with a full-text index over it. |
+| `sdb_docs.meta` | The hash and layout version of the embedded documentation. |
+
+`sdb_docs` is read-only for every user, including superusers. The server rebuilds it
+when the embedded documentation or its layout changes; writing to it raises an error.
+
+## Documentation paths
+
+Every row is addressed by a `path`. A page indexed as a whole is addressed by its file
+name, and a page split into sections is addressed by its file name followed by the
+chain of headings that leads to the section:
+
+```text
+sql/indexes/index.md#Indexes
+sql/functions/timestamp.md#Timestamp_Functions#Scalar_Timestamp_Functions#date_trunc(part,_timestamp)
+```
+
+Spaces in a heading become underscores and a literal `#` in a heading is escaped as
+`\#`, so the unescaped `#` characters count the heading depth. Paths returned by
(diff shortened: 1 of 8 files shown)


gnusi:
Parking this for now. It depends on serenedb/duckdb_markdown#1, which is landing on its own, and the catalog work should come back as its own change once the markdown pin has moved.

Branch `gnusi/docs-object-catalog` is kept, so reopening is just a matter of reopening this PR.

PR #1151, opened 2026-09-11
fix: register cmark-gfm core extensions at load time

gnusi:
Bumps `third_party/duckdb_markdown` by one commit to pick up serenedb/duckdb_markdown#1.

## The bug

`cmark_gfm_core_extensions_ensure_registered()` guards itself with a plain `static int`:

```c
void cmark_gfm_core_extensions_ensure_registered(void) {
  static int registered = 0;
  if (!registered) { cmark_register_plugin(core_extensions_registration); registered = 1; }
}
```

and `cmark_register_node_flag()` calls `abort()` when a flag is registered twice:

```c
if (*flags) {
  fprintf(stderr, "flag initialization error in cmark_register_node_flag\n");
  abort();
}
```

Every markdown parsing entry point calls `ensure_registered()` lazily, never at load. Two threads that reach it for the first time together both see `registered == 0`, both register, and the second one aborts the process.

This is reachable from plain SQL: **any first concurrent use of the `md_*` functions can take down the server.** It is not specific to any one caller, though a query that fans `md_extract_tables_json` across worker threads hits it far more readily than a scalar call.

## Measured

Cold start, then a query whose first markdown call is parallel over ~2.5k rows:

| Pin | Cold starts | Aborts |
| --- | --- | --- |
| `a0e5969` (current) | 10 | **1** |
| `cf9c0fd` (this PR) | 35 | 0 |

TSan on the failing run:

```
ThreadSanitizer: data race third_party/cmark-gfm/src/node.c:27 in cmark_register_node_flag
Location is global 'CMARK_NODE__TABLE_VISITED' of size 2
flag initialization error in cmark_register_node_flag
```

## The fix

Register during extension load, while still single threaded. `LoadInternal` runs from `DatabaseInstance::Initialize`, before any listener binds, so by the time queries run `registered` is already 1 and the lazy call sites only ever read it. That closes the window by construction rather than narrowing it.

## Note for review

The submodule PR targets `v2026.08.29`, the branch this pin follows. It should also be forward-ported to that repo's `main`, or it will be lost at the next version bump.

This pin currently points at the PR branch commit. If serenedb/duckdb_markdown#1 is squash-merged, the pin needs re-pointing at the merged SHA before this lands.


mkornaukhov (approved):

gnusi:
serenedb/duckdb_markdown#1 is merged as `fa41de9` on `v2026.08.29`, and the pin here now follows that commit rather than the PR branch head, as the description said it would need to.

Both required checks are red for reasons that predate this PR, verified against its own base `6c16cacfc`:

- **pre-commit** fails on `tests/libs/iresearch/parser/lucene_parser_test.cpp`, which has 8 outstanding `clang-format` replacements on `main` itself. It will fail the same way on every PR opened against `main` until someone reformats that file.
- **serenedb | build & test** is red on `main` at `6c16cacfc` too (dev, asan, tsan and perf all fail there). On this branch the dev job failed `sdb/pg/index/inverted_index_view_duckdb.test` and `sdb/clickhouse/write_extra_chscan.test_slow`, neither of which touches markdown parsing.

This change is four lines in a vendored submodule that move `cmark_gfm_core_extensions_ensure_registered()` from the lazy per-parse call sites to extension load. Merging.

Issue #1113, opened 2026-09-09
Variadic phrase slot document frequency should be max over alternatives, not sum


MBkkt:
## Idea

When a phrase slot can be filled by several alternative terms, the slot needs one document frequency derived from its alternatives'. Today the alternatives are accumulated into the slot one after another, which is effectively a **sum**, and summing is wrong in a specific way: it can exceed the collection size, and it makes the slot's contribution grow with the *number* of alternatives rather than with what actually matched. Widening an expansion inflates every document the phrase matches.

`max` is the right default. It can never exceed the collection size, it is stable against the rare tail of an expansion — one obscure alternative cannot swing the slot — and it is what Lucene's blended term does.

## This is already settled elsewhere in the engine

#1093 made a fuzzy expansion share one stats slot with `df = max` and `ttf = sum` over the matched terms, via `BlendedTermsCollector`, specifically because the previous summed `df` shrank `idf`. The variadic phrase slot is the same problem and did not get the same fix, so the two paths now disagree: the identical set of alternatives is blended as `max` when it arrives as a fuzzy term and summed when it arrives as a phrase slot.

## Done looks like

- A variadic slot's document frequency is `max` over its alternatives, and `ttf` is summed, matching the fuzzy path.
- The score no longer moves when alternatives are added that nothing matches.
- The same expansion yields the same slot statistic whether it reaches the scorer as a fuzzy term or as a variadic phrase slot — with a test asserting exactly that.
- Tests with slots whose alternatives span a wide document-frequency range.

## Note

Team work, not a student task. Still present after #1093, which fixed the same shape of problem for fuzzy but not for the phrase path.

## Links

- Blocks #1107; informs #1109.
- #1112 — the other half of variadic slot scoring: that the slot contributes once at all.

MBkkt:
Superseded by #445, and the premise needs correcting for anyone who finds this later.

`max` is not what Lucene does for a phrase slot. `MultiPhraseQuery` hands every term's statistics to `BM25Similarity`, whose `idfExplain` **sums** them -- so summing is the Lucene-compatible behaviour and we already match it. `max` is `BlendedTermQuery`, which is a different thing: it is what a *standalone* fuzzy term does, and we match Lucene there too (`BlendedTermsCollector`).

[...]

Issue #1112, opened 2026-09-09
Variadic phrase scoring: a slot must contribute exactly one statistic


MBkkt:
## Idea

A variadic phrase is a phrase where a slot can be filled by any of several alternative terms — what a synonym expansion, a fuzzy term or an explicit alternatives list produces. Such a slot still contributes **once** to the phrase, conceptually: it is one position, and whether it was filled from one candidate term or from twelve should not change how many times that position influences the score.

It does today. `PhraseCollector::Finish` walks each phrase part, and for every alternative within the part calls `_scorer->collect(...)` against the **same** stats slot. Since the scorer accumulates into that slot, a slot with N alternatives contributes N times. Widening an expansion therefore changes the phrase's statistics even when nothing new matches, and a phrase silently changes character as soon as one of its slots becomes variadic.

## The template for the fix now exists

#1093 solved exactly this shape for fuzzy: a fuzzy expansion shares **one** stats slot, filled by `BlendedTermsCollector` with `df = max` and `ttf = sum` over the matched terms, rather than by repeated accumulation. The variadic phrase slot wants the same treatment, and that is most of the design work already done.

## Done looks like

- A variadic slot contributes exactly one statistic, whatever its number of alternatives — a slot with one alternative behaves identically to a plain term slot.
- Adding alternatives that match nothing does not change the score of any document.
- Consistency with the fuzzy path: the same expansion scored as a fuzzy term and as a variadic phrase slot yields the same slot statistic.
- Tests pinning the per-slot contribution rule, including a slot whose alternatives span a wide document-frequency range.
- Existing phrase scoring tests unchanged, or changed with the reason recorded.

## Note

Team work, not a student task. Still present after #1093, which fixed the same shape of problem for fuzzy but not for the phrase path.

## Links

- Blocks #1107.
- #1113 — how the single statistic should be derived, which is the other half of this.
- #445 — the per-document counterpart: which alternative actually matched in a given document.

MBkkt:
Superseded by #445.

This asks which single statistic a slot should contribute. Volatile idf makes the question disappear rather than answer it: the matched alternative is known per document, so its idf can be read instead of approximated by any static reduction over the whole expansion.

Keeping this open would mean choosing between approximations of a quantity we are about to read exactly.

Issue #445, opened 2026-03-27
Volatile IDF for variadic phrase and ngram similarity filters


gnusi:
> **Student task.** A self-contained piece of scoring work with an existing template to copy. `stats->idf +=` accumulates, and both the variadic phrase and n-gram similarity go through the same collector, so a slot's idf is still the sum over every alternative rather than the one that matched.
>
> **Why it is contained.** The mechanism already exists and does not have to be invented: per-document volatile data is already handed to the scorer this way for frequency and for boost. This task adds a third array of the same shape, so the shape of the answer is known and most of the work is threading it through the two iterators that need it.
>
> **What to decide.** Where the matched alternative's index is captured. The non-slop variadic path walks aligned positions and can read it as it goes; other paths move positions while probing, so the index has to be captured at the point a match is confirmed rather than read afterwards. Also: what happens when several alternatives match at the same position in one document — pick one, or combine.
>
> **Interacts with #1112 and #1113**, which are about the *prepare-time* statistic for a slot (one statistic per slot, `max` rather than a sum). This task is the per-document counterpart, and if it lands the prepare-time question changes shape, since per-alternative values would then be available at scoring time. Worth raising in the issue rather than assuming an order — do not treat it as a blocker without asking.
>
> **Done looks like:** a variadic phrase or n-gram similarity query scores a document using the idf of the alternative that actually matched in that document; a rare alternative no longer inflates the score of documents where only the common one matched; tests with alternatives spanning a wide document-frequency range; and no measured regression on the non-variadic path, which must not grow a branch for this.

---

## Problem

Filters like `VariadicPhrase` and ngram similarity operate over a disjunction of terms at each phrase position. The current scoring API computes IDF at query preparation time by calling `Scorer::collect()` once per term, accumulating into a shared stats buffer:

```cpp
// bm25.cpp:421 — IDF is summed across all terms
stats->idf += log1p((docs_with_field - docs_with_term + 0.5) / (docs_with_term + 0.5));
```

For a variadic phrase like `(word1|synonym1) NEAR (word2|synonym2)`, this produces:

```
idf = idf(word1) + idf(synonym1) + idf(word2) + idf(synonym2)
```

This is wrong — only the **matched** alternative at each position should contribute its IDF. A rare synonym inflates the score even when the common term matched.

## Expected Behavior

IDF should reflect which terms actually matched in each document. For example, if `word1` matched at position 1 and `synonym2` matched at position 2:

```
[...]

PR #1152, opened 2026-09-11
fix: sloppy phrases match and score like lucene

### LINKED ISSUE #1082: Sloppy phrases ignore match tightness (Lucene's sloppyWeight is computed but never scored)
A sloppy phrase currently scores the same whether the terms sit next to each
other or at the far edge of the slop window. Tight and loose matches rank
identically.

**History**

`SlopPhrase` computes the tightness factor and has always had it:

```cpp
// slop_phrase.hpp
score_t GetBoost() const noexcept {
  if (_best_distance == 0) return kNoBoost;
  return 1.f / (1.f + static_cast<score_t>(_best_distance));
}
```

- In `iresearch-main` it is computed and **discarded**: the only consumer of a
  positional `GetBoost()` in main's `search/` is `ngram_similarity_query.cpp:555`,
  and no phrase code reads it. So main's sloppy phrases score as exact ones.
- On the plan-layer branch it became live by accident: `TwoPhaseScored::kHasBoost`
  is `requires(const Slots& s){ s.Boost(); }`, so porting the class verbatim wired
  the factor into `BoostBlockAttr` and it started **multiplying the finished
  score**.
- That multiply is not what Lucene does either, so it has now been removed
  (`kHasBoost = false`, `GetBoost()` and `_best_distance` deleted) to restore
  main's behaviour. `MatchResult::best_distance` is kept — `slop_phrase_fuzz.cpp`
  and `slop_phrase_gather_tests.cpp` assert on it as a matcher invariant.

**What Lucene does**

`lucene-core-10.5.0` has `PhraseMatcher.sloppyWeight()` next to `maxFreq()`:
the tightness factor contributes to the phrase **frequency** the similarity
sees, i.e. it feeds `tf`, rather than scaling the final score. So a loose match
behaves like a lower term frequency.

**Decision needed**

1. Leave it: sloppy phrases rank by term statistics only, tightness ignored
   (current state, and main's).
2. Lucene parity: fold `1/(1+best_distance)` into the frequency handed to the
   scorer, not into a post-hoc multiplier.

Option 2 is the only one that makes `::slop(N)` express a ranking preference.
Note it interacts with the constant-score fast path (#1081): a factor inside
`tf` is invisible to `constant(N)`, which is the desirable side of doing it
Lucene's way — `constant(N)` stays N, while BM25 sees the looser match as
rarer.

### LINKED ISSUE #1083: Variadic slop phrases silently ignore per-term boosts (fuzzy weighting lost when ::slop is added)
Adding `::slop(N)` to a fuzzy phrase silently discards the fuzzy weighting.

```sql
ts_phrase(ts_levenshtein('quik', 1), 'fox')              -- 'quick' outranks 'quack'
ts_phrase(ts_levenshtein('quik', 1), 'fox')::slop(2)     -- 'quick' and 'quack' score the same
```

A variadic phrase slot can match several terms, and the expansion hands each
one its own boost (`phrase_filter.cpp:187` sets `_volatile_boost |= (boost !=
kNoBoost)`, so which alternative matched in a given document changes its
score). Non-slop variadic phrases honour that: `TakeBoost()` sums
`Traits::Boost(slot)` per aligned position and `GetBoost()` returns
`_phrase_boost / (pos.size() * _phrase_freq)`, the mean boost of the
alternatives that actually matched.

The slop path never instantiates it:

```cpp
// common/phrase_of.hpp, ResolveSlotMatcherOf
if constexpr (M == PhraseMatch::Slop) {
  using Slot = std::pair<search::PhraseVariadicPositions<Leaf>*, TermInterval>;   // no HasBoost
  return f.template operator()<SlopPhrase<Slot, Offs, HasFreq>, Leaf>(...);
} else {
  using Slot = std::pair<search::PhraseVariadicPositions<Leaf, HasBoost>*, TermInterval>;
  return f.template operator()<PhraseFrequency<Slot, Offs, HasFreq, kIntervals, HasBoost>, Leaf>();
}
```

`iresearch-main` is the same: `SlopVariadicPhraseFrequency<typename Adapter,
bool HasFreq>` has no `HasBoost` parameter at all, and main's
`volatile_boost` branch (`phrase_query.cpp:393`) selects the non-slop
`VariadicPhraseIterator`. So this has never worked in either engine — it is a
gap, not a regression.

**What the fix needs**

The ingredient exists — `common/phrase_variadic_pos.hpp:385` already provides
`Traits::Boost(pos)`, and `SlopPhrase` already has `using Traits =
TermPositionTraits<TermPosition>`. Threading `HasBoost` through the Slop branch
and `SlopPhrase::kHasBoost` is a few lines.

The work is *when* to read it. `pos.first->Boost()` returns the boost of the
alternative a slot is currently parked on. Non-slop can call `TakeBoost()`
during its aligned walk, so it always reads the winning ones. `SlopPhrase`
instead runs a min-cost search — a pair fast path (~`slop_phrase.hpp:286`) and
a recursive enumeration (~`:357`) — that moves positions while probing, so by
the time `Match()` returns the slots need not sit on the winning chain. The
boost therefore has to be captured inside both algorithms where `res.any =
true` is set, which means `MatchResult` grows a `score_t` and both paths thread
it.

**Semantics to decide**

- Sum over the best-distance chain only, or average over all matching chains
  (what non-slop does via `_phrase_freq`)?
- `!HasFreq` returns on the first match, so that path can only sample one
  chain.

**Ordering**

Best done after #1082, which decides whether a per-document phrase factor
scales the final score or feeds the frequency the scorer sees. Whatever is
chosen for match tightness should apply to per-term boosts too, so settling
#1082 first makes this a smaller and better-specified change.

### LINKED ISSUE #1114: Slop phrases with three or more words match differently from Lucene
## Idea

A slop phrase allows the words to be spread out and to come in the wrong order, each move costing from a shared budget. For two words this is unambiguous. For three or more, our matches diverge from Lucene's: documents Lucene accepts we reject, or the reverse, and the divergence grows with the number of words.

Sloppy matching over `n` words is a minimum-cost assignment problem, and the divergence is almost certainly in how the cost of a multi-word rearrangement is accumulated — Lucene charges for moves in a particular order and our matcher does not reproduce that. Because the budget is shared across the whole phrase, a difference in how one junction is charged changes which documents fit inside the budget at all, not merely their scores.

## What the fix needs

- A reproducer set: phrases of three or more words against documents where the two engines disagree, produced by differential testing against Lucene rather than by hand.
- The rule Lucene actually applies, written down, from its matcher rather than from documentation.
- Our matcher brought into agreement, or a deliberate documented divergence if agreement turns out to be indefensible.

## Done looks like

- A differential test over generated phrases and documents shows no disagreement with Lucene on match/no-match for phrases of arbitrary length.
- The charging rule is documented next to the matcher.
- Existing slop tests still pass, or the ones that encoded the wrong behaviour are updated with the reason recorded.

## Note

Team work, not a student task. This blocks the phrase-gaps combining work: rules cannot meaningfully be layered on top of a matcher whose base semantics disagree with the reference implementation.
## Links

- Blocks #1051 (combining phrase rules) and #1106 (unordered proximity).
- #1082, #1083 — the two scoring-side slop defects, which are separate from this match-side one.

zhemalb:
Fixes phrase queries with slop rejecting valid matches, and then follows the
consequences: the match was wrong, the frequency was wrong in a different way,
and the channel the score arrived on could not carry the answer.

## Match

The matcher summed deviations between neighbouring terms. It now applies
Lucene's rule: a match is accepted when the largest shift minus the smallest
does not exceed the slop, where a shift is `document position - query
position`.

$$\text{cost} = \max_i(d_i - q_i) - \min_i(d_i - q_i)$$

`a b c` against `a c b` at slop 2: shifts are 0, 2, 0, so the cost is 2 and it
matches. The old sum-of-deviations form rejected it.

n >= 3 previously followed a different rule from n == 2. Both now run the same
one-pass sweep, transcribed from `SloppyPhraseMatcher`. Closes #1114.

## Frequency

Match was the visible bug; frequency was the larger one. Lucene's
`PhraseScorer` accumulates `sum 1/(1+matchLength)` over that sweep. We counted
position tuples instead -- a median 1.5x divergence, up to 40x.

The sweep now produces Lucene's number directly. A clause publishes an integer
frequency plus a per-document scale, `scale = weight/freq`, so `tf = freq *
scale` is the float Lucene computes, not an approximation of it. Closes #1082.

## One channel, one rule

`BoostBlockAttr` becomes `ScaleBlockAttr`, published by the same producers and
read under a single rule:

> a scorer that has a frequency applies the scale inside `tf`, where a partial
> match saturates like any other weak evidence; a scorer without one applies it
> to the score, which is the only expression it has.

Both readings already existed -- the vector similarity scorer used the channel
as the score itself, BM25 and friends as a multiplier on the finished score --
so this names the rule rather than adding a second channel. Per-expansion
fidelity reduces by `min` within an occurrence and `max` across them: a phrase
is a conjunction, so one slot that barely matched cannot hide behind the others
matching perfectly.

## Two defects that removing `ReadAll` exposed

A sloppy **variadic** phrase dropped expansion fidelity entirely: the slop
branch pinned `HasBoost` to false, so `"quicm fox"~2` scored as though every
expansion were exact while the same phrase without slop did not. It could not
have been fixed earlier -- bulk decoding discarded which sub-iterator supplied
each position. Closes #1083.

And n >= 3 disagreed with n == 2 on same-position synonyms, because the bulk
path sorted without deduplicating. Both now read the same merged stream.

## Performance

`JoinPair` is gone: it was the n == 2 case of the same algorithm, so the sweep
takes its arity as a template parameter and `ResolveArity` pins it where it
can. That cost nothing and bought the skip -- a run-internal position never
raises `end`, so when a run provably cannot emit, the lead seeks straight to
`min(end - slop, runner + 1)`, clamped to the boundary or runs merge and the
count shifts.

Measured against `iresearch-main` on the benchmark corpus, warm, min of ten:

| | COUNT | TOP_100 |
|---|---|---|
| phrase (300) | 1.02 | 1.02 |
| slop, original 38 | 1.02 | 1.01 |
| variadic (104) | 0.99 | 0.98 |
| interval (38) | 1.02 | 1.01 |
| ngram (45) | 1.02 | 1.00 |

A dense-stopword `"the english restoration"~2` was 2.6x of main before the skip
and is at parity after.

## Verification

Thirteen repeated-term queries join the corpus -- the case that had **no**
differential coverage: the corpus contained no sloppy phrase that repeated a
term, and the fuzz oracle's repeat handling is the same algorithm as
production, so it was not independent there.

Patch agrees with Lucene on all thirteen; `main` disagrees on seven.

They are also 2.4x slower than main, worst 3.95x on `"the end of the"~1` with
identical counts. That is structural -- main anchors on the rarest slot and
binary-searches the rest, while a run-based sweep walks, and the run structure
is what produces Lucene's frequency. Written up in
`plans/phrase-and-ngram-semantics.md` rather than papered over.

The benchmark's Lucene engine also grew support for the two forms it silently
answered zero for -- a prefix inside a phrase (`MultiPhraseQuery` with the
expanded term set, which keeps `PhraseQuery` slop semantics) and a gap range (a
disjunction of fixed-gap phrases). All 493 comparable queries now agree on
counts: 300 phrase, 51 slop, 104 variadic, 38 interval. n-gram reports
UNSUPPORTED instead of a large unrelated number.

4221/4221 tests pass.

## One state for a phrase

Phrase and n-gram were one concept in three spellings. `FixedPhraseState` and
`VariadicPhraseState` differed in layout rather than meaning, and prepare
copied one into the other per segment -- not as an optimisation the plan layer
could skip, since `MakeVariadicPhraseOf` asserted the copy had happened. That
conversion is where the boost reduction silently disagreed with the matcher's
own `min`-within rule.

There is one `PhraseState` now: flat `metas` by value, `boosts` empty when
there are none, n+1 prefix offsets. Slot arity is data, so the filter reads
`Fixed()` and constructs the matching query -- no copy, no re-derivation. It
drops the parallel pointer vector, and `PhraseVariadicSlots` takes the offsets
instead of recomputing each one in O(n^2). `NGramState` and `PhraseState` share
`PostingsState`, which is what the two actually have in common.

A slot is a sub-filter, so one `SlotKind` answers every question the prepare
loop asks, and `ByPhraseOptions` caches nothing derived -- the cached counts
went stale under `LowerParts` and indexed a counter past the end of its block.

## Statistics a slot owes

Three bugs, all in how a slot's terms reach the scorer.

A **set slot** summed its terms into one counter, so BM25 saw `idf(sum df)`
where the slot owes `sum idf(df_i)`. Terms and set members are now driven
straight from the options on one shared iterator and collect by member index --
the index is query-static, so it is a valid key and no term bytes are held.
Only an expansion, whose terms differ per segment, needs a map.

`Finish` **collected empty counters**: a term present in no segment handed the
scorer `docs_with_term == 0`, which is the maximal idf for a term that exists
nowhere. It matters for n-gram, where a missing n-gram does not make the query
empty.

A **fixed-shaped segment with boosts** folded them into the query boost, so the
same fuzzy phrase scored by one formula where a slot expanded to two terms and
by another where it expanded to one. The fixed shape is now chosen only when
there are no boosts, and the scale channel carries them either way.

`TermSetOptions` loses its per-term boost, `min_match` and `merge_type`. The
product set none of them, `min_match` was asserted to be 1, and the boosts only
ever carried levenshtein keys -- those now travel as `TopTerm`. Regexp phrase
parts go with them: nothing outside tests ever built one.

## Lowering

Every rule was read for the configurations it declines, and the question asked
of each was whether a rewrite exists that keeps both the matches and the score.

**One was changing matches.** N-gram similarity rewrote to a phrase when
`min_match == terms_count`. The sequence wants increasing positions and a
phrase wants adjacency, so `{1, 2}` matched the document `1 9 2` and the
lowered form matched nothing. In production `allow_phrase` was
`static constexpr true`, so it fired on every such query. Removed, with a test.

**Several were possible and not taken.** A one-term set collapses to a term
slot, and an empty one makes the phrase `Empty` at any length. A constant score
survives being split into clauses when the boolean merges with `Max`, so term
sets and the n-gram disjunction lower there too. A merge type nobody can
observe becomes `Sum`, which lets the existing flatten path fire. Same-
direction range bounds merge. An unscored `All` is dropped where the bucket
still requires something else.

**Subsumption was missing entirely**, across all six bucket pairs: drop the
wider clause where a narrower one implies it, the narrower where a wider one
already matches or excludes it, and reduce to `Empty` where the required set is
excluded. Containment of term sets carries to documents however many terms a
document holds, so this needs nothing known about the field.

**A `MustNot` subtree scores nothing**, and the traversal never said so -- so
every score-gated rule refused inside a negation. The occurrence now travels
with the traversal and those subtrees optimise unscored, which unblocks seven
rules at once. Separately, nine gates moved from the query-wide `scored` flag
to whether the node's own score depends on terms, so an `Unscored` node stops
blocking dedup, fusion and absorption.

Two rewrites were implemented and reverted rather than left as unexplained
gaps: deduplicating a Should bucket at `min_should_match >= 2` (a document's
count falls by however many duplicates it matched, so no threshold change
compensates), and conjunctive narrowing to `Empty` (`AND(open, source, open)`
matches, because a document holds many terms on one field).

## Also in here

Containers follow the same rule as the scale channel -- a fixed size is not a
vector. `FixedArray`, `FixedRun` and `RunOf` collapse into
`containers::Fixed<T, N>` with `N == 0` choosing heap storage, and fifteen
members across the search layer stop allocating. The remaining vectors
genuinely grow.

The three skip readers shared their level state and differed only in `Read()`;
that state is now `SkipLevels<Copy>`. The pruned reader keeps its own
implementation -- it owns its previous state and redefines five of nine
operations -- but its levels were the only ones still reaching the heap, and
`kMaxSkipLevels` bounds them at five.

## Left open deliberately

- #445 volatile idf, which is where fuzzy fidelity should eventually live
  rather than on the scale channel, behind a `default`/`lucene` switch.
- #1158 reducing occurrences by max instead of sum.
- #1112 and #1113 closed as superseded by #445. Worth noting #1113's premise
  was wrong: Lucene *sums* slot idf, so summing is the compatible behaviour.
- #1162 carrying "this field holds one term per document" into the optimizer.
  `TokenTraits::unique` records it and `OptimizeContext` cannot see it, which is
  what stops the contradiction rewrites -- two different required terms reducing
  to `Empty`, and the same for a term outside a required range or prefix.
- Subsumption where the covering clause is an automaton or a levenshtein filter.
  `Optimize` runs the rules before `LowerAutomatons`, so no rule ever sees an
  `AutomatonFilter`; parity with main, no regression.

### DIFF
--- a/docs/cookbook/search/phrase-and-proximity-search.md
+++ b/docs/cookbook/search/phrase-and-proximity-search.md
@@ -60,7 +60,7 @@ Nothing — the tokens are not adjacent. With one unit of slop, the intervening 
 
 <SqlLogicTest id="cookbook/search/phrase-and-proximity-search/example_011" />
 
-`slop` is an edit budget for phrases, playing the role [Levenshtein distance](./fuzzy-search.md#levenshtein-matching-with-ts_levenshtein) plays for a single word — except the only edit it can buy is *moving* a term, never substituting or dropping one. It is the total number of positions the query tokens may be shifted to line up with the document, shared across the whole phrase: each token sitting between two query tokens costs one unit, and `slop := 0` is an exact phrase, identical to omitting it. Every token you type must still appear in the document, so no amount of slop rescues a misspelled word — that is [`ts_levenshtein`](./fuzzy-search.md)'s job, and the two compose.
+`slop` is an edit budget for phrases, playing the role [Levenshtein distance](./fuzzy-search.md#levenshtein-matching-with-ts_levenshtein) plays for a single word — except the only edit it can buy is *moving* a term, never substituting or dropping one. Line the query up with the document and note how far each token had to move: slop is the difference between the largest and the smallest of those moves, one figure for the whole phrase rather than a sum per token. When the tokens stay in order that is simply the number of words wedged between them, and `slop := 0` is an exact phrase, identical to omitting it. Every token you type must still appear in the document, so no amount of slop rescues a misspelled word — that is [`ts_levenshtein`](./fuzzy-search.md)'s job, and the two compose.
 
 ### Widening the window
 
@@ -80,7 +80,7 @@ A budget of 2 also pays for one swap of an adjacent pair, so a phrase can match 
 
 <SqlLogicTest id="cookbook/search/phrase-and-proximity-search/example_014" />
 
-Reordering is strictly more expensive than insertion: one intervening word costs 1, one transposition costs 2. If word order matters to you, keep `slop` at `0` or `1`.
+Reordering is strictly more expensive than insertion: one intervening word costs 1, one transposition costs 2, whether the swapped pair stands alone or sits inside a longer phrase. If word order matters to you, keep `slop` at `0` or `1`.
 
 ### Other spellings
 
--- a/docs/sql/functions/search/full-text.md
+++ b/docs/sql/functions/search/full-text.md
@@ -95,6 +95,8 @@ The gap form expresses proximity directly. To require `quick` within three token
 
 **Slop.** `slop := N` spends a budget of `N` position moves to line the query up with the document, shared across the whole phrase: `ts_phrase('quick fox', slop := 1)` matches `quick brown fox`, because shifting `fox` one position costs one unit. An intervening token costs 1 per token and a transposed adjacent pair costs 2, so `slop := 2` also matches `fox quick`. `slop := 0` is an exact phrase.
 
+Precisely, a match costs the spread of its shifts: shift each query token by its document position minus its query position, then take the largest shift minus the smallest. A swapped pair inside a longer phrase therefore still costs 2, so `ts_phrase('quick brown fox', slop := 2)` matches `quick fox brown`.
+
 When a `gap` is declared, the budget counts deviation from *that* gap rather than from adjacency: `ts_phrase('quick', 1, 'fox', slop := 1)` accepts `quick fox` and `quick a b fox` — one step either side of the declared single-token gap. Interval gaps already express a range, so `ts_phrase('a', [1, 3], 'b', slop := 2)` is an error.
 
 `(...)::slop(N)` applies the same budget as a modifier to an already-built phrase, including one from [`phraseto_tsquery`](#phraseto_tsquery). Lucene's `"..."~N` reaches it through [`to_tsquery`](#to_tsquery). The forms are mutually exclusive: specifying slop twice on one phrase is an error, and `::slop` on a non-phrase query (or on a `##` part) is rejected.
--- a/libs/iresearch/include/iresearch/analysis/ngram_tokenizer.hpp
+++ b/libs/iresearch/include/iresearch/analysis/ngram_tokenizer.hpp
@@ -22,6 +22,7 @@
 
 #pragma once
 
+#include <magic_enum/magic_enum.hpp>
 #include <tuple>
 #include <vector>
 
@@ -111,3 +112,27 @@ class NGramTokenizer final : public TypedTokenizer<NGramTokenizer>,
 extern template class TypedTokenizer<NGramTokenizer>;
 
 }  // namespace irs::analysis
+namespace magic_enum {
+
+// The one reflection of NGramMode, beside the enum: every TU that reflects it
+// must see the same names, or the linker mixes the per-TU tables (ODR). The
+// names are the user surface -- the `mode` option of the ngram dictionary.
+template<>
+constexpr customize::customize_t
+customize::enum_name<irs::analysis::NGramTokenizer::NGramMode>(
+  irs::analysis::NGramTokenizer::NGramMode value) noexcept {
+  using NGramMode = irs::analysis::NGramTokenizer::NGramMode;
+  switch (value) {
+    case NGramMode::All:
+      return "all";
+    case NGramMode::Prefix:
+      return "only_prefix";
+    case NGramMode::Suffix:
+      return "only_suffix";
+    case NGramMode::PrefixAndSuffix:
+      return "only_prefix_and_suffix";
+  }
+  return invalid_tag;
+}
+
+}  // namespace magic_enum
--- a/libs/iresearch/include/iresearch/analysis/pattern_tokenizer.cpp
+++ b/libs/iresearch/include/iresearch/analysis/pattern_tokenizer.cpp
@@ -124,19 +124,78 @@ re2::RE2::Options RegexOptions(int group) {
   return options;
 }
 
-delim::Finder SplitOnLiteral(std::span<const re2::Rune> runes, bool fold_case) {
-  bstring literal;
-  literal.reserve(runes.size() * utf8_utils::kMaxCharSize);
+bool AppendLiteral(std::span<const re2::Rune> runes, bool fold_case,
+                   bstring& literal) {
   for (const auto rune : runes) {
     if (fold_case && (rune >= 128 ||
                       absl::ascii_isalpha(static_cast<unsigned char>(rune)))) {
-      return {};
+      return false;
     }
     byte_type buf[utf8_utils::kMaxCharSize];
     literal.append(buf,
                    utf8_utils::FromChar32(static_cast<uint32_t>(rune), buf));
   }
+  return true;
+}
+
+bool AppendLiteralOf(re2::Regexp& re, bool fold_case, bstring& literal) {
+  const bool fold =
+    fold_case || (re.parse_flags() & re2::Regexp::FoldCase) != 0;
+  if (re.op() == re2::kRegexpLiteral) {
+    const re2::Rune rune = re.rune();
+    return AppendLiteral({&rune, 1}, fold, literal);
+  }
+  if (re.op() == re2::kRegexpLiteralString) {
+    return AppendLiteral({re.runes(), static_cast<size_t>(re.nrunes())}, fold,
+                         literal);
+  }
+  return false;
+}
+
+delim::Finder SplitOnLiteral(std::span<const re2::Rune> runes, bool fold_case) {
+  bstring literal;
+  literal.reserve(runes.size() * utf8_utils::kMaxCharSize);
+  if (!AppendLiteral(runes, fold_case, literal)) {
+    return {};
+  }
   return delim::FinderFor(std::move(literal));
+}
+
+delim::Finder SplitOnAlternation(re2::Regexp& re, bool fold_case) {
+  std::vector<bstring> delimiters;
+  delimiters.reserve(static_cast<size_t>(re.nsub()));
+  for (int i = 0; i < re.nsub(); ++i) {
+    bstring literal;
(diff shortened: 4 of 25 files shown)


MBkkt:
@zhemalb sign CLA plz: https://serenedb.com/cla

zhemalb:
> @zhemalb sign CLA plz: https://serenedb.com/cla

Just filled out the form, not seeing any new mail atm, though. I suppose the delivery is manual then, isn't it?

MBkkt (approved):

zhemalb:
Commit [39d933fd](https://github.com/serenedb/serenedb/commit/39d933fd95df7fea273755ecd2db199187ddf190) introduced inconsistencies, such as:

- the vector iterator still used Boost after the interface changed to Scale, causing NaN distances,
- the test reference wasn’t updated for the new scoring rules.

Currently fixed, @MBkkt so please do restart the CI build =)

PR #1157, opened 2026-09-13
perf: shared segment cursors, dictionary-code gathers and TIMESTAMP_NS stats for search GROUP BY

gnusi:
Search `GROUP BY` and windowed `ORDER BY ... LIMIT` queries over the OpenTelemetry corpus were slower than they needed to be for three independent reasons. This fixes all three; the two DuckDB-side pieces live in serenedb/duckdb#100 and the submodule points at that branch until it merges.

**One thread per segment.** Every Stream-mode scan handed a whole segment to one worker (`MaxThreads() == segments`), so a query over a two-segment index ran on two cores no matter what `threads` said, and with 36 segments on 32 threads the four workers that drew a second segment set the wall time. Each segment now has a shared cursor (`StreamSegmentCursor`: the plan root, its zonemap skipper and a mutex). Whoever holds the cursor pulls the next 2048 matches; every worker materializes its own runs through its own `HitBatcher`. Workers claim unclaimed segments first and then join the unfinished segment with the fewest workers, reserving the slot with a CAS so a burst of joiners spreads out instead of piling onto one segment. `MaxThreads()` for Stream is `max(threads, segments)`.

**Dictionary strings were decoded row by row.** A gather of a few hundred rows out of a 2048-row span on a `dict_fsst` column decoded and copied every string of the span, then the hash aggregate compressed each one back into an integer. The codec select now emits a dictionary vector over the block dictionary (fork PR), and `GatherDense` no longer runs the validity child's select for codecs that describe their own nulls, which would have flattened that vector again. `SELECT SeverityText, count(*) ... GROUP BY SeverityText` over 10.8M hits drops from 184 ms to 78 ms single-threaded.

**`Timestamp::TIMESTAMP_NS` lost the bucket rewrite.** Statistics did not propagate through casts between timestamp precisions, so `date_trunc('minute', ts::TIMESTAMP_NS)` on a TIMESTAMP column ran on a hash aggregate instead of the perfect-hash bucket plan. The fork PR lets min/max flow through those casts; the plan is pinned in `date_trunc_bucket_groups_precision.test`.

Measured against a build of the same base commit without these changes, same box, same loaded data, hot medians.

SearchBench layout (`gnusi/es-9.5` harness: search table, `refresh_interval=10000, compaction_interval=5000`, `split_by_non_alpha` dictionary on `Body`, first 100M rows of `part_000.parquet`, 36 segments, 32 threads):

| family | n | before ms | after ms |
|---|---|---|---|
| group_by | 14 | 283 | 195 |
| recent | 16 | 134 | 86 |
| count | 32 | 216 | 212 |
| top_k | 21 | 284 | 276 |
| join | 9 | 1708 | 1716 |

Q66 (`GROUP BY SeverityText`, `ServiceName = 'frontend' AND Body @@ 'failed'`) 23.3 -> 11.9 ms, Q56 15.3 -> 8.4, Q54 19.4 -> 9.1, Q71 16.1 -> 8.1. The minute histograms (Q61 to Q65) move 26 -> 23.5 ms; their time is Timestamp bitpacking decode and the aggregate, not parallelism.

Two-segment layout (inverted index with `INCLUDE` over a plain table, 100M rows, default compaction): the first query above goes 110 -> 17 ms and the `ts_all(['failed','order'])` minute histogram 90 -> 20 ms.

All 92 harness queries return the same results on both builds; the only differences are row order among ties (`ORDER BY ... LIMIT` with tied keys, `LIMIT` without `ORDER BY`, `GROUP BY` without `ORDER BY`), and the top-k score multisets are identical query for query.

Not touched: Count, TopK and TsDict modes still run one worker per segment. Range-splitting a segment between workers was tried and dropped: on the 36-segment layout it added nothing over the shared cursors, and window plans re-decode the posting prefix for every extra range because `PostingFill` has no skip-list seek.


gnusi:
1B measurements on the same setup (`gnusi/es-9.5` harness `create.sql`, full `part_000.parquet`, 320 segments, 120.4 GB, 32 threads). Before and after are builds of the same base commit run against the same loaded datadir; two interleaved passes per binary, fresh process each pass, hot = min of tries 2 and 3.

| family | n | before ms | after ms |
|---|---|---|---|
| group_by | 14 | 2009 | 1486 |
| recent | 16 | 150 | 100 |
| count | 32 | 988 | 1006 |
| top_k | 21 | 753 | 757 |
| join | 9 | 20118 | 20193 |

| query | before ms | after ms |
|---|---|---|
| Q54 `GROUP BY SeverityText`, `ts_any(['error','failed'])` | 122.5 | 49.8 |
| Q55 `GROUP BY SeverityText`, `'charge'` | 94.2 | 35.0 |
| Q56 `GROUP BY SeverityText`, `ts_all(['failed','order'])` | 92.6 | 44.8 |
| Q57 `GROUP BY ScopeName`, `ts_any` | 103.3 | 45.4 |
| Q58 `GROUP BY ScopeName`, regexp | 81.8 | 30.7 |
| Q60 `GROUP BY ScopeName`, `ts_any` | 103.0 | 44.8 |
| Q66 `GROUP BY SeverityText`, `ServiceName = 'frontend' AND 'failed'` | 163.4 | 91.0 |
| Q67 `GROUP BY SeverityText, ScopeName`, `ts_any` | 196.0 | 108.4 |
| Q61 to Q65 minute histograms | 156 to 222 | unchanged |
| Q69, Q70, Q72, Q74, Q75 `ORDER BY Timestamp DESC LIMIT 100` | 12 to 18 | 6 to 9.5 |

Count, top-k and join are untouched code paths and read as noise (0.98x to 1.00x). The minute histograms do not move at 1B: with 320 segments the scan was already parallel, and their time is the Timestamp decode plus the aggregate. Results agree query for query; the only differences are tie order in `ORDER BY ... LIMIT`, `LIMIT` without `ORDER BY`, and `GROUP BY` without `ORDER BY`.

MBkkt (changes requested):
should be min(threads, segments), not max(threads, segments). e.g. 10000 segments and 2 threads. should be possible to get these speedups with segments distributed to threads

server/connector/search_table_dispatch.h
```diff
@@ -62,6 +62,11 @@ void RejectIfSearchTable(catalog::TableEngine engine,
 // the entry so nothing reaches back into the catalog while the query runs, and
 // so the shard and the generated-PK counter -- shared side state, one per
 // table, never per version -- are pinned for the life of the plan.
+class SearchTableWriteOperator {
+ public:
+  virtual ~SearchTableWriteOperator() = default;
+};
```
MBkkt: bad

MBkkt (approved):

PR #1159, opened 2026-09-13
fix: CREATE INDEX IF NOT EXISTS scanned the whole table to discard it

deadtrickster:
**SLOP ALERT**

I'm debugging my RAGFlow integration and pipeline. Noticed constant background 4 GB/s write going. Chased it down to
`CREATE INDEX IF NOT EXISTS`. Then asked AGI to fix. 

```
Measured on a 1.2M-document RAGFlow corpus: the redundant create read at ~4 GB/s for minutes while making 1-4 MB/s of progress, starving concurrent ingest on the same box.

Sink now returns FINISHED, which FinishProcessing()es the pipeline (pipeline_executor.cpp:412). Finalize() already returned READY on !created, so nothing it would have done is skipped.

Verified against the same 300k-row table, one binary apart:
    before   300,000 rows scanned
    after      6,144 rows scanned (3 chunks already in flight)
    threads=1  2,048 rows, and STILL 2,048 after tripling the table to 900k

The residual is bounded by chunks in flight, not by table size, which is what the regression test asserts - with the bug that figure tracks the row count. The index is intact afterwards: a point lookup still answers.
```


MBkkt (approved):

MBkkt:
@deadtrickster Thanks for PR, I think in current state some tests are stuck, but you don't need to think about this. My collegue @pashandor789 rewriting catalog right now, so he will account your case and make create index if exists for case when index already exists fast.

PR #1183, opened 2026-09-17
perf: merge top-k inside the inverted-index scan and claim work in units

MBkkt:
The inverted-index scan was parallel only across segments. One segment meant one
worker however many threads were free, and a top-k query materialised every
worker's k candidates before an operator above threw most of them away. This
reworks how the scan divides and finishes its work.

**Top-k is merged inside the scan.** A new duckdb `consume_top_n` table-function
hook offers the limit and offset to the scan; when it takes them the `TOP_N`
operator is dropped and the columns that survive it are pushed into the scan's
own projection. The sort key the scan no longer hands back — a score computed
only for the ordering it now does itself — is never materialised. On four
segments at 32 threads this reads columns for 100 rows where main read 600.

**Work is claimed in units.** Whole segments while segments outnumber workers,
row-group units past that, with stealing. A stream scan whose segments run out
now joins a cursor another worker is still draining instead of idling, so four
uneven segments keep sixteen workers busy.

**Barriers no longer block.** The stats-prepare and top-k merge barriers park the
task with the executor rather than sleeping a worker, whenever the pipeline runs
under `TASK_EXECUTOR`; the prepare phase still falls back to a blocking wait in
other execution modes.

**ts_dict counts split across doc ranges.** Each worker counts its own range into
a shared per-term accumulator and the worker finishing the segment emits the sums.

Also: the scan moves to `server/connector/scan/`, EXPLAIN gains `Offset:` beside
`Top:`, and `sdb_scan_split` / `sdb_scan_order` / `sdb_scan_no_split_row_groups`
are added for benchmarking and tests.

### Results

Same datadir, same box, both binaries:

| | main t1 | this t1 | main t16 | this t16 |
| :-- | --: | --: | --: | --: |
| ts_dict, filtered | 161ms | 130ms | 163ms | 14ms |
| count / docs / top, filtered | ~28ms | ~28ms | ~29ms | ~6.6ms |
| hits, projection | 458ms | 417ms | 47ms | 37ms |
| `PREPARE`/`EXECUTE`, rows | — | — | 515ms | 48ms |

Multi-segment (4 uneven segments) is at parity with main from t1 to t4 and ahead
at t8/t16.

### Fixes found on the way

- `TermCountsOf::Count` bypassed range clipping for single-document terms and
  read out of bounds under a bounded scan (server crash; covered by a test).
- `LazyBitset::Reach` clamped after its early return, so a call past the end
  dereferenced a null fill node on a folded set.
- `PhysicalExecute` hid the real sources from the wire collector, so any
  row-returning `EXECUTE` ran single-threaded.
- A table function may now return `BLOCKED` without handing back async tasks.

### Testing

Full `sdb` sqllogic sweep green on both wire protocols. New coverage for the
consumed top-k plan, ts_dict range splitting including single-document terms,
stream cursor joining, and a `TopKFetchBudget` failure point that fails the query
if top-k reads columns for more than `limit` rows. Each new test was checked by
reverting the fix it covers and confirming it fails.


pashandor789 (approved):

MBkkt:
Superseded by #1196, which is the same rework with the doc range carried in `Root::Run(min, max, ...)` instead of bound at plan time. Closing this one; the branch `mbkkt/simple-improvement` stays as the range-bound alternative.

PR #1195, opened 2026-09-18
fix: one vended iceberg credential secret per storage path

pashandor789:
Follow-up to #1165, closing the third issue from the same field report: *"leaked credential secrets poison scope resolution"*.

Bumps two gitlinks:
- serenedb/duckdb-iceberg#15 — the fix
- serenedb/duckdb#103 — the `CreateSecret` overload it needs

## Problem

Vended storage credentials are registered as duckdb secrets so httpfs can resolve them by path. The secret name carried the transaction id, so every transaction that read an iceberg table added another secret with the same scope, and duckdb breaks an equal-scope tie on the name — an expired credential could outrank the fresh one. Measured on the reported version: five reads, five secrets.

## Fix

Name the secret after the **storage path it unlocks**, and register it through a committed transaction rather than the reader's:

- a secret created inside a transaction is invisible to a reader that started earlier (`CatalogSet::UseTimestamp`), and an entry written through a committed transaction is visible to everyone
- two sessions scanning one table from inside their own transactions otherwise both create the same entry and one fails with `Catalog write-write conflict`

Registration is then skipped while the cached credential is still valid, so a scan writes to the catalog only on first read and after a re-vend. Token refresh is unchanged: with an always-expired token, three reads in one transaction still produce four `loadTable` calls.

## Test

`tests/sqllogic/sdb/pg/simple/vended_credentials_iceberg.test_slow`, on the existing iceberg-rest + MinIO fixture, on both wire engines. It attaches with `sigv4` deliberately — with `AUTHORIZATION_TYPE 'none'` the fixture returns no credentials and no internal secret is ever created, so the bug is invisible there.

| serened built from | outcome |
| --- | --- |
| duckdb_iceberg `9756778eb2` (pre-#1165, the reported version) | **fails** — 2 secrets and growing |
| duckdb_iceberg `4bf1e9648a` (current main) | **fails** — 0, the old cleanup churns it away |
| duckdb_iceberg `ad1502a3c9` (first attempt here) | **fails** — `Catalog write-write conflict` between two readers |
| this branch | **passes** |

The last row is why the test carries two overlapping transactions on one table: the first attempt at this fix traded the leak for a concurrency failure, and nothing in the suite would have caught it.

`catalog_iceberg` and `scan_iceberg` stay green; duckdb `test/sql/secrets/*` and `test/sql/catalog/*` pass (one pre-existing failure expecting `INSTALL` from a URL, unsupported in SereneDB).

Both gitlinks point at PR branches and need a re-bump once those merge.


pashandor789:
Closing: the approach in this PR regresses concurrency — two sessions reading the same iceberg table hit a catalog write-write conflict on the shared secret name. Reverting to the shipped behaviour.

PR #1199, opened 2026-09-19
fix: clear vended iceberg credential secrets, and keep them per path

pashandor789:
Follow-up to #1165, closing the third issue from the same field report: *"leaked credential secrets poison scope resolution"*.

Bumps `third_party/duckdb_iceberg` to serenedb/duckdb-iceberg#16 and adds the regression test.

## What was wrong

**Reads never cleared their credential.** Cleanup only ran on the write path, so every transaction that read an iceberg table left its vended credential behind — 5 reads, 5 secrets. They all claim the same scope, and duckdb breaks an equal-scope tie by comparing secret names, so an expired entry can outrank the fresh one. That is the 401 the customer reported.

**The name carried no path.** Two attached catalogs each holding a table of the same name, read in one transaction, collapsed to a single secret scoped to the second catalog's bucket — the first table's files were then covered by nothing.

## Fix

A credential is now named after the transaction that fetched it and the storage path it unlocks, and is dropped when that transaction ends — commit, failed commit, or rollback, plus the internal helper transactions that vend their own while committing.

## Test

`tests/sqllogic/sdb/pg/simple/vended_credentials_iceberg.test_slow`, on the existing iceberg-rest + MinIO fixture:

* one credential per table while the transaction runs, **with the path in its name**
* re-reads inside that transaction replace it — the count stays 1 rather than piling up
* nothing left after commit, after rollback, or after repeated autocommit reads
* two tables in one transaction keep separate credentials and scopes
* two named connections interleaving transactions on the same table both succeed

It attaches with `sigv4` deliberately: with `AUTHORIZATION_TYPE 'none'` the fixture vends nothing and no internal secret is ever created, so a test written that way passes against broken code.

Verified: passes on both wire engines, **fails on current main** (`expected 1, actual 0` — main's name has no path), and `catalog_iceberg` / `scan_iceberg` stay green.

## Beyond the test

| check | result |
| --- | --- |
| 135 mixed ops (reads, transactions, rollbacks, errors in a txn, insert/update/delete, cross-catalog) | **0** leftover secrets, every round |
| token TTL 45s, one 70s transaction | **5** re-mints, one credential live throughout, 0 read errors |
| two catalogs sharing a credential prefix | 4 secrets, each covering its own location (was 2) |
| iceberg suite | 473 cases, 10 failures — identical to the pre-change baseline |

Refresh-on-expiry from #1165 is untouched — no line in the fork commit mentions refresh or expiry.

The gitlink points at the fork's PR branch; it needs a re-bump once serenedb/duckdb-iceberg#16 merges.


pashandor789:
Superseded by #1200, which takes upstream's fix (transaction-scoped secrets) instead of our own naming plus an explicit sweep.

